PR #14, opened 2020-08-18
Improve wording of ExoPlayer setting.

KucharczykL:
`Enable the experimental ExoPlayer video player which supports more video formats and codecs is more integrated into the OS`

The part `is more integrated into the OS` seems like it's talking about the codecs (`and codecs is more integrated...`) which I think was not the intention and is also not grammatically correct (`codecs is`). I have added `, and` to make it clear what is being said.


Maxr1998:
Oops, I reworded it very recently and accidentally introduced this grammar mistake. Thanks for spotting it!

PR #19, opened 2020-08-19
Add editorconfig

### LINKED ISSUE #4: Use EditorConfig from Android TV app
We have 2 Android apps now based on Kotlin. I think it would be a good thing to have the same styling guidelines for both of them. At this moment the android-next app uses spaces as indentation while Android TV chose to use tabs.

The EditorConfig for Android TV is [here](https://github.com/jellyfin/jellyfin-androidtv/blob/master/.editorconfig). We could just copy it over and reformat all Kotlin files.

nielsvanvelzen:
Copied from ATV but changed the indent style as the code base is currently 100% spaces 😢 .

Closes #4

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*]
+insert_final_newline = true
+
+[{*.kts, *.kt}]
+charset = utf-8
+indent_style = space
+indent_size = 4


Maxr1998:
For reference, [Kotlin language conventions](https://kotlinlang.org/docs/reference/coding-conventions.html) explicitly recommend spaces, so I think it's fine :P

Maxr1998 (approved):

PR #25, opened 2020-08-19
Add basic playback info dialog

### LINKED ISSUE #22: Add dialog with playback/transcoding info

Maxr1998:
Closes #22.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -18,6 +18,7 @@
     <string name="downloading">Downloading</string>
 
     <string name="menu_item_none">None</string>
+    <string name="playback_info">Transcoding: %b\n\nVideo streams:\n%s\n\nAudio streams:\n%s</string>
 
     <string name="activity_name_settings">Jellyfin Settings</string>
     <string name="pref_category_music_player">Music player</string>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -18,7 +18,10 @@
     <string name="downloading">Downloading</string>
 
     <string name="menu_item_none">None</string>
-    <string name="playback_info">Transcoding: %b\n\nVideo streams:\n%s\n\nAudio streams:\n%s</string>
+    <string name="playback_info_transcoding">Transcoding: %b</string>
+    <string name="playback_info_video_streams">Video streams:\n</string>
+    <string name="playback_info_audio_streams">Audio streams:\n</string>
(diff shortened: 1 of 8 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -18,6 +18,7 @@
     <string name="downloading">Downloading</string>
 
     <string name="menu_item_none">None</string>
+    <string name="playback_info">Transcoding: %b\n\nVideo streams:\n%s\n\nAudio streams:\n%s</string>
```
nielsvanvelzen, line 21: I think this will become a mess when something gets added and all translations need to be updated

nielsvanvelzen (commented):
Code looks good but I have one comment on the strings.

_next time screenshots of UI changes are appreciated_

app/src/main/res/values/strings.xml
Maxr1998: > I think this will become a mess when something gets added and all translations need to be updated

Hm, this is true. I'll split them up into multiple strings, and upload some screenshots afterwards.

Maxr1998:
Here's a screenshot as well:
Screenshot

app/src/main/res/values/strings.xml
```diff
@@ -18,6 +18,10 @@
     <string name="downloading">Downloading</string>
 
     <string name="menu_item_none">None</string>
+    <string name="playback_info_transcoding">Transcoding: %b</string>
+    <string name="playback_info_video_streams">Video streams:\n</string>
```
nielsvanvelzen, line 22: Maybe move the `\n` part to the code because it should always be included

nielsvanvelzen (changes requested):
I think it would be better to move the text view to a corner (top left or top right) and remove some padding. Right now it's pretty much impossible to watch something with the playback info enabled.

The web player or other services (like YouTube) use the same approach.

Maxr1998:
> I think it would be better to move the text view to a corner (top left or top right) and remove some padding. Right now it's pretty much impossible to watch something with the playback info enabled.
> 
> The web player or other services (like YouTube) use the same approach.

The playback info is part of the playback interface though, and gets automatically hidden once you tap anywhere outside of it. I _can_ make it that it can be shown independently from the playback interface, but I'm not sure whether we want that.

Maxr1998:
This is how it looks now:
Player controls visible
Player controls hidden

nielsvanvelzen (approved):

PR #30, opened 2020-08-20
Use constants for intent actions and extras

Maxr1998:


nielsvanvelzen (approved):
I prefer having the constants in the relevant file instead of one big constants file but the changes look good.

Maxr1998:
> I prefer having the constants in the relevant file instead of one big constants file but the changes look good.

It depends, in this case the constants span multiple files, and I prefer to keep files concise, and not bloat them with constant declarations.

PR #50, opened 2020-08-24
Split up web app and connection handling, use view binding for setup layout

Maxr1998:
Extract connection handling into extra class, don't reset server URL on errors, replace toasts with error text.


Maxr1998:
@nielsvanvelzen I have decided to put the multi-server stuff into the next PR, I have to spend some extra thoughts before implementing that.

app/src/main/java/org/jellyfin/mobile/webapp/ConnectionHelper.kt
```diff
@@ -0,0 +1,207 @@
...
+import org.jellyfin.mobile.WebappActivity
+import org.jellyfin.mobile.utils.Constants
+import org.jellyfin.mobile.utils.Constants.SERVER_INFO_PATH
+import org.jellyfin.mobile.utils.requestNoBatteryOptimizations
+import org.json.JSONException
+import org.json.JSONObject
+import timber.log.Timber
+import java.io.IOException
+
+
```
nielsvanvelzen, line 36:
```suggestion

```
(1 empty line too many)

app/src/main/java/org/jellyfin/mobile/webapp/ConnectionHelper.kt
```diff
@@ -0,0 +1,207 @@
...
+        connectionErrorText.isVisible = false
+    }
+
+    private suspend fun checkServerUrlAndConnection(enteredUrl: String): HttpUrl? {
+        val normalizedUrl = enteredUrl.run {
+            if (lastOrNull() == '/') this
+            else "$this/"
+        }
+
+        val urls = when {
```
nielsvanvelzen, line 152: We should change this part to use the helper method from the apiclient in the future.

app/src/main/java/org/jellyfin/mobile/webapp/ConnectionHelper.kt
```diff
@@ -0,0 +1,207 @@
...
+        }
+
+        val isValidInstance = try {
+            val serverInfo = JSONObject(serverInfoResponse)
+            val version = serverInfo.getString("Version")
+                .split('.')
+                .mapNotNull(String::toIntOrNull)
+            when {
+                version.size != 3 -> false
+                version[0] == 10 && version[1] < 3 -> true // Valid old version
```
nielsvanvelzen, line 184: We might want to add the "supported version" as a constant somewhere so it can be changed more easily.

app/src/main/java/org/jellyfin/mobile/webapp/ConnectionHelper.kt
```diff
@@ -0,0 +1,207 @@
...
+                else -> serverInfo.getString("ProductName") == "Jellyfin Server"
```
nielsvanvelzen, line 185: We should always check the product name (it won't check if it can check the version right now)

app/src/main/java/org/jellyfin/mobile/webapp/ConnectionHelper.kt
Maxr1998: Definitely, yes.

app/src/main/java/org/jellyfin/mobile/webapp/ConnectionHelper.kt
Maxr1998: I think the check is only there because of the "ProductName" variable that isn't available before 10.3..

app/src/main/java/org/jellyfin/mobile/webapp/ConnectionHelper.kt
Maxr1998: I think that's intended, see above (I copied the code from @thornbill's Expo client).

app/src/main/java/org/jellyfin/mobile/webapp/ConnectionHelper.kt
nielsvanvelzen: Hmm, I can imagine that some features in the nativeshell (and in the future, the apiclient) need a minimum version set too. Might even be a good thing to require 10.6.x at the release too.

app/src/main/java/org/jellyfin/mobile/webapp/ConnectionHelper.kt
```diff
@@ -0,0 +1,208 @@
...
+class ConnectionHelper(private val activity: WebappActivity) {
+    private val appPreferences: AppPreferences get() = activity.appPreferences
+    private val rootView: FrameLayout get() = activity.rootView
+    private val webView: WebView get() = activity.webView
+
+    private var cachedInstanceUrl: HttpUrl? = null
+    var connected = false
+        private set
+
+    private val connectServerBinding: ConnectServerBinding by lazy { ConnectServerBinding.inflate(activity.layoutInflater, rootView, false) }
```
nielsvanvelzen, line 46:
```suggestion
    private val connectServerBinding: ConnectServerBinding by lazy {
         ConnectServerBinding.inflate(activity.layoutInflater, rootView, false)
     }
```
(indentation might be bad)/ Line is a bit long so I'd split it

nielsvanvelzen (approved):

PR #64, opened 2020-08-25
Use different application id, app name and icon for debug builds

Maxr1998:
I tested this locally and it seems to be working fine. I think we don't use the package name hardcoded anywhere anyway.


nielsvanvelzen:
Can we use the debug icon for debug builds? (it's the orange one)

Maxr1998:
> Can we use the debug icon for debug builds? (it's the orange one)

There is a debug icon? But yeah, sure. Just need to find the resources.

app/build.gradle.kts
```diff
@@ -12,7 +12,7 @@ plugins {
 android {
     compileSdkVersion(30)
     defaultConfig {
-        applicationId = "org.jellyfin.mobile"
```
nielsvanvelzen, line 15: For ATV we only append .debug on debug, I think we should mirror that behavior

nielsvanvelzen: (so always include the .mobile part)

nielsvanvelzen (changes requested):

app/build.gradle.kts
Maxr1998: I decided against that because the package name then gets quite long and gets a bit weird..

nielsvanvelzen: While I get your point I still think it's a bad decision. "org.jellyfin" is basically our prefix while "mobile" is the name for this specific app and "androidtv" is the tv app.

The ".debug" part could be interpreted as a flavor (although not as "android flavor") which is added only for non-default builds.

Maxr1998: Hmm, that argument makes sense. I'm still not 100% happy with the alternative. Maybe `.dbg` or `.dev` as suffix to keep it shorter?

nielsvanvelzen: We could do that but I'd like to keep this consistent with the ATV app so we'd have to change it there too.

dkanada: Appending `.debug` is pretty common from what I've seen, I think that would be fine.

Maxr1998: Fine, changed it.

nielsvanvelzen (approved):

PR #68, opened 2020-08-26
Add apiclient dependency, resolve device info via apiclient

Maxr1998:
This is for now the most basic use of the apiclient, I'm already working on using it for more things such as server connecting and progress submission, but that still needs some planning regarding architecture and design.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -55,6 +55,7 @@ dependencies {
     implementation(Dependencies.Kotlin.coroutinesAndroid)
 
     // Core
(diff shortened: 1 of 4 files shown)


nielsvanvelzen (approved):
Looks good, also a friendly reminder that the next major version of the apiclient will refactor the api part (like using coroutines and use camelcase naming). So quite some additional changes need to happen in the app when that update is ready.

nielsvanvelzen:
The sample in the apiclient repository has a helper function to use the current apiclient with coroutines, might come in handy if you plan to use it now

PR #78, opened 2020-08-30
Use Coroutine Channels for Webapp communication

Maxr1998:
Replaces (overkill) binder-based IPC Messenger/Handler, supports receiving player events from webapp in ExoPlayer.


nielsvanvelzen (approved):
Looks good, I noticed some behavior changes in the code but I assume those are all good.

Maxr1998:
> Looks good, I noticed some behavior changes in the code but I assume those are all good.

Yep, should be.

PR #100, opened 2020-09-05
Change battery optimization dialog to snackbar

nielsvanvelzen:
It's really annoying for first-time users to get that annoying alert. The snackbar will always show when the app starts unless the "ok" button is pressed once or the optimization is applied.

I changed the main activity to a coordinatorlayout because the snackbar requires that for the dismiss swipe gesture. I had to change some additional things in the webview inset code thingy for it to work properly.

image

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/MainActivity.kt
+++ b/app/src/main/java/org/jellyfin/mobile/MainActivity.kt
@@ -11,12 +11,10 @@ import android.os.Bundle
 import android.os.IBinder
 import android.view.OrientationEventListener
 import android.webkit.*
-import android.widget.FrameLayout
 import androidx.appcompat.app.AppCompatActivity
+import androidx.coordinatorlayout.widget.CoordinatorLayout
 import androidx.core.content.ContextCompat
-import androidx.core.view.ViewCompat
-import androidx.core.view.doOnNextLayout
-import androidx.core.view.updatePadding
+import androidx.core.view.*
 import androidx.lifecycle.lifecycleScope
 import androidx.webkit.WebResourceErrorCompat
 import androidx.webkit.WebViewClientCompat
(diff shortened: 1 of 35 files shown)


nielsvanvelzen:
I would've used a [banner](https://material.io/components/banners) if the material library supported it on Android, for some reason they only implemented it on ios and flutter.

app/src/main/java/org/jellyfin/mobile/MainActivity.kt
```diff
@@ -70,15 +68,17 @@ class MainActivity : AppCompatActivity(), WebViewController {
...
 
         // Bind player service
         bindService(Intent(this, RemotePlayerService::class.java), serviceConnection, Service.BIND_AUTO_CREATE)
 
         // Handle window insets
         setStableLayoutFlags()
         ViewCompat.setOnApplyWindowInsetsListener(rootView) { v, insets ->
-            v.updatePadding(top = insets.systemWindowInsetTop, bottom = insets.systemWindowInsetBottom)
+            val layoutParams = webView.layoutParams as CoordinatorLayout.LayoutParams
+            layoutParams.updateMargins(top = insets.systemWindowInsetTop, bottom = insets.systemWindowInsetBottom)
```
Maxr1998, line 80: Does setting the padding not work for CoordinatorLayout? :flushed:

Maxr1998 (approved):

Maxr1998:
Hopefully Weblate doesn't ruin the show now ^^

app/src/main/java/org/jellyfin/mobile/MainActivity.kt
nielsvanvelzen: It does, but it will make the snackbar appear in a wrong position. So i'm now using margins on the webview to prevent that :)

PR #107, opened 2020-09-08
Rework player implementation setting

Maxr1998:
Allow to switch between web player, ExoPlayer and external player via client settings

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/settings/VideoPlayerType.java
+++ b/app/src/main/java/org/jellyfin/mobile/settings/VideoPlayerType.java
@@ -1,0 +1,15 @@
+package org.jellyfin.mobile.settings;
+
+
+import androidx.annotation.StringDef;
+
+@StringDef({
+        VideoPlayerType.WEB_PLAYER,
+        VideoPlayerType.EXO_PLAYER,
+        VideoPlayerType.EXTERNAL_PLAYER
+})
+public @interface VideoPlayerType {
+    String WEB_PLAYER = "webui";
+    String EXO_PLAYER = "exoplayer";
+    String EXTERNAL_PLAYER = "external";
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/jellyfin/mobile/settings/VideoPlayerType.java
```diff
@@ -0,0 +1,15 @@
...
+
+@StringDef({
+        VideoPlayerType.WEB_PLAYER,
+        VideoPlayerType.EXO_PLAYER,
+        VideoPlayerType.EXTERNAL_PLAYER
+})
+public @interface VideoPlayerType {
+    String WEB_PLAYER = "web";
+    String EXO_PLAYER = "exo";
+    String EXTERNAL_PLAYER = "external";
```
nielsvanvelzen, line 14:
```suggestion
    String WEB_PLAYER = "webui";
    String EXO_PLAYER = "exoplayer";
    String EXTERNAL_PLAYER = "external";
```

nielsvanvelzen (changes requested):
Looks good, some notes though:

- The "allow playing audio in the background..." setting should be disabled when the video player is NOT native.
- The "external video player" isn't vlc or mxplayer only but those contain some extra code to work better because android apparently doesn't provide a useful api for that

Maxr1998:
> * The "allow playing audio in the background..." setting should be disabled when the video player is NOT native.

For that, I'd have to do further changes to the settings library first (as in, add a callback). Don't know if it's important enough to do that now or keep that for later. The preference at least says it's for the native player only.

nielsvanvelzen:
We can do it later but we should make an issue then so we won't forget it.

nielsvanvelzen (approved):

PR #110, opened 2020-09-09
Properly report playback stop events

Maxr1998:


app/src/main/java/org/jellyfin/mobile/player/PlayerViewModel.kt
```diff
@@ -107,11 +110,33 @@ class PlayerViewModel(application: Application) : AndroidViewModel(application),
...
+            val state = player.playbackState to player.currentPosition
             player.removeListener(this)
             player.release()
+            state
         }
         _player.value = null
+
+        val mediaSource = mediaSourceManager.jellyfinMediaSource.value
+        if (playerState != null && mediaSource != null) {
+            // viewModelScope is already cancelled at this point, so we need a fallback
```
nielsvanvelzen, line 123: I think you can fix this by calling `releasePlayer` in the `onDestroy` function instead of `onCleared`.

image

nielsvanvelzen: And what about the lifecycle observer (PlayerLifecycleObserver)? Won't that one trigger before the onCleared?

Maxr1998: It'd also trigger on every `onDestroy` from the activity afaik.

Maxr1998: Interestingly, no. The (Player)LifecycleObserver correctly handles rotation events and e.g. only dispatches `onCreate` and `onStop` once, but onDestroy interestingly never fires. I assume that the ViewModel gets destroyed before the activity.

nielsvanvelzen (approved):

PR #111, opened 2020-09-09
ExoPlayer: Fix playlists (like intros) not working

Maxr1998:


nielsvanvelzen:
It worked partially for me. It worked when playing from the details page but not from the home screen.

```
externalplayer.js?v=9:32 Uncaught (in promise) TypeError: Cannot read property '0' of undefined
    at Object.self.canPlayItem (externalplayer.js?v=9:32)
    at playbackmanager.js?v=9:1
    at Array.filter (<anonymous>)
    at getPlayer (playbackmanager.js?v=9:1)
    at Object.onPlaybackStopped (playbackmanager.js?v=9:1)
    at bundle.js?v=9:266
    at Array.forEach (<anonymous>)
    at Object.trigger (bundle.js?v=9:266)
    at exoplayer.js?v=9:167
    at new Promise (<anonymous>)
```

I fixed that, but now I found another bug. When playing something and pressing the back button (in the Android navigation bar) to stop the current playing item the web ui will automatically open it again.

nielsvanvelzen:
Actually, what happens is when playing a series from the start. When stopping the player at episode 2 the webui will start episode 3. It does this until the playlist is empty.

nielsvanvelzen (approved):

Maxr1998:
> It worked partially for me. It worked when playing from the details page but not from the home screen.
> 
> ```
> externalplayer.js?v=9:32 Uncaught (in promise) TypeError: Cannot read property '0' of undefined
>     at Object.self.canPlayItem (externalplayer.js?v=9:32)
>     at playbackmanager.js?v=9:1
>     at Array.filter (<anonymous>)
>     at getPlayer (playbackmanager.js?v=9:1)
>     at Object.onPlaybackStopped (playbackmanager.js?v=9:1)
>     at bundle.js?v=9:266
>     at Array.forEach (<anonymous>)
>     at Object.trigger (bundle.js?v=9:266)
>     at exoplayer.js?v=9:167
>     at new Promise (<anonymous>)
> ```
> 
> I fixed that, but now I found another bug. When playing something and pressing the back button (in the Android navigation bar) to stop the current playing item the web ui will automatically open it again.

Wait, that's something else though.. something in external player. If you want, I can include the fix still into this PR.

app/src/main/assets/native/externalplayer.js
```diff
@@ -29,7 +29,7 @@ define(['events', 'playbackManager', 'toast'], function (events, playbackManager
         };
 
         self.canPlayItem = function (item, playOptions) {
-            var mediaSource = item.MediaSources[0] || false;
+            var mediaSource = item.MediaSources && item.MediaSources[0] || false;
```
nielsvanvelzen, line 32: Basically the `&&` will convert both the left and right side to booleans. The result will be `true` or `false.
This will then end up doing `true || false`. So the `|| false` part can be removed.

Maxr1998: Fixed. Easy commit farming I guess :D

CarlosOlivo:
> Actually, what happens is when playing a series from the start. When stopping the player at episode 2 the webui will start episode 3. It does this until the playlist is empty.

If you mean the externalplayer it is correct, since there is no way to play a playlist directly in an external player on Android (unless the player supports it), you can only know when a single element starts and ends due to the limitations of the Intent Actions Results

A workaround would be to treat the playback list as in Chromecast or as in music player like [this](https://user-images.githubusercontent.com/61046533/92014811-1db83700-ed50-11ea-9dbe-54b6d092de8d.png) that plays the next element automatically but I don't know how to do it

Maxr1998:
I think external player should be able to function similiarly as the ExoPlayer now, just starting the next video after playback ended.

nielsvanvelzen (approved):

PR #114, opened 2020-09-11
Use display language when available

### LINKED ISSUE #83: "Client settings" parameter page not following the user display language
**Describe the bug**
The "client settings" parameter page doesn't follow the display language used by the logged user but the language used by Android.

**To Reproduce**

1. Set a display language different than the phone's
2. Restart the app to apply the changes
3. Go to "client settings" and see the strings displayed in the phone's language

**Expected behavior**
It should use the display language specified by the user.

 - App version: 2.0.0-rc.3
 - Android version: Android 8
 - Device manufacturer and model: Samsung Galaxy A5 2017
 - Jellyfin server version: 10.6.4
 - The type of video player implementation used: N/A

CarlosOlivo:
In the next order:
1. User settings
2. ~~Server settings~~
3. Android settings

For now, it's necessary to restart the application manually when the language is changed, as long as no [events](https://github.com/jellyfin/jellyfin-web/blob/c8b6b44b7bf79c5a14055c6262b85b5af9cd7554/src/scripts/globalize.js#L259) are handled.

Closes #83

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/MainActivity.kt
+++ b/app/src/main/java/org/jellyfin/mobile/MainActivity.kt
@@ -158,6 +158,7 @@ class MainActivity : AppCompatActivity(), WebViewController {
                                     val token = server.getString("AccessToken")
                                     apiClient.ChangeServerLocation(address)
                                     apiClient.SetAuthenticationInfo(token, user)
+                                    initLocale()
                                 } catch (e: JSONException) {
                                     Timber.e(e, "Failed to extract apiclient credentials")
                                 }
--- a/app/src/main/java/org/jellyfin/mobile/settings/SettingsActivity.kt
+++ b/app/src/main/java/org/jellyfin/mobile/settings/SettingsActivity.kt
@@ -20,6 +20,7 @@ class SettingsActivity : AppCompatActivity() {
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+        setTitle(R.string.activity_name_settings)
         setContentView(R.layout.activity_settings)
         setSupportActionBar(findViewById(R.id.toolbar))
         supportActionBar?.setDisplayHomeAsUpEnabled(true)
--- a/app/src/main/java/org/jellyfin/mobile/utils/ApiExtensions.kt
+++ b/app/src/main/java/org/jellyfin/mobile/utils/ApiExtensions.kt
@@ -3,6 +3,7 @@ package org.jellyfin.mobile.utils
 import org.jellyfin.apiclient.interaction.ApiClient
 import org.jellyfin.apiclient.interaction.EmptyResponse
 import org.jellyfin.apiclient.interaction.Response
+import org.jellyfin.apiclient.model.configuration.ServerConfiguration
 import org.jellyfin.apiclient.model.dto.UserItemDataDto
 import org.jellyfin.apiclient.model.session.PlaybackProgressInfo
 import org.jellyfin.apiclient.model.session.PlaybackStopInfo
@@ -19,6 +20,10 @@ val PRODUCT_NAME_SUPPORTED_SINCE: Pair<Int, Int> = 10 to 3
 // Can be removed/replaced once the api client supports coroutines natively
 suspend fun ApiClient.getPublicSystemInfo(): PublicSystemInfo? = suspendCoroutine { continuation ->
     GetPublicSystemInfoAsync(ContinuationResponse(continuation))
+}
+
+suspend fun ApiClient.getServerConfiguration(): ServerConfiguration? = suspendCoroutine { continuation ->
+    GetServerConfigurationAsync(ContinuationResponse(continuation))
 }
 
 suspend fun ApiClient.reportPlaybackProgress(progressInfo: PlaybackProgressInfo) = suspendCoroutine<Unit> { continuation ->
--- a/app/src/main/java/org/jellyfin/mobile/utils/LocaleUtils.kt
+++ b/app/src/main/java/org/jellyfin/mobile/utils/LocaleUtils.kt
@@ -1,0 +1,53 @@
+package org.jellyfin.mobile.utils
+
+import android.app.Activity
+import android.content.res.Configuration
+import android.os.Build
+import androidx.lifecycle.lifecycleScope
+import kotlinx.coroutines.launch
+import org.jellyfin.mobile.MainActivity
+import timber.log.Timber
+import java.util.*
+import kotlin.coroutines.resume
+import kotlin.coroutines.suspendCoroutine
+
+fun MainActivity.initLocale() = lifecycleScope.launch {
+    // Try to set locale via user settings
+    val userSettings = suspendCoroutine<String> { continuation ->
+        webView.evaluateJavascript("window.localStorage.getItem('${apiClient.currentUserId}-language')") { result ->
+            continuation.resume(result)
+        }
+    }
+    if (setLocale(userSettings.unescapeJson()))
+        return@launch
+
+    // Fallback to device locale
+    Timber.i("Couldn't acquire locale from config, keeping current")
+}
+
+private fun Activity.setLocale(localeString: String?): Boolean {
+    if (localeString.isNullOrEmpty() || localeString == "null")
+        return false
+
+    val localeSplit = localeString.split('-')
+    val locale = when (localeSplit.size) {
+        1 -> Locale(localeString, "")
+        2 -> Locale(localeSplit[0], localeSplit[1])
+        else -> return false
+    }
+
+    val configuration = resources.configuration
+    if (locale != configuration.primaryLocale) {
+        Locale.setDefault(locale)
+        configuration.setLocale(locale)
+        @Suppress("DEPRECATION")
+        resources.updateConfiguration(configuration, resources.displayMetrics)
+
+        Timber.i("Updated locale from web: '$locale'")
+    } // else: Locale is already applied
+    return true
+}
+
+@Suppress("DEPRECATION")
+private val Configuration.primaryLocale: Locale
+    get() = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) locales[0] else locale


app/src/main/java/org/jellyfin/mobile/MainActivity.kt
```diff
@@ -194,6 +196,55 @@ class MainActivity : AppCompatActivity(), WebViewController {
         if (connectionHelper.connected) webView.loadUrl(url)
     }
 
+    @Suppress("DEPRECATION")
+    private fun initLocale() {
+        runOnUiThread {
```
Maxr1998, line 201: The extra `runOnUiThread` shouldn't be necessary.

app/src/main/java/org/jellyfin/mobile/MainActivity.kt
```diff
@@ -194,6 +196,55 @@ class MainActivity : AppCompatActivity(), WebViewController {
...
+            webView.evaluateJavascript("window.localStorage.getItem('${apiClient.currentUserId}-language')") { result ->
+                if (!setLocale(result.unescapeJson(), "userSettings")) {
+                    GlobalScope.launch {
```
Maxr1998, line 204: Don't use GlobalScope, it should only be used in very rare circumstances.
```suggestion
                    lifecycleScope.launch {
```

app/src/main/java/org/jellyfin/mobile/settings/SettingsActivity.kt
```diff
@@ -18,6 +18,7 @@ class SettingsActivity : AppCompatActivity() {
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+        setTitle(R.string.activity_name_settings)
```
Maxr1998, line 21: This shouldn't be necessary, as it's already set in the manifest.

app/src/main/java/org/jellyfin/mobile/utils/ApiExtensions.kt
```diff
@@ -33,6 +34,10 @@ suspend fun ApiClient.markPlayed(itemId: String, userId: String): UserItemDataDt
     MarkPlayedAsync(itemId, userId, Date(), ContinuationResponse(continuation))
 }
 
+suspend fun ApiClient.getServerConfigurationAsync(): ServerConfiguration? = suspendCoroutine { continuation ->
```
Maxr1998, line 37: To be more in line with the other functions:
```suggestion
suspend fun ApiClient.getServerConfiguration(): ServerConfiguration? = suspendCoroutine { continuation ->
```

Also, you can move this below `getPublicSystemInfo`.

app/src/main/java/org/jellyfin/mobile/MainActivity.kt
```diff
@@ -194,6 +196,55 @@ class MainActivity : AppCompatActivity(), WebViewController {
...
+                val language = culture.split('-')
+                if (language.size == 2) {
+                    appPreferences.language = language[0]
+                    appPreferences.country = language[1]
+                }
+            } else {
+                appPreferences.language = culture
+                appPreferences.country = ""
+            }
+            Timber.d("initLocale [method: $method, language: '${appPreferences.language}', country: '${appPreferences.country}']")
```
Maxr1998, line 242:
```suggestion
            Timber.d("setLocale [method: $method, language: '${appPreferences.language}', country: '${appPreferences.country}']")
```

Maxr1998 (changes requested):
Thanks for the PR!
Apart from my review comments, could you maybe move those function to another file/class as well? Maybe even with dependency injection, scoped to the Activity. See the [Koin docs](https://start.insert-koin.io/#/getting-started/koin-for-android?id=extended-scope-api) for more info on that.

app/src/main/java/org/jellyfin/mobile/settings/SettingsActivity.kt
CarlosOlivo: Unfortunately it's necessary to load a new translation or something like that.

**Without setTtile**
image


**With setTitle**
image

Maxr1998:
Thanks for the changes! I'll review this later today (after sleeping).

app/src/main/java/org/jellyfin/mobile/utils/LocaleUtils.kt
```diff
@@ -0,0 +1,58 @@
...
+    val configuration = resources.configuration
+    val locale = Locale(
+        language ?: configuration.locale.language,
+        country ?: configuration.locale.country
+    )
+    Timber.d("loadLocale [class: ${localClassName}, language: '${locale.language}', country: '${locale.country}']")
+    Locale.setDefault(locale)
+    configuration.setLocale(locale)
+    resources.updateConfiguration(configuration, resources.displayMetrics)
+}
```
Maxr1998, line 38: Could you merge this function into the `setLocale` function? Then you could also keep the language/locale variables inside `setLocale`, and wouldn't require the global ones.

app/src/main/java/org/jellyfin/mobile/utils/LocaleUtils.kt
```diff
@@ -0,0 +1,58 @@
...
+
+private fun Activity.setLocale(culture: String?, method: String): Boolean {
```
Maxr1998, line 40:
```suggestion
private fun Activity.setLocale(localeString: String?, source: String): Boolean {
```

Maxr1998 (changes requested):

app/src/main/java/org/jellyfin/mobile/utils/LocaleUtils.kt
```diff
@@ -0,0 +1,53 @@
...
+}
+
+@Suppress("DEPRECATION")
+private fun Activity.setLocale(localeString: String?, source: String): Boolean {
+    var language: String? = null
+    var country: String? = null
+
+    if (localeString.isNullOrEmpty() || localeString == "null") {
+        return false
+    } else {
```
Maxr1998, line 31: I'd invert the if condition, to cover the success case first

app/src/main/java/org/jellyfin/mobile/utils/LocaleUtils.kt
```diff
@@ -0,0 +1,53 @@
...
+        if (localeString.contains('-')) {
+            val cultureSplit = localeString.split('-')
+            if (cultureSplit.size == 2) {
+                language = cultureSplit[0]
+                country = cultureSplit[1]
+            }
```
Maxr1998, line 37:
```suggestion
            } else {
                return false
            }
```

app/src/main/java/org/jellyfin/mobile/utils/LocaleUtils.kt
```diff
@@ -0,0 +1,53 @@
...
+        }
+        Timber.d("setLocale [source: $source, language: '${language}', country: '${country}']")
+        val configuration = resources.configuration
+        val locale = Locale(
+            language ?: configuration.locale.language,
+            country ?: configuration.locale.country
+        )
+        Locale.setDefault(locale)
+        configuration.setLocale(locale)
+        resources.updateConfiguration(configuration, resources.displayMetrics)
```
Maxr1998, line 50: Could you only update the configuration if there's actually a locale set, and maybe even check if the locales differ?

Maxr1998 (changes requested):

Maxr1998 (approved):

Maxr1998:
@CarlosOlivo did some more tweaks, please take another look at those and lmk if the are correct, then I'll merge.
@nielsvanvelzen would also appreciate a review from you :grin:

CarlosOlivo:
> @CarlosOlivo did some more tweaks, please take another look at those and lmk if they are ok, then I'll merge.
> @nielsvanvelzen would also appreciate a review from you 😁

Yep, it works fine, you just have to remove all the code that loads the locale from the API [serverSettings], I noticed after the next test that it is not necessary and even counterproductive.

1. Set client setting locale [userSettings] to Auto or clear App Data
2. Set android system locale to English (United States) [es_US]
3. Set jellyfin server locale [uiCulture] to Spanish (México) [es-MX]

With serverSettings code:

- 2020-09-21 00:43:40.491 3417-3417/org.jellyfin.mobile.debug D/LocaleUtilsKt: Updated locale from [serverSettings], locale: 'es_MX'
- webView: English (United States) [es_US]
- settingsActivity: Spanish (México) [es_MX]

Without serverSettings code:

- 2020-09-21 00:49:29.693 3422-3422/org.jellyfin.mobile.debug D/LocaleUtilsKt$initLocale: Couldn't acquire locale from config, keeping current
- webView: English (United States) [es_US]
- settingsActivity: English (United States) [es_US]

app/src/main/java/org/jellyfin/mobile/utils/LocaleUtils.kt
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    val configuration = resources.configuration
+    if (locale != configuration.primaryLocale) {
+        Locale.setDefault(locale)
+        configuration.setLocale(locale)
+        @Suppress("DEPRECATION")
+        resources.updateConfiguration(configuration, resources.displayMetrics)
+
+        Timber.d("Updated locale from [$source], locale: '$locale'")
```
nielsvanvelzen, line 50:
```suggestion
        Timber.i("Updated locale from [$source], locale: '$locale'")
```

nielsvanvelzen (changes requested):

Maxr1998:
@CarlosOlivo So I should just keep the user settings source for locale?

CarlosOlivo:
> @CarlosOlivo So I should just keep the user settings source for locale?

Yep, when no language is set in user settings the web UI fallback to system locale, not to uiCulture locale.

https://github.com/jellyfin/jellyfin-web/blob/7f8019d01f73a235edbd3e7872cf44310ce013e1/src/scripts/globalize.js#L39

https://github.com/jellyfin/jellyfin-web/blob/7f8019d01f73a235edbd3e7872cf44310ce013e1/src/scripts/globalize.js#L20

That is the default behavior.

Maxr1998 (approved):

app/src/main/java/org/jellyfin/mobile/utils/LocaleUtils.kt
nielsvanvelzen: This one isn't resolved.

Maxr1998: > This one isn't resolved.

Yikes, I only fixed the other one..

nielsvanvelzen (approved):

PR #118, opened 2020-09-11
Keep ApiClient models enumerations members

### LINKED ISSUE #116: v2.0.0-rc.4 crashes when connecting to a server
**Describe the bug**
Trying to enter with the new compilation of the Android version with the release version, it does not enter the server, the app crashes and closes itself. On the other hand, with the DEBUG version it does it perfectly. It will be an error when compiling the app. I hope it is a fixable error.

**To Reproduce**
1. Enter the IP
2. The screen goes black
3. The app closes by itself

 - App version: [2.0.0-rc.4]
 - Android version: [Android 9]
 - Device manufacturer and model: [LG K50s]
 - Jellyfin server version: [10.6.4]
 - The type of video player implementation used: [Integrated player (ExoPlayer)]

CarlosOlivo:
Closes #116 

See: https://stackoverflow.com/questions/15543607/assertionerror-in-gson-enumtypeadapter-when-using-proguard-obfuscation#comment112854882_30167048

### DIFF
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -19,6 +19,9 @@
     public *;
 }
 
+# Keep ApiClient models enumerations members
+-keepclassmembers enum org.jellyfin.apiclient.model.**.* { *; }
+
 # Keep file names/line numbers
 -keepattributes SourceFile,LineNumberTable
 


nielsvanvelzen:
I also identified R8 as the cause of the issue. But we should fix it in the android module of the apiclient, not the app.

CarlosOlivo:
Anyway I think the correct thing would be to launch a hotfix with a working release version v2.0.0-rc.4 unless you launch v2.0.0-rc.5 today.

nielsvanvelzen:
We should make a proper fix and that's not what this PR is (although I'm really grateful for your help!).
I guess we can have the fix released tomorrow (In my timezone the day is almost over) but that depends not only on me but also other team members that test, review and publish the app.

The 2.0.0 app is not stable yet and therefor not released on the main channels. It's only usable by sideloading of opting into the Google Play beta. If you join the beta you can expect issues like this to happen as we're not using it as "try new features early" but as an actual beta channel.

CarlosOlivo:
Yup, sorry if I sound rude, an ETA for tomorrow is very good, I said it because the previous rc3 was like 17 days ago, but if the rc5 is going to be launched when the proper fix are made in a few days I don't see the problem.

Also, feel free to close this PR.

nielsvanvelzen:
No problem. It's still a major issue so of course we want to fix it. The third release candidate worked fine so we didn't want to release the next one until everything we wanted to fix was included.

Thanks again for the fast response to the issue, I'll close this PR for now and will look into making a proper fix in the apiclient tomorrow.

PR #126, opened 2020-09-13
Implement auto discovery

### LINKED ISSUE #115: Implement auto discovery
**Describe the feature you'd like**
The old Android app was able to autodiscover servers on the local network. I'd like to have this in the new app so I don't need to type the server address manually.

**Additional context**
I think apiclient developers created a function for that. 😉

CarlosOlivo:
Button only visible when at least one server is found

Closes #115 👀

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -2,6 +2,7 @@
     xmlns:tools="http://schemas.android.com/tools"
     package="org.jellyfin.mobile">
 
+    <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.WAKE_LOCK" />
     <uses-permission android:name="android.permission.BLUETOOTH" />
--- a/app/src/main/java/org/jellyfin/mobile/webapp/ConnectionHelper.kt
+++ b/app/src/main/java/org/jellyfin/mobile/webapp/ConnectionHelper.kt
@@ -1,5 +1,6 @@
 package org.jellyfin.mobile.webapp
 
+import android.app.AlertDialog
 import android.view.KeyEvent
 import android.view.View
 import android.view.inputmethod.EditorInfo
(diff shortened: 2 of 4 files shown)


Maxr1998:
I love this, thank you so much! The only criticism I could spot in a quick review would probably be the standard AlertDialog with the newlines, but for now that's absolutely fine and in the future I may replace it with another UI anyways.
Gonna review it in detail later today, if I find the time.

app/src/main/AndroidManifest.xml
```diff
@@ -2,6 +2,7 @@
     xmlns:tools="http://schemas.android.com/tools"
     package="org.jellyfin.mobile">
 
+    <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
```
nielsvanvelzen, line 5: Was this permission required for discovery to work? If so I'll add it to the apiclient for the next release.

nielsvanvelzen (commented):
I don't think this is the correct approach in adding server discovery. In my opinion it should always start when the server selection screen is shown and display the results in the activity instead of using an dialog.

app/src/main/AndroidManifest.xml
CarlosOlivo: Yep, is required, also is already documented in the source code by ... you (?) https://github.com/jellyfin/jellyfin-apiclient-java/blame/1565f286a3b676cdf85093b0ee99c4b940b99660/android/src/main/kotlin/org/jellyfin/apiclient/discovery/AndroidBroadcastAddressesProvider.kt#L18

nielsvanvelzen: It's funny how I knew it was required but forgot to actually add it to the manifest lol

CarlosOlivo:
> I don't think this is the correct approach in adding server discovery. In my opinion it should always start when the server selection screen is shown and display the results in the activity instead of using an dialog.

Unfortunately for me the discovery of the cordova client never worked so, I don't know how it used to be the discovery UI, for me the UI is the login screen with a "Cancel" button that redirects to a no-op selectServer.html bundled in the web client (notice the height change in the title bar), but you are free to reimplement the UI to follow the original behavior / flow.

image
image

Maxr1998 (approved):

PR #127, opened 2020-09-14
External player changes

CarlosOlivo:
Some changes in external player

1. Fixes in time tracking which resets last playback position to zero when there was no valid external player selection and / or Activity / Thread is killed by Android System (EVENT_CANCELED)

2. Added a device profile that assumes all video and audio formats are supported on external players which allows to process the file / DirectStream as-is.
So even if the device does not report a compatible codec or the external player has SW codecs, it allows the playback. (like MX Player with Custom Codec [DTS, AC3, etc]
Side effect: Transcoding is officially no longer possible on external players. (Well, it never was supported from the beginning)

3. About the change in fun notifyEvent on ExternalPlayer.kt
This fix a very rare runtime exception that happens when Android kills Jellyfin thread in background when external player is in use for a long time and spawn another thread when onActivityResult is handled
I don't have the complete log, but I've been "testing" the change over the weekend and the exception has no longer appeared.

`2020-09-11 18:38:34.502 1020-22699/org.jellyfin.mobile.debug W/System.err: java.lang.RuntimeException: java.lang.Throwable: A WebView method was called on thread 'JavaBridge'. All WebView methods must be called on the same thread. (Expected Looper Looper (main, tid 2) {a1283a0} called on Looper (JavaBridge, tid 8129) {a7ac882}, FYI main Looper is Looper (main, tid 2) {a1283a0})`

4. Added some validations in notifyEvent to prevent code injection.

5. Use Android toast for messages.

### DIFF
--- a/app/src/main/assets/native/externalplayer.js
+++ b/app/src/main/assets/native/externalplayer.js
@@ -1,4 +1,4 @@
-define(['events', 'playbackManager', 'toast'], function (events, playbackManager, toast) {
+define(['appSettings', 'events', 'playbackManager'], function (appSettings, events, playbackManager) {
     "use strict";
 
     return function () {
@@ -43,7 +43,7 @@ define(['events', 'playbackManager', 'toast'], function (events, playbackManager
 
         self.play = function (options) {
             return new Promise(function (resolve) {
-                self._currentTime = 0;
+                self._currentTime = (options.playerStartPositionTicks || 0) / 10000;
                 self._paused = false;
                 self._currentSrc = options.url;
                 window.ExternalPlayer.initPlayer(JSON.stringify(options));
@@ -128,15 +128,14 @@ define(['events', 'playbackManager', 'toast'], function (events, playbackManager
             });
         };
 
-        self.notifyCanceled = function (message) {
+        self.notifyCanceled = function () {
             // required to not mark an item as seen / completed without time changes
-            var currentTime = self._currentTime || 0;
-            self.notifyTimeUpdate(currentTime + 1);
+            let currentTime = self._currentTime || 0;
             self.notifyTimeUpdate(currentTime - 1);
-            self.notifyEnded();
-            if (message) {
-                toast(message);
+            if (currentTime > 0) {
+                self.notifyTimeUpdate(currentTime);
             }
+            self.notifyEnded();
         };
 
         self.currentTime = function () {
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/jellyfin/mobile/bridge/ExternalPlayer.kt
```diff
@@ -48,7 +48,9 @@ class ExternalPlayer(private val activity: MainActivity) : KoinComponent {
     }
 
     private fun notifyEvent(event: String, parameters: String = "") {
-        activity.loadUrl("javascript:window.ExtPlayer.notify$event($parameters)")
+        activity.runOnUiThread {
+            activity.loadUrl("javascript:window.ExtPlayer.notify$event($parameters)")
```
nielsvanvelzen, line 52: We might want to change this to use JSON or something else because right now we can use an XSS attack from within the Kotlin code.

Given the event `(); alert('example'); noop(` it will fail

Maxr1998: @nielsvanvelzen the same already happens in the ExoPlayer component however. The question is whether we can ignore it, because the function is private and only called by us anyway, or if we should add some sanitization, like checking is every char is a letter.

app/src/main/assets/native/externalplayer.js
```diff
@@ -130,9 +130,11 @@ define(['events', 'playbackManager', 'toast'], function (events, playbackManager
 
         self.notifyCanceled = function (message) {
             // required to not mark an item as seen / completed without time changes
-            var currentTime = self._currentTime || 0;
-            self.notifyTimeUpdate(currentTime + 1);
+            let currentTime = self._currentTime || 0;
             self.notifyTimeUpdate(currentTime - 1);
+            if (currentTime > 0) {
+                self.notifyTimeUpdate(currentTime);
+            }
```
Maxr1998, line 137: Doesn't this basically notify twice? Also, the first notify could push a `-1`, can't this cause issues?

app/src/main/java/org/jellyfin/mobile/bridge/ExternalPlayer.kt
nielsvanvelzen: Better be safe then sorry

CarlosOlivo: Yep, the same can happen in https://github.com/jellyfin/jellyfin-android/blob/39accd32e0b6dd0b3830e62414e0f4008dff966e/app/src/main/java/org/jellyfin/mobile/MainActivity.kt#L119

```
Function
    loadUrl(String)
Found usages  (6 usages found)
    Function call  (6 usages found)
        app  (6 usages found)
            org.jellyfin.mobile  (1 usage found)
                MainActivity.kt  (1 usage found)
                    MainActivity  (1 usage found)
                        onCreate  (1 usage found)
                            119 loadUrl("javascript:$function")
            org.jellyfin.mobile.bridge  (3 usages found)
                Commands.kt  (1 usage found)
                    Commands  (1 usage found)
                        triggerInputManagerAction  (1 usage found)
                            13 loadUrl(buildInputManagerCommand("trigger('$action')"))
                ExternalPlayer.kt  (1 usage found)
                    ExternalPlayer  (1 usage found)
                        notifyEvent  (1 usage found)
                            54 activity.loadUrl("javascript:window.ExtPlayer.notify$event($parameters)")
                NativeInterface.kt  (1 usage found)
                    NativeInterface  (1 usage found)
                        execCast  (1 usage found)
                            160 activity.loadUrl("""javascript:window.NativeShell.castCallback("$action", $keep, $err, $result);""")
            org.jellyfin.mobile.webapp  (2 usages found)
                RemotePlayerService.kt  (2 usages found)
                    RemotePlayerService  (2 usages found)
                        ServiceBinder  (2 usages found)
                            436 webViewController?.loadUrl(Commands.buildInputManagerCommand("trigger('seek', $pos)"))
                            440 webViewController?.loadUrl(Commands.buildPlaybackManagerCommand("sendCommand({Name:'SetVolume', Arguments:{Volume:$value}})"))
```

for now I added some validations inside the function

app/src/main/assets/native/externalplayer.js
CarlosOlivo: > Doesn't this basically notify twice? 

That's the point.
`// required to not mark an item as seen / completed without time changes`

Examples:

Case 1 - User start playback with media that has playerStartPositionTicks > 0 , external player not chosen
```
S1:E1  - playerStartPositionTicks : 60000
Android emits EVENT_CANCELED
currentTime = 6
notifyTimeUpdate(5);
notifyTimeUpdate(6);
notifyEnded()
Jellyfin report playback stopped by app
S1:E1  - playerStartPositionTicks: 60000
```

Case 2 - User start playback with media that has playerStartPositionTicks > 0 , external player not chosen, without time changes
```
S1:E1  - playerStartPositionTicks : 60000
Android emits EVENT_CANCELED
currentTime = 6
notifyEnded()
Jellyfin report playback stopped by app without time tracking, mark as watched
S1:E1  - playerStartPositionTicks: RunTimeTicks
```

Case 3 - User start playback with media that has playerStartPositionTicks > 0 , external player not chosen, with current time as time change
```
S1:E1  - playerStartPositionTicks : 60000
Android emits EVENT_CANCELED
currentTime = 6
notifyTimeUpdate(6);
notifyEnded()
Jellyfin report playback stopped by app without time tracking, mark as watched
S1:E1  - playerStartPositionTicks: RunTimeTicks
```

> Also, the first notify could push a `-1`, can't this cause issues?

Case 4 - User start playback with media that has playerStartPositionTicks = 0, external player not chosen
```
S1:E1  - playerStartPositionTicks: 0
Android emits EVENT_CANCELED
currentTime = 0
notifyTimeUpdate(-1);
notifyEnded()
Jellyfin server aborts playback, invalid time
S1:E1  - playerStartPositionTicks: 0
```

Case 5 - User start playback with media that has playerStartPositionTicks = 0, external player not chosen, without time changes
```
S1:E1  - playerStartPositionTicks : 0
Android emits EVENT_CANCELED
currentTime = 0
notifyEnded()
Jellyfin report playback stopped by app without time tracking, mark as watched
S1:E1  - playerStartPositionTicks: RunTimeTicks
```

Maxr1998: I see, thanks for the info. That makes sense then.

app/src/main/java/org/jellyfin/mobile/bridge/ExternalPlayer.kt
Maxr1998: That would be something that I have to look into at a later time then..

Maxr1998 (approved):

PR #132, opened 2020-09-17
Fixed controller hiding when seeking in native player

yashx:
### Before
old

### Now
new

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/player/PlayerActivity.kt
+++ b/app/src/main/java/org/jellyfin/mobile/player/PlayerActivity.kt
@@ -51,6 +51,13 @@ class PlayerActivity : AppCompatActivity() {
      */
     private val hideUnlockButtonAction = Runnable { unlockScreenButton.isVisible = false }
 
+    /**
+     * Runnable that hides [playerView] controller
+     */
+    private val hidePlayerViewControllerAction = Runnable {
+        playerView.hideController()
+    }
+
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_player)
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/jellyfin/mobile/player/PlayerActivity.kt
```diff
@@ -51,6 +51,12 @@ class PlayerActivity : AppCompatActivity() {
      */
     private val hideUnlockButtonAction = Runnable { unlockScreenButton.isVisible = false }
 
+    /**
+     * Runnable that hides [playerView] controller
+     */
+    private val hidePlayerViewControllerAction = Runnable { playerView.hideController() }
```
nielsvanvelzen, line 57:
```suggestion
    private val hidePlayerViewControllerAction = Runnable {
        playerView.hideController()
    }
```

app/src/main/java/org/jellyfin/mobile/player/PlayerActivity.kt
```diff
@@ -51,6 +51,12 @@ class PlayerActivity : AppCompatActivity() {
...
+
+
```
nielsvanvelzen, line 59:
```suggestion

```

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/mobile/player/PlayerActivity.kt
Maxr1998: I'd say that's fine, doesn't really need extra wrapping imo. The line above also doesn't do that ^^

app/src/main/java/org/jellyfin/mobile/player/PlayerActivity.kt
nielsvanvelzen: I should've been more clear on this one; there were two empty lines and it should've been just one. You now removed both so you need to add one back.

yashx: Oh. I didn't see what this was for. I just committed all the suggestions on Github and squashed the commit.

nielsvanvelzen (approved):

Maxr1998 (approved):

PR #133, opened 2020-09-18
Added Native Player Gesture for Volume & Brightness 

### LINKED ISSUE #13: Add gesture controls for volume and brightness

yashx:
closes #13
### How it looks
gesture2

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/player/PlayerActivity.kt
+++ b/app/src/main/java/org/jellyfin/mobile/player/PlayerActivity.kt
@@ -2,14 +2,16 @@ package org.jellyfin.mobile.player
 
 import android.annotation.SuppressLint
 import android.app.PictureInPictureParams
+import android.content.Context
 import android.content.Intent
 import android.content.pm.ActivityInfo
 import android.content.res.Configuration
+import android.media.AudioManager
 import android.os.Build
 import android.os.Bundle
+import android.provider.Settings.System
 import android.view.*
-import android.widget.ImageButton
-import android.widget.TextView
+import android.widget.*
 import androidx.activity.viewModels
 import androidx.appcompat.app.AppCompatActivity
 import androidx.core.view.ViewCompat
@@ -18,14 +20,19 @@ import androidx.core.view.postDelayed
 import androidx.core.view.updatePadding
 import com.google.android.exoplayer2.Player
 import com.google.android.exoplayer2.ui.PlayerView
+import org.jellyfin.mobile.AppPreferences
 import org.jellyfin.mobile.R
 import org.jellyfin.mobile.utils.*
+import org.jellyfin.mobile.utils.Constants.DEFAULT_CENTER_OVERLAY_TIMEOUT_MS
 import org.jellyfin.mobile.utils.Constants.DEFAULT_CONTROLS_TIMEOUT_MS
 import org.jellyfin.mobile.utils.Constants.DEFAULT_SEEK_TIME_MS
+import org.koin.android.ext.android.inject
+import kotlin.math.abs
 
 
 class PlayerActivity : AppCompatActivity() {
 
+    private val appPreferences: AppPreferences by inject()
     private val viewModel: PlayerViewModel by viewModels()
     private val playerView: PlayerView by lazyView(R.id.player_view)
     private val playerControlsView: View by lazyView(R.id.player_controls)
@@ -34,7 +41,21 @@ class PlayerActivity : AppCompatActivity() {
     private val titleTextView: TextView by lazyView(R.id.track_title)
     private val fullscreenSwitcher: ImageButton by lazyView(R.id.fullscreen_switcher)
     private val unlockScreenButton: ImageButton by lazyView(R.id.unlock_screen_button)
+    private val gestureIndicatorOverlayLayout: LinearLayout by lazyView(R.id.gesture_overlay_layout)
+    private val gestureIndicatorOverlayImage: ImageView by lazyView(R.id.gesture_overlay_image)
+    private val gestureIndicatorOverlayProgress: ProgressBar by lazyView(R.id.gesture_overlay_progress)
     private lateinit var playbackMenus: PlaybackMenus
+    private val audioManager: AudioManager by lazy { (getSystemService(Context.AUDIO_SERVICE) as AudioManager) }
+
+    private val swipeGesturesEnabled
+        get() = appPreferences.exoPlayerAllowSwipeGestures
+
+    /**
+     * Tracks a value during a swipe gesture (between multiple onScroll calls).
+     * When the gesture starts it's reset to an initial value and gets increased or decreased
+     * (depending on the direction) as the gesture progresses.
+     */
+    private var swipeGestureValueTracker = -1f
 
     /**
      * Listener that watches the current device orientation.
@@ -56,6 +77,13 @@ class PlayerActivity : AppCompatActivity() {
      */
     private val hidePlayerViewControllerAction = Runnable {
         playerView.hideController()
+    }
+
+    /**
+     * Runnable that hides [gestureIndicatorOverlayLayout]
+     */
+    private val hideGestureIndicatorOverlayAction = Runnable {
+        gestureIndicatorOverlayLayout.isVisible = false
     }
 
     override fun onCreate(savedInstanceState: Bundle?) {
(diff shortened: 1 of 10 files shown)


nielsvanvelzen:
Can we add an option to disable this behavior?

app/src/main/java/org/jellyfin/mobile/player/PlayerActivity.kt
```diff
@@ -232,9 +247,77 @@ class PlayerActivity : AppCompatActivity() {
...
+                    overlayBrightnessVolumeProgressBar.max = maxVolume
+                    overlayBrightnessVolumeProgressBar.progress = toSet
+                }
+                else{
+                    // On left so changing brightness
+                    val windowLayoutParams = window.attributes
+                    if(counter == -1f){
+                        counter = windowLayoutParams.screenBrightness
+                        if(counter < 0f)
+                            counter = android.provider.Settings.System.getFloat(
```
yashx, line 292: @Maxr1998 @nielsvanvelzen This line has a few problems. According to the docs, it should give value between 0 and 255 but on my phone (which has MIUI) it is goes over 255.
Worst case scenario, it jumps to full brightness first time the gesture is used.
I can disable it but imo the utility it offers is worth dealing with the bug.

yashx:
> Can we add an option to disable this behavior?
Should I do this in this PR or send another one?

dkanada:
Within this one should be fine.

yashx:
@nielsvanvelzen Added option to disable it in "Client Settings"

app/src/main/java/org/jellyfin/mobile/player/PlayerActivity.kt
```diff
@@ -232,9 +253,80 @@ class PlayerActivity : AppCompatActivity() {
...
+                if (e1 == null || isBrightnessVolumeGestureAllowed.not())
+                    return false
+
+                if (abs(distanceY / distanceX) < 2)
+                    return true
+
+                val viewCenterX = playerView.measuredWidth / 2
+
+                // Distance to swipe to go from min to max
+                val distanceFull = playerView.measuredHeight
```
Maxr1998, line 272: I think 2/3 would be enough for that..

app/src/main/java/org/jellyfin/mobile/player/PlayerActivity.kt
```diff
@@ -232,9 +253,80 @@ class PlayerActivity : AppCompatActivity() {
...
+
```
Maxr1998, line 273:
```suggestion
```

app/src/main/java/org/jellyfin/mobile/player/PlayerActivity.kt
```diff
@@ -232,9 +253,80 @@ class PlayerActivity : AppCompatActivity() {
...
+                    overlayBrightnessVolumeImage.setImageResource(R.drawable.ic_volume_white_84dp)
+                    overlayBrightnessVolumeProgressBar.max = maxVolume
+                    overlayBrightnessVolumeProgressBar.progress = toSet
+                } else {
+                    // On left so changing brightness
+                    val windowLayoutParams = window.attributes
+                    if (counter == -1f) {
+                        counter = windowLayoutParams.screenBrightness
+                        if (counter < 0f)
+                            counter = android.provider.Settings.System.getFloat(
```
Maxr1998, line 297: Please add an import (alias) for either Settings or System, if possible

app/src/main/java/org/jellyfin/mobile/player/PlayerActivity.kt
```diff
@@ -34,7 +39,16 @@ class PlayerActivity : AppCompatActivity() {
     private val titleTextView: TextView by lazyView(R.id.track_title)
     private val fullscreenSwitcher: ImageButton by lazyView(R.id.fullscreen_switcher)
     private val unlockScreenButton: ImageButton by lazyView(R.id.unlock_screen_button)
+    private val overlayBrightnessVolumeLayout: LinearLayout by lazyView(R.id.overlay_b_v_layout)
+    private val overlayBrightnessVolumeImage: ImageView by lazyView(R.id.overlay_b_v_image)
+    private val overlayBrightnessVolumeProgressBar: ProgressBar by lazyView(R.id.overlay_b_v_progress_bar)
     private lateinit var playbackMenus: PlaybackMenus
+    private val audioManager: AudioManager by lazy { (getSystemService(Context.AUDIO_SERVICE) as AudioManager) }
+    private var counter = -1f
```
Maxr1998, line 47: This should be a bit more descriptive

app/src/main/res/drawable/ic_brightness_white_84dp.xml
```diff
@@ -0,0 +1,5 @@
+<vector android:height="84dp" android:tint="#FFFFFF"
+    android:viewportHeight="24" android:viewportWidth="24"
+    android:width="84dp" xmlns:android="http://schemas.android.com/apk/res/android">
+    <path android:fillColor="@android:color/white" android:pathData="M20,8.69L20,4h-4.69L12,0.69 8.69,4L4,4v4.69L0.69,12 4,15.31L4,20h4.69L12,23.31 15.31,20L20,20v-4.69L23.31,12 20,8.69zM12,18c-3.31,0 -6,-2.69 -6,-6s2.69,-6 6,-6 6,2.69 6,6 -2.69,6 -6,6zM12,8c-2.21,0 -4,1.79 -4,4s1.79,4 4,4 4,-1.79 4,-4 -1.79,-4 -4,-4z"/>
+</vector>
```
Maxr1998, line 5: Please reformat the file in Android Studio

app/src/main/res/drawable/ic_volume_white_84dp.xml
```diff
@@ -0,0 +1,5 @@
+<vector android:height="84dp" android:tint="#FFFFFF"
+    android:viewportHeight="24" android:viewportWidth="24"
+    android:width="84dp" xmlns:android="http://schemas.android.com/apk/res/android">
+    <path android:fillColor="@android:color/white" android:pathData="M3,9v6h4l5,5L12,4L7,9L3,9zM16.5,12c0,-1.77 -1.02,-3.29 -2.5,-4.03v8.05c1.48,-0.73 2.5,-2.25 2.5,-4.02zM14,3.23v2.06c2.89,0.86 5,3.54 5,6.71s-2.11,5.85 -5,6.71v2.06c4.01,-0.91 7,-4.49 7,-8.77s-2.99,-7.86 -7,-8.77z"/>
+</vector>
```
Maxr1998, line 5: Please reformat the file in Android Studio

app/src/main/res/layout/activity_player.xml
```diff
@@ -16,6 +16,11 @@
         android:layout_width="match_parent"
         android:layout_height="match_parent">
 
+        <include
+            layout="@layout/overlay_brightness_volume"
+            android:visibility="gone"
+            tools:visibility="visible"/>
```
Maxr1998, line 22:
```suggestion
            tools:visibility="visible" />
```

Maxr1998 (changes requested):
Code looks mostly good, I have a few comments about naming and formatting stuff however. I'd generally use a shorter name for the gesture feature.
Also, I'm not a fan of that `counter` attribute.. I'm not sure how you could do it differently though, I think I'll have to play with the code myself once the other mentioned issues are fixed.

yashx:
@Maxr1998 While brightness gesture can be implemented without the counter. The problem is with the volume gesture. I can also set volume by using the distance between  `e1` and `e2` in `onScroll` but I would need to store the initial volume level when the gesture was started.
So I decided to go with the `counter` approach so that all gestures  (brightness, volume and horizontal seek(future PR)) can have a similar implementation.

Maxr1998:
I see, then there's probably no way around that. Could you still make the counter variable a bit more descriptive then, and add some documentation/comments on how you're using it?

yashx:
Yes I will do that and all the requested changes in a few hours.

app/src/main/java/org/jellyfin/mobile/player/PlayerActivity.kt
```diff
@@ -232,9 +261,72 @@ class PlayerActivity : AppCompatActivity() {
                 }
                 return true
             }
+
+            override fun onScroll(e1: MotionEvent?, e2: MotionEvent?, distanceX: Float, distanceY: Float): Boolean {
+                if (e1 == null || !isBrightnessVolumeGestureAllowed)
+                    return false
```
nielsvanvelzen, line 267:
```suggestion
                e1 as? isBrightnessVolumeGestureAllowed ?: return false
```

app/src/main/java/org/jellyfin/mobile/player/PlayerActivity.kt
```diff
@@ -232,9 +261,72 @@ class PlayerActivity : AppCompatActivity() {
                 }
                 return true
             }
+
+            override fun onScroll(e1: MotionEvent?, e2: MotionEvent?, distanceX: Float, distanceY: Float): Boolean {
```
nielsvanvelzen, line 265: Please change `e1` and `e2` to more descriptive names.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/mobile/player/PlayerActivity.kt
yashx: Uhm isn't this code incorrect? I am new to kotlin so correct me if I am wrong but why should I cast a MotionEvent object to a Boolean object?

image

app/src/main/java/org/jellyfin/mobile/player/PlayerActivity.kt
yashx: `e1` and `e2` are the names used in the documentation of `onScroll` method. Are you sure I should change them?

image

app/src/main/java/org/jellyfin/mobile/player/PlayerActivity.kt
nielsvanvelzen: Sorry I read the code wrong due too GitHub's syntax highlighting. Ignore this one!

app/src/main/java/org/jellyfin/mobile/player/PlayerActivity.kt
nielsvanvelzen: I think something like `firstMotionEvent` and `secondMotionEvent` is more clear. @Maxr1998 what do you think?

Maxr1998: Maybe `first` and `current`? Imo no need to state the type in the variable. And even if, then only append `Event`, not the full name.

nielsvanvelzen (approved):

Maxr1998:
@yashx did some more tweaks and code style improvements, please let me know if it's ok for you, and if you @nielsvanvelzen also don't have any comments anymore, I'll squash and merge it.

nielsvanvelzen (approved):

Maxr1998 (approved):

PR #143, opened 2020-09-24
Add android auto support

Spacetech:
This change adds basic android auto support. A video of it in action for reference: https://youtu.be/Zl-ETVcv5BQ

Main changes:
- Implements a `MediaBrowserServiceCompat` which lets Android auto detect Jellyfin as a music player
- Supports library selection and drilling down into the library by different views (latest, albums, artists, songs, genres)
- Supports shuffling in the album view
- Supports prev / next tracks
- Shows thumbnails when possible

Notes:
- The `MediaService` (implementation of `MediaBrowserServiceCompat`) is a fairly standalone thing. Due to how android auto music players work, it will not rely on the main webview for ui. Instead it will use the `ApiClient` to load data from Jellyfin.
- Since it's a separate service and could/will launch independently from the main app, I added preferences for the server url, user id, and access token. Those preferences will be updated when the main app is launched. This makes it so the last used server will be used in android auto. Note: Saving the access token should be safe since the preferences are marked as private. 
- I named the folder `car` because that's more inline with what Google calls it - [Android for Cars](https://developer.android.com/training/cars)
- I based the url I use to load the audio track in the `createMediaSource` method on what the web client did. Is there a better way to obtain this url? (is there an api property for it?)

I see this as basic support that should be good enough to merge for now. There's certainly room for improvement. Here's a few ideas:
- Add shuffle button to more views
- Most of the lists are currently limited to 100 items. We need to figure out how to display more things in the android auto music ui. Some other music player apps list all the artist initials first and you slowly drill down to find what you're looking for. This has the advantage of supporting a large number of artists/albums. The disadvantage is that it requires a lot more user interaction to pick stuff out
- Better integration with the main app - Something like auto resuming whatever was last played would be great
- Add search support - Because clicking through large libraries to find a specific item is not fun
- Support for some sort of offline mode

This was my first time using Kotlin so let me know if I did anything weird. Thanks

References:
- Building media apps for cars - https://developer.android.com/training/cars/media
- uamp sample media app - https://github.com/android/uamp

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -100,6 +100,10 @@ dependencies {
     implementation(Dependencies.Cast.playServicesCast)
     implementation(Dependencies.Cast.playServicesCastFramework)
 
+    // Media
+    implementation(Dependencies.Media.media)
+    implementation(Dependencies.Media.exoPlayerMediaSession)
+
     // Health
     implementation(Dependencies.Health.timber)
     debugImplementation(Dependencies.Health.leakCanary)
--- a/app/src/main/java/org/jellyfin/mobile/AppPreferences.kt
+++ b/app/src/main/java/org/jellyfin/mobile/AppPreferences.kt
@@ -19,6 +19,22 @@ class AppPreferences(context: Context) {
             }
         }
 
+    var instanceAccessToken: String?
+        get() = sharedPreferences.getString(Constants.PREF_INSTANCE_ACCESS_TOKEN, null)
+        set(value) {
+            sharedPreferences.edit {
+                if (value != null) putString(Constants.PREF_INSTANCE_ACCESS_TOKEN, value) else remove(Constants.PREF_INSTANCE_ACCESS_TOKEN)
+            }
+        }
+
+    var instanceUserId: String?
+        get() = sharedPreferences.getString(Constants.PREF_INSTANCE_USER_ID, null)
+        set(value) {
+            sharedPreferences.edit {
+                if (value != null) putString(Constants.PREF_INSTANCE_USER_ID, value) else remove(Constants.PREF_INSTANCE_USER_ID)
+            }
+        }
+
     var ignoreBatteryOptimizations: Boolean
         get() = sharedPreferences.getBoolean(Constants.PREF_IGNORE_BATTERY_OPTIMIZATIONS, false)
         set(value) {
--- a/app/src/main/java/org/jellyfin/mobile/media/MediaService.kt
+++ b/app/src/main/java/org/jellyfin/mobile/media/MediaService.kt
@@ -1,0 +1,566 @@
+package org.jellyfin.mobile.media
+
+import android.net.Uri
+import android.os.Bundle
+import android.os.ResultReceiver
+import android.support.v4.media.MediaBrowserCompat.MediaItem
+import android.support.v4.media.MediaDescriptionCompat
+import android.support.v4.media.MediaMetadataCompat
+import android.support.v4.media.session.MediaControllerCompat
+import android.support.v4.media.session.MediaSessionCompat
+import android.support.v4.media.session.PlaybackStateCompat
+import androidx.media.MediaBrowserServiceCompat
+import com.google.android.exoplayer2.ControlDispatcher
+import com.google.android.exoplayer2.ExoPlayer
+import com.google.android.exoplayer2.Player
+import com.google.android.exoplayer2.SimpleExoPlayer
+import com.google.android.exoplayer2.ext.mediasession.MediaSessionConnector
+import com.google.android.exoplayer2.ext.mediasession.TimelineQueueNavigator
+import kotlinx.coroutines.*
+import org.jellyfin.apiclient.interaction.ApiClient
+import org.jellyfin.apiclient.model.dto.BaseItemDto
+import org.jellyfin.apiclient.model.dto.BaseItemType
+import org.jellyfin.apiclient.model.dto.ImageOptions
+import org.jellyfin.apiclient.model.entities.CollectionType
+import org.jellyfin.apiclient.model.entities.ImageType
+import org.jellyfin.apiclient.model.entities.SortOrder
+import org.jellyfin.apiclient.model.playlists.PlaylistItemQuery
+import org.jellyfin.apiclient.model.querying.*
+import org.jellyfin.mobile.AppPreferences
+import org.jellyfin.mobile.R
+import org.jellyfin.mobile.utils.*
+import org.koin.android.ext.android.inject
+import java.util.*
+import com.google.android.exoplayer2.MediaItem as ExoPlayerMediaItem
+
+class MediaService : MediaBrowserServiceCompat() {
+
+    private val appPreferences: AppPreferences by inject()
+    private val apiClient: ApiClient by inject()
+
+    private val ioScope = CoroutineScope(Dispatchers.IO + Job())
+
+    private lateinit var mediaController: MediaControllerCompat
+    private lateinit var mediaSession: MediaSessionCompat
+    private lateinit var mediaSessionConnector: MediaSessionConnector
+
+    private var currentPlaylistItems: MutableList<MediaDescriptionCompat> = mutableListOf()
+
+    private val exoPlayer: SimpleExoPlayer by lazy {
+        SimpleExoPlayer.Builder(this).build()
+    }
+
+    /**
+     * List of different views when browsing media
+     * Libraries is the initial view
+     */
+    private enum class MediaItemType {
+        Libraries,
+        Library,
+        LibraryLatest,
+        LibraryAlbums,
+        LibraryArtists,
+        LibrarySongs,
+        LibraryGenres,
+        LibraryPlaylists,
+        Album,
+        Artist,
+        Shuffle,
+        Playlist,
+    }
+
+    override fun onCreate() {
+        super.onCreate()
+
+        mediaSession = MediaSessionCompat(this, "MediaService").apply {
+            isActive = true
+        }
+
+        sessionToken = mediaSession.sessionToken
+
+        mediaController = MediaControllerCompat(this, mediaSession)
+
+        mediaSessionConnector = MediaSessionConnector(mediaSession).apply {
+            setPlayer(exoPlayer)
+            setPlaybackPreparer(MediaPlaybackPreparer(exoPlayer, appPreferences))
+            setQueueNavigator(MediaQueueNavigator(mediaSession))
+        }
+    }
+
+    override fun onDestroy() {
+        mediaSession.run {
+            isActive = false
+            release()
+        }
+        exoPlayer.release()
+        ioScope.cancel()
+    }
+
+    override fun onGetRoot(
+        clientPackageName: String,
+        clientUid: Int,
+        rootHints: Bundle?
+    ): BrowserRoot? {
+        val rootExtras = Bundle().apply {
+            putBoolean(CONTENT_STYLE_SUPPORTED, true)
+            putInt(CONTENT_STYLE_BROWSABLE_HINT, CONTENT_STYLE_LIST_ITEM_HINT_VALUE)
+            putInt(CONTENT_STYLE_PLAYABLE_HINT, CONTENT_STYLE_LIST_ITEM_HINT_VALUE)
+        }
+        return BrowserRoot(MediaItemType.Libraries.toString(), rootExtras)
+    }
+
+    /**
+     * The parent id will be in various formats such as
+     * Libraries
+     * Library|{id}
+     * LibraryAlbums|{id}
+     * Album|{id}
+     */
+    override fun onLoadChildren(parentId: String, result: Result<MutableList<MediaItem>>) {
+        if (appPreferences.instanceUrl == null || appPreferences.instanceAccessToken == null || appPreferences.instanceUserId == null) {
+            // the required properties for calling the api client are not available
+            // this should only occur if the user does not have a server connection in the app
+            result.sendError(null)
+            return
+        }
+
+        // ensure the api client is up to date
+        apiClient.ChangeServerLocation(appPreferences.instanceUrl)
+        apiClient.SetAuthenticationInfo(
+            appPreferences.instanceAccessToken,
+            appPreferences.instanceUserId
+        )
+
+        result.detach()
+
+        ioScope.launch {
+            loadView(parentId, result)
+        }
+    }
+
+    private suspend fun loadView(parentId: String, result: Result<MutableList<MediaItem>>) {
+        when (parentId) {
+            /**
+             * View that shows all the music libraries
+             */
+            MediaItemType.Libraries.toString() -> {
+                val response = apiClient.getUserViews(appPreferences.instanceUserId!!)
+                if (response != null) {
+                    val views = response.items
+                        .filter { item -> item.collectionType == CollectionType.Music }
+                        .map { item ->
+                            val itemImageUrl = apiClient.GetImageUrl(item, ImageOptions().apply {
+                                imageType = ImageType.Primary
+                                maxWidth = 1080
+                                quality = 90
+                            })
+
+                            val description: MediaDescriptionCompat =
+                                MediaDescriptionCompat.Builder().apply {
+                                    setMediaId(MediaItemType.Library.toString() + "|" + item.id)
+                                    setTitle(item.name)
+                                    setIconUri(Uri.parse(itemImageUrl))
+                                }.build()
+
+                            MediaItem(description, MediaItem.FLAG_BROWSABLE)
+                        }
+
+                    result.sendResult(views.toMutableList())
+
+                } else {
+                    result.sendError(null)
+                }
+            }
+
+            /**
+             * Views that show albums, artists, songs, genres, etc for a specific library
+             */
+            else -> {
+                val mediaItemTypeSplit = parentId.split("|")
+                val primaryType = mediaItemTypeSplit[0]
+                val primaryItemId = mediaItemTypeSplit[1]
+                val secondaryItemId =
+                    if (mediaItemTypeSplit.count() > 2) mediaItemTypeSplit[2] else null
+
+                if (primaryType == MediaItemType.Library.toString()) {
+                    /**
+                     * The default view for a library that lists various ways to browse
+                     */
+                    val libraryViews = arrayOf(
+                        Pair(MediaItemType.LibraryLatest, R.string.mediaservice_library_latest),
+                        Pair(MediaItemType.LibraryAlbums, R.string.mediaservice_library_albums),
+                        Pair(
+                            MediaItemType.LibraryArtists,
+                            R.string.mediaservice_library_artists
+                        ),
+                        Pair(MediaItemType.LibrarySongs, R.string.mediaservice_library_songs),
+                        Pair(MediaItemType.LibraryGenres, R.string.mediaservice_library_genres),
+                        Pair(
+                            MediaItemType.LibraryPlaylists,
+                            R.string.mediaservice_library_playlists
+                        ),
+                    )
+
+                    val views = libraryViews.map { item ->
+                        val description: MediaDescriptionCompat =
+                            MediaDescriptionCompat.Builder().apply {
+                                setMediaId(item.first.toString() + "|" + primaryItemId)
+                                setTitle(getString(item.second))
+                            }.build()
+
+                        MediaItem(description, MediaItem.FLAG_BROWSABLE)
+                    }
+
+                    result.sendResult(views.toMutableList())
+                    return
+                }
+
+                /**
+                 * Processes items from api responses
+                 * Updates the current play list items and sends results back to android auto
+                 */
+                fun processItems(items: Array<BaseItemDto>) {
+                    val mediaItems = items
+                        .map { item -> createMediaItem(primaryType, primaryItemId, item) }
+                        .toMutableList()
+
+                    currentPlaylistItems = mediaItems
+                        .filter { item -> item.isPlayable }
+                        .map { item -> item.description }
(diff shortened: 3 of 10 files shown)


nielsvanvelzen:
Before I start reviewing I'd like to point out a few things:
- We will not include this in the initial 2.0.0 release. (But probably 2.1 as "main feature")
- We need to look into the publishing guidelines for Android Auto, I know Google is quite strict on those
- Since none of our Android contributors have Android Auto we won't be able to test it on real hardware, I think the emulator should work sufficiently though
This means that the PR might be open a while after approval before we merge it.

And lastly, thanks for this contribution! I'm really excited to see our efforts for the app rewrite are enabling features like this.

neopc10:
Cool! Does this have playlist support?

Spacetech:
@nielsvanvelzen That all sounds good. For reference, I have been doing the majority of testing on my phone using the [Android Auto for phones app](https://play.google.com/store/apps/details?id=com.google.android.projection.gearhead.phonescreen&hl=en_US)


@neopc10 Good idea. I just added that.

Going forward I don't want to add too much more to this pr, otherwise it may become harder to review.

app/src/main/java/org/jellyfin/mobile/car/MediaService.kt
```diff
@@ -0,0 +1,609 @@
...
+                            }
+
+                            val views = response.items
+                                .filter { item -> item.collectionType == "music" }
+                                .map { item ->
+                                    val itemImageUrl =
+                                        "%s/Items/%s/Images/Primary?maxWidth=1080&quality=90"
+                                            .format(
+                                                appPreferences.instanceUrl, item.id
+                                            )
```
nielsvanvelzen, line 164:
```suggestion
                                    val itemImageUrl = apiClient.GetImageUrl(item, ImageOptions().apply {
                                        maxWidth = 180
                                        quality = 90
                                    })
```

app/src/main/java/org/jellyfin/mobile/car/MediaService.kt
```diff
@@ -0,0 +1,609 @@
...
+
+                                    val description: MediaDescriptionCompat =
+                                        MediaDescriptionCompat.Builder().apply {
+                                            setMediaId(MediaItemType.Library.toString() + "|" + item.id)
```
nielsvanvelzen, line 168: Only pass the item id as media id, if you need the library id you can pass that in a bundle using `setExtras`

app/src/main/java/org/jellyfin/mobile/car/MediaService.kt
```diff
@@ -0,0 +1,609 @@
...
+            appPreferences.instanceAccessToken,
+            appPreferences.instanceUserId
+        )
+
+        when (parentId) {
+            /**
+             * View that shows all the music libraries
+             */
+            MediaItemType.Libraries.toString() -> {
+                apiClient.GetUserViews(
```
nielsvanvelzen, line 148: Please add an extension function in `utils/ApiExtensions.kt` to use coroutines instead of callbacks for the GetUserViews function.

app/src/main/java/org/jellyfin/mobile/car/MediaService.kt
nielsvanvelzen: (Same for the other places that used this way of communicating data)

app/src/main/java/org/jellyfin/mobile/car/MediaService.kt
nielsvanvelzen: (Same for other used api functions)

app/src/main/java/org/jellyfin/mobile/car/MediaService.kt
```diff
@@ -0,0 +1,609 @@
...
+                        val query = ItemQuery()
+                        query.artistIds = arrayOf(primaryItemId)
+                        query.userId = appPreferences.instanceUserId
+                        query.sortBy = arrayOf("SortName")
+                        query.sortOrder = SortOrder.Ascending
+                        query.recursive = true
+                        query.imageTypeLimit = 1
+                        query.enableImageTypes = arrayOf(ImageType.Primary)
+                        query.limit = 100
+                        query.includeItemTypes = arrayOf("MusicAlbum")
```
nielsvanvelzen, line 295:
```suggestion
                        query.includeItemTypes = arrayOf(BaseItemType.MusicAlbum)
```

nielsvanvelzen: (also in some other places)

app/src/main/java/org/jellyfin/mobile/car/MediaService.kt
```diff
@@ -0,0 +1,609 @@
...
+                    }
+
+                    /**
+                     * View for a specific artist
+                     */
+                    MediaItemType.Artist.toString() -> {
+                        val query = ItemQuery()
+                        query.artistIds = arrayOf(primaryItemId)
+                        query.userId = appPreferences.instanceUserId
+                        query.sortBy = arrayOf("SortName")
```
nielsvanvelzen, line 289:
```suggestion
                        query.sortBy = arrayOf(ItemSortBy.SortName)
```

nielsvanvelzen: (also in some other places)

app/src/main/java/org/jellyfin/mobile/car/MediaService.kt
```diff
@@ -0,0 +1,609 @@
...
+            controlDispatcher: ControlDispatcher,
+            command: String,
+            extras: Bundle?,
+            cb: android.os.ResultReceiver?
+        ): Boolean {
+            return false
+        }
+
+        fun createMediaSource(mediaId: String): ProgressiveMediaSource {
+            val url =
```
nielsvanvelzen, line 586: I'm quite sure we have a function to generate this URL but can't find it right now.

nielsvanvelzen (changes requested):

nielsvanvelzen:
I had some issues to get the Android Auto app working but fortunately Google made the "Android Media Controller Test" app which is useful for testing the media browser:

https://github.com/googlesamples/android-media-controller

app/src/main/java/org/jellyfin/mobile/car/MediaService.kt
Maxr1998: > Please add an extension function in `utils/ApiExtensions.kt` to use coroutines instead of callbacks for the GetUserViews function.

Wrt this: I think we shouldn't merge this before we added the new api client in, so maybe it makes sense to wait a bit when it has to be updated/changed anyways at a later point.

nielsvanvelzen: I'm not sure when the new apiclient is ready so I'd rather merge this when it's ready and 2.0.0 is released instead of waiting for the apiclient.

mholt:
Sorry I am of no help here whatsoever, but I just want to cheer you on from the sidelines, sooooo yeah! :partying_face:  :raised_hands:  Really looking forward to this!

app/src/main/java/org/jellyfin/mobile/car/MediaService.kt
Spacetech: Unfortunately I don't think this could work. The `onLoadChildren` method only has that `parentId` string arg. This means the media id has to have all the information required to load the next view, which is why I'm playing with strings.

There is an overload method of [onLoadChildren with a bundle](https://developer.android.com/reference/androidx/media/MediaBrowserServiceCompat#onLoadChildren(java.lang.String,%20androidx.media.MediaBrowserServiceCompat.Result%3Cjava.util.List%3Candroid.support.v4.media.MediaBrowserCompat.MediaItem%3E%3E,%20android.os.Bundle)), but I never saw it called while testing. There's a [stackoverflow ](https://stackoverflow.com/questions/55863290/in-mediabrowserservice-onloadchildren-when-how-is-the-options-bundle-supplied)post about it but it has no answer.

app/src/main/java/org/jellyfin/mobile/car/MediaService.kt
Spacetech: I added the api extensions. Since `onLoadChildren` is not a suspend function, I had to add a `CoroutineScope` to media service. Let me know if there's a better way to do that. It seems to be working well so it's probably okay.

buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -92,6 +95,11 @@ object Dependencies {
         const val playServicesCastFramework = "com.google.android.gms:play-services-cast-framework:${Versions.playServicesCast}"
     }
 
+    object Car {
```
nielsvanvelzen, line 98: Shouldn't call this "car" since the media library / mediabrowser feature is not specifically for cars.

buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -92,6 +95,11 @@ object Dependencies {
...
+        val mediaMedia = androidx("media", Versions.mediaMedia)
+        var mediaSessionExtension = "com.google.android.exoplayer:extension-mediasession:2.12.0"
```
nielsvanvelzen, line 100: Version isn't aligned with the actual ExoPlayer extension, should use `Versions.exoPlayer`.

app/src/main/java/org/jellyfin/mobile/car/MediaService.kt
```diff
@@ -0,0 +1,596 @@
+package org.jellyfin.mobile.car
```
nielsvanvelzen, line 1: Shouldn't call this "car" since the media library / mediabrowser feature is not specifically for cars.

app/src/main/java/org/jellyfin/mobile/car/MediaService.kt
```diff
@@ -0,0 +1,596 @@
...
+
+            val uri = Uri.parse(url)
+
+            return ProgressiveMediaSource.Factory(dataSourceFactory).createMediaSource(uri)
+        }
+    }
+
+}
+
+private const val CONTENT_STYLE_SUPPORTED = "android.media.browse.CONTENT_STYLE_SUPPORTED"
```
nielsvanvelzen, line 592: Move all constants to a companion object in MediaService

app/src/main/java/org/jellyfin/mobile/car/MediaService.kt
```diff
@@ -0,0 +1,596 @@
...
+    private suspend fun loadView(parentId: String, result: Result<MutableList<MediaItem>>) {
+        when (parentId) {
+            /**
+             * View that shows all the music libraries
+             */
+            MediaItemType.Libraries.toString() -> {
+                val response = apiClient.getUserViews(appPreferences.instanceUserId!!)
+                if (response != null) {
+                    val views = response.items
+                        .filter { item -> item.collectionType == "music" }
```
nielsvanvelzen, line 163: Use the enum here

nielsvanvelzen (changes requested):

app/build.gradle.kts
```diff
@@ -87,6 +87,10 @@ dependencies {
     implementation(Dependencies.Cast.playServicesCast)
     implementation(Dependencies.Cast.playServicesCastFramework)
 
+    // Car
```
nielsvanvelzen, line 90: This comment still says car

nielsvanvelzen (changes requested):
Looks good! Just one tiny change left.

app/src/main/java/org/jellyfin/mobile/AppPreferences.kt
```diff
@@ -18,6 +18,22 @@ class AppPreferences(context: Context) {
...
+        }
+
+    var instanceUserId: String?
+        get() = sharedPreferences.getString(Constants.PREF_INSTANCE_USER_ID, null)
+        set(value) {
+            sharedPreferences.edit {
+                if (value != null) putString(Constants.PREF_INSTANCE_USER_ID, value) else remove(Constants.PREF_INSTANCE_USER_ID)
+            }
+        }
+
```
Maxr1998, line 36: I'm not sure whether I want to save those into SharedPreferences.. I've already integrated a database for server management in the `compose` branch, and I think I'll cherry-pick that very soon. If it's ok for you, I'd wait for that and refactor your credential persistence to use the new database then.

app/src/main/java/org/jellyfin/mobile/media/MediaService.kt
```diff
@@ -0,0 +1,597 @@
...
+            it.setPlayer(exoPlayer)
+            it.setPlaybackPreparer(
+                MediaPlaybackPreparer(
+                    exoPlayer,
+                    dataSourceFactory,
+                    appPreferences
+                )
+            )
+            it.setQueueNavigator(MediaQueueNavigator(mediaSession))
+        }
```
Maxr1998, line 97: Consider using `apply` here.

app/src/main/java/org/jellyfin/mobile/media/MediaService.kt
```diff
@@ -0,0 +1,597 @@
...
+    }
+
+    override fun onDestroy() {
+        mediaSession.run {
+            isActive = false
+            release()
+        }
+
+        exoPlayer.release()
+
```
Maxr1998, line 107:
```suggestion
```

app/src/main/java/org/jellyfin/mobile/media/MediaService.kt
```diff
@@ -0,0 +1,597 @@
...
+            it.setQueueNavigator(MediaQueueNavigator(mediaSession))
+        }
+    }
+
+    override fun onDestroy() {
+        mediaSession.run {
+            isActive = false
+            release()
+        }
+
```
Maxr1998, line 105:
```suggestion
```

app/src/main/java/org/jellyfin/mobile/media/MediaService.kt
```diff
@@ -0,0 +1,597 @@
...
+        clientPackageName: String,
+        clientUid: Int,
+        rootHints: Bundle?
+    ): BrowserRoot? {
+        val rootExtras = Bundle().apply {
+            putBoolean(CONTENT_STYLE_SUPPORTED, true)
+            putInt(CONTENT_STYLE_BROWSABLE_HINT, CONTENT_STYLE_LIST_ITEM_HINT_VALUE)
+            putInt(CONTENT_STYLE_PLAYABLE_HINT, CONTENT_STYLE_LIST_ITEM_HINT_VALUE)
+        }
+
```
Maxr1998, line 121:
```suggestion
```

app/src/main/java/org/jellyfin/mobile/media/MediaService.kt
```diff
@@ -0,0 +1,597 @@
...
+                        Pair(MediaItemType.LibrarySongs, R.string.mediaservice_library_songs),
+                        Pair(MediaItemType.LibraryGenres, R.string.mediaservice_library_genres),
+                        Pair(
+                            MediaItemType.LibraryPlaylists,
+                            R.string.mediaservice_library_playlists
+                        ),
+                    )
+
+                    val views = libraryViews
+                        .map { item ->
```
Maxr1998, line 218:
```suggestion
                    val views = libraryViews.map { item ->
```

app/src/main/java/org/jellyfin/mobile/media/MediaService.kt
```diff
@@ -0,0 +1,597 @@
...
+            command: String,
+            extras: Bundle?,
+            cb: android.os.ResultReceiver?
+        ): Boolean {
+            return false
+        }
+
+        fun createMediaSource(mediaId: String): ProgressiveMediaSource {
+            val url =
+                "%s/Audio/%s/universal?UserId=%s&DeviceId=%s&MaxStreamingBitrate=140000000&Container=opus,mp3|mp3,aac,m4a,m4b|aac,flac,webma,webm,wav,ogg&TranscodingContainer=ts&TranscodingProtocol=hls&AudioCodec=aac&api_key=%s&PlaySessionId=%s&StartTimeTicks=0&EnableRedirection=true&EnableRemoteMedia=true"
```
Maxr1998, line 573: Can you break this up or potentially (don't know if currently possible) use the API client for this?

app/src/main/java/org/jellyfin/mobile/media/MediaService.kt
```diff
@@ -0,0 +1,597 @@
...
+            var mediaItemIndex = 0
+
+            if (!shouldShuffle) {
+                mediaItemIndex = currentPlaylistItems.indexOfFirst { item ->
+                    item.mediaId == mediaId
+                }
+            }
+
+            exoPlayer.prepare(ConcatenatingMediaSource(*mediaSources.toTypedArray()))
+
```
Maxr1998, line 547:
```suggestion
```

app/src/main/java/org/jellyfin/mobile/media/MediaService.kt
```diff
@@ -0,0 +1,597 @@
...
+            exoPlayer.seekTo(mediaItemIndex, 0)
+        }
+
+        override fun onPrepareFromSearch(
+            query: String,
+            playWhenReady: Boolean,
+            extras: Bundle?
+        ) {
+
+        }
```
Maxr1998, line 557:
```suggestion
        ) {}
```

app/src/main/java/org/jellyfin/mobile/media/MediaService.kt
```diff
@@ -0,0 +1,597 @@
...
+            // show album art when playing the song
+            extras.putString(MediaMetadataCompat.METADATA_KEY_ALBUM_ART_URI, primaryImageUrl)
+
+            extras.putString(MediaMetadataCompat.METADATA_KEY_TITLE, item.name)
+            extras.putString(MediaMetadataCompat.METADATA_KEY_ARTIST, item.albumArtist)
+
+            if (item.indexNumber != null) {
+                extras.putInt(MediaMetadataCompat.METADATA_KEY_TRACK_NUMBER, item.indexNumber)
+            }
+
```
Maxr1998, line 459:
```suggestion
```

Maxr1998 (changes requested):
Thanks for the contribution! The changes generally look quite good, I'm not happy with some smaller parts and some of the formatting however. Also, I think some things require advance changes to the rest of the app, as described in the individual comments.

app/src/main/java/org/jellyfin/mobile/media/MediaService.kt
Spacetech: Niel mentioned that too but we don't see an api client method for it. How should I break it up?

app/src/main/java/org/jellyfin/mobile/AppPreferences.kt
Spacetech: Sounds good to me. Let me know when that's available.

app/src/main/java/org/jellyfin/mobile/media/MediaService.kt
Maxr1998: You could split it into multiple lines for example. Also, I recommend using the Kotlin format strings, like `"string${variable}string"`.

Spacetech: Thanks. Looks much better now

Maxr1998:
@Spacetech I've decided to not do the database stuff and merge this pretty much as is. I rebased it onto the latest master and fixed the conflicts and squashed the commits in that process. If it's ok for you, I'd force push the changes to this branch. Then, both Niels and me will give it another review, so we can hopefully merge it!

Spacetech:
@Maxr1998 Yea that sounds great. Go for it

Maxr1998:
Perfect! I just noticed that since we update ExoPlayer meanwhile, this uses deprecated APIs now. I'll fix those in a minute :grin:

Maxr1998 (approved):

Maxr1998:
Updated to the latest ExoPlayer APIs, refactored some code, fixed the album covers for the currently playing song (didn't work for me, cause only the albums have images in my config, now it falls back to those), and gave it a final test, I'm happy now :blush:

app/src/main/java/org/jellyfin/mobile/media/MediaService.kt
```diff
@@ -0,0 +1,568 @@
...
+                "Container=opus,mp3|mp3,aac,m4a,m4b|aac,flac,webma,webm,wav,ogg&" +
+                "TranscodingContainer=ts&" +
+                "TranscodingProtocol=hls&" +
+                "AudioCodec=aac&" +
+                "api_key=${appPreferences.instanceAccessToken}&" +
+                "PlaySessionId=${UUID.randomUUID()}&" +
+                "StartTimeTicks=0&" +
+                "EnableRedirection=true&" +
+                "EnableRemoteMedia=true"
+
```
nielsvanvelzen, line 556: Looking at https://github.com/jellyfin/jellyfin-apiclient-java/blob/master/jellyfin-api/src/main/kotlin-generated/org/jellyfin/apiclient/api/operations/UniversalAudioApi.kt#L43

- The deviceId should probably be plaintext, not base64
- PlaySessionId doesn't exist
- StartTimeTicks _should_ default to 0 when missing
- EnableRedirection is true by default

nielsvanvelzen (approved):

dkanada:
@Spacetech if you have any interest in adding this to Gelli feel free :sweat_smile: lots of people ask about Android Auto but I can't exactly test it easily without a device that supports it.

nielsvanvelzen:
Hey @Spacetech you mind it if I use your YouTube video in the 2.1 release notes?

Spacetech:
@nielsvanvelzen I don't mind, go for it.

PR #148, opened 2020-09-29
Add support for external player choice inside client settings

CarlosOlivo:
System choice by default.
Useful to force playback on a specific player without depending on android settings.
For now it only shows a list of the players that support playback tracking as long as they are installed and enabled.

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/bridge/ExternalPlayer.kt
+++ b/app/src/main/java/org/jellyfin/mobile/bridge/ExternalPlayer.kt
@@ -2,6 +2,7 @@ package org.jellyfin.mobile.bridge
 
 import android.app.Activity
 import android.content.Context
+import android.content.ComponentName
 import android.content.Intent
 import android.net.Uri
 import android.webkit.JavascriptInterface
@@ -10,9 +11,11 @@ import org.jellyfin.mobile.AppPreferences
 import org.jellyfin.mobile.R
 import org.jellyfin.mobile.fragment.WebViewFragment
 import org.jellyfin.mobile.player.source.JellyfinMediaSource
+import org.jellyfin.mobile.settings.ExternalPlayerPackage
 import org.jellyfin.mobile.settings.VideoPlayerType
 import org.jellyfin.mobile.utils.Constants
 import org.jellyfin.mobile.utils.runOnUiThread
+import org.jellyfin.mobile.utils.isPackageInstalled
 import org.jellyfin.mobile.utils.toast
 import org.jellyfin.mobile.webapp.WebappFunctionChannel
 import org.json.JSONException
@@ -36,6 +39,9 @@ class ExternalPlayer(private val fragment: WebViewFragment) : KoinComponent {
             val mediaSource = JellyfinMediaSource(JSONObject(args))
             if (mediaSource.playMethod == "DirectStream") {
                 val playerIntent = Intent(Intent.ACTION_VIEW).apply {
+                    if (fragment.isPackageInstalled(appPreferences.externalPlayerApp)) {
+                        component = getComponent(appPreferences.externalPlayerApp)
+                    }
                     setDataAndType(mediaSource.uri, mediaSource.mimeType)
                     putExtra("title", mediaSource.title)
                     putExtra("position", mediaSource.mediaStartMs.toInt())
@@ -194,6 +200,18 @@ class ExternalPlayer(private val fragment: WebViewFragment) : KoinComponent {
                 notifyEvent(Constants.EVENT_CANCELED)
                 context.toast(R.string.external_player_unknown_error, Toast.LENGTH_LONG)
             }
+        }
+    }
+
+    /**
+     * To ensure that the correct activity is called.
+     */
+    private fun getComponent(@ExternalPlayerPackage packageName: String): ComponentName? {
+        return when (packageName) {
+            ExternalPlayerPackage.MPV_PLAYER -> ComponentName(packageName, "$packageName.MPVActivity")
+            ExternalPlayerPackage.MX_PLAYER_FREE, ExternalPlayerPackage.MX_PLAYER_PRO -> ComponentName(packageName, "$packageName.ActivityScreen")
+            ExternalPlayerPackage.VLC_PLAYER -> ComponentName(packageName, "$packageName.gui.video.VideoPlayerActivity")
+            else -> null
         }
     }
 }
--- a/app/src/main/java/org/jellyfin/mobile/settings/SettingsFragment.kt
+++ b/app/src/main/java/org/jellyfin/mobile/settings/SettingsFragment.kt
@@ -16,6 +16,7 @@ import org.jellyfin.mobile.utils.Constants
 import org.jellyfin.mobile.utils.applyWindowInsetsAsMargins
 import org.jellyfin.mobile.utils.requireMainActivity
 import org.jellyfin.mobile.utils.withThemedContext
+import org.jellyfin.mobile.utils.isPackageInstalled
 import org.koin.android.ext.android.inject
 
 class SettingsFragment : Fragment() {
@@ -24,6 +25,7 @@ class SettingsFragment : Fragment() {
     private val settingsAdapter: PreferencesAdapter by lazy { PreferencesAdapter(buildSettingsScreen()) }
     private lateinit var backgroundAudioPreference: Preference
     private lateinit var swipeGesturesPreference: Preference
+    private lateinit var externalPlayerChoicePreference: Preference
 
     override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View {
         val localInflater = inflater.withThemedContext(requireContext(), R.style.AppTheme_Settings)
@@ -65,6 +67,7 @@ class SettingsFragment : Fragment() {
             defaultOnSelectionChange { selection ->
                 swipeGesturesPreference.enabled = selection == VideoPlayerType.EXO_PLAYER
                 backgroundAudioPreference.enabled = selection == VideoPlayerType.EXO_PLAYER
+                externalPlayerChoicePreference.enabled = selection == VideoPlayerType.EXTERNAL_PLAYER
             }
         }
         swipeGesturesPreference = checkBox(Constants.PREF_EXOPLAYER_ALLOW_SWIPE_GESTURES) {
@@ -75,6 +78,17 @@ class SettingsFragment : Fragment() {
         backgroundAudioPreference = checkBox(Constants.PREF_EXOPLAYER_ALLOW_BACKGROUND_AUDIO) {
             titleRes = R.string.pref_exoplayer_allow_background_audio
             enabled = appPreferences.videoPlayerType == VideoPlayerType.EXO_PLAYER
+        }
+        val externalPlayerOptions = listOf(
+            SelectionItem(ExternalPlayerPackage.MPV_PLAYER, R.string.external_player_mpv, R.string.external_player_mpv_description),
+            SelectionItem(ExternalPlayerPackage.MX_PLAYER_FREE, R.string.external_player_mx_player_free, R.string.external_player_mx_player_free_description),
+            SelectionItem(ExternalPlayerPackage.MX_PLAYER_PRO, R.string.external_player_mx_player_pro, R.string.external_player_mx_player_pro_description),
+            SelectionItem(ExternalPlayerPackage.VLC_PLAYER, R.string.external_player_vlc_player, R.string.external_player_vlc_player_description),
+        ).filter { isPackageInstalled(it.key) }.plus(SelectionItem(ExternalPlayerPackage.SYSTEM_DEFAULT, R.string.external_player_system_default, R.string.external_player_system_default_description))
+        if (!isPackageInstalled(appPreferences.externalPlayerApp)) appPreferences.externalPlayerApp = ExternalPlayerPackage.SYSTEM_DEFAULT
+        externalPlayerChoicePreference = singleChoice(Constants.PREF_EXTERNAL_PLAYER_APP, externalPlayerOptions) {
+            titleRes = R.string.external_player_app
+            enabled = appPreferences.videoPlayerType == VideoPlayerType.EXTERNAL_PLAYER
         }
     }
 
--- a/app/src/main/java/org/jellyfin/mobile/utils/SystemUtils.kt
+++ b/app/src/main/java/org/jellyfin/mobile/utils/SystemUtils.kt
@@ -4,6 +4,7 @@ import android.Manifest.permission.WRITE_EXTERNAL_STORAGE
 import android.app.*
 import android.content.Context
 import android.content.Intent
+import android.content.pm.PackageManager
 import android.content.pm.PackageManager.PERMISSION_GRANTED
 import android.net.Uri
 import android.os.Build
@@ -13,6 +14,7 @@ import android.provider.Settings
 import android.provider.Settings.System.ACCELEROMETER_ROTATION
 import androidx.appcompat.app.AppCompatActivity
 import androidx.core.content.getSystemService
+import androidx.fragment.app.Fragment
 import com.google.android.material.snackbar.Snackbar
 import kotlinx.coroutines.suspendCancellableCoroutine
 import kotlinx.coroutines.withTimeout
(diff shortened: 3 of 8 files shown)


nielsvanvelzen:
Since we're targeting SDK 30 we need to add the apps we query to the manifest: https://developer.android.com/about/versions/11/privacy/package-visibility

CarlosOlivo:
Ok, new changes:
- Added MPV Player
- [Query and interact with apps given an intent filter](https://developer.android.com/training/basics/intents/package-visibility#intent-signature)
- [Query and interact with specific packages](https://developer.android.com/training/basics/intents/package-visibility#package-name)

Note: New warning in Android Studio 4.1.
> WRITE_EXTERNAL_STORAGE no longer provides write access when targeting Android 10+

h1dden-da3m0n:
Android 10+ makes use of the new Storage API IIRC, there still is [a way to get the old behavior](https://developer.android.com/reference/android/R.attr#requestLegacyExternalStorage) though (but for how long :thinking:)

Maxr1998:
> Note: New warning in Android Studio 4.1.
> 
> > WRITE_EXTERNAL_STORAGE no longer provides write access when targeting Android 10+

You can safely ignore it, we only use `WRITE_EXTERNAL_STORAGE` below Android 10 for downloads, it isn't necessary on Android 10 and above.

app/src/main/java/org/jellyfin/mobile/bridge/ExternalPlayer.kt
```diff
@@ -196,4 +202,16 @@ class ExternalPlayer(private val fragment: WebViewFragment) : KoinComponent {
             }
         }
     }
+
+    /**
+     * To ensure that the correct activity is called.
+     */
+    private fun getComponent(packageName: String): ComponentName? {
+        return when (packageName) {
+            ExternalPlayerType.MPV_PLAYER -> ComponentName(packageName, "$packageName.MPVActivity")
```
nielsvanvelzen, line 211: Reading the documentation for [ComponentName](https://developer.android.com/reference/android/content/ComponentName) makes me think it should be like this:

```suggestion
            ExternalPlayerType.MPV_PLAYER -> ComponentName(packageName, "MPVActivity")
```

First parameter is `pkg`, the package. While the second is the actual class. (so no dots)

app/src/main/java/org/jellyfin/mobile/settings/SettingsFragment.kt
```diff
@@ -63,10 +65,23 @@ class SettingsFragment : Fragment() {
...
                 backgroundAudioPreference.enabled = selection == VideoPlayerType.EXO_PLAYER
             }
         }
+        val externalPlayerOptions = listOf(
+            SelectionItem(ExternalPlayerType.MPV_PLAYER, R.string.external_player_mpv, R.string.external_player_mpv_description),
+            SelectionItem(ExternalPlayerType.MX_PLAYER_FREE, R.string.external_player_mx_player_free, R.string.external_player_mx_player_free_description),
+            SelectionItem(ExternalPlayerType.MX_PLAYER_PRO, R.string.external_player_mx_player_pro, R.string.external_player_mx_player_pro_description),
+            SelectionItem(ExternalPlayerType.VLC_PLAYER, R.string.external_player_vlc_player, R.string.external_player_vlc_player_description),
+            SelectionItem(ExternalPlayerType.SYSTEM_CHOICE, R.string.external_player_system_choice, R.string.external_player_system_choice_description),
+        ).filter { it.key.isEmpty() || isPackageInstalled(it.key) }
```
nielsvanvelzen, line 79: Since we're building the list ourselves we can just exclude the empty key

nielsvanvelzen: Also @Maxr1998 what happens if I select VLC, remove the app and then open the settings? The option wouldn't exist because `isPackageInstalled` returns false. Does the library crash or use the `initialSelection`?

app/src/main/java/org/jellyfin/mobile/utils/SystemUtils.kt
```diff
@@ -115,6 +117,14 @@ private fun Context.downloadFile(request: DownloadManager.Request, @DownloadMeth
 
 fun Activity.isAutoRotateOn() = Settings.System.getInt(contentResolver, ACCELEROMETER_ROTATION, 0) == 1
 
+fun Fragment.isPackageInstalled(packageName: String): Boolean {
+    return try {
+        packageName.isNotEmpty() && requireContext().packageManager.getApplicationInfo(packageName, 0).enabled
+    } catch (e: PackageManager.NameNotFoundException) {
+        false
+    }
+}
```
nielsvanvelzen, line 126:
```suggestion
fun Fragment.isPackageInstalled(packageName: String) = try {
    packageName.isNotEmpty() && requireContext().packageManager.getApplicationInfo(packageName, 0).enabled
} catch (e: PackageManager.NameNotFoundException) {
    false
}
```

This is just a personal preference but you can write it like this

app/src/main/res/values/strings.xml
```diff
@@ -35,6 +35,17 @@
     <string name="external_player_not_supported_yet">Playback tracking not supported in selected external player yet. Contact developers to try to add support for it.</string>
     <string name="external_player_invalid_player">No valid external player selected. Playback tracking disabled.</string>
     <string name="external_player_unknown_error">Unknown error in external player. Playback tracking disabled.</string>
+    <string name="external_player_choice">External player choice</string>
```
nielsvanvelzen, line 38:
```suggestion
    <string name="external_player_choice">External player app</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -35,6 +35,17 @@
...
+    <string name="external_player_mpv">MPV Player</string>
+    <string name="external_player_mpv_description">A simple, small and light-weight free (as in freedom) video player. It supports a wide variety of media file formats.</string>
+    <string name="external_player_mx_player_free">MX Player Free</string>
+    <string name="external_player_mx_player_free_description">Powerful video player with advanced hardware acceleration and subtitle support. Ads everywhere.</string>
```
nielsvanvelzen, line 42: The "Ads everywhere." part sounds a bit offensive to the developer

app/src/main/res/values/strings.xml
```diff
@@ -35,6 +35,17 @@
...
+    <string name="external_player_mx_player_pro">MX Player Pro</string>
+    <string name="external_player_mx_player_pro_description">The paid version of MX Player which provides an uninterrupted video experience without any ads.</string>
+    <string name="external_player_vlc_player">VLC Player</string>
+    <string name="external_player_vlc_player_description">Free and open source cross-platform multimedia player that plays most multimedia files.</string>
+    <string name="external_player_system_choice">System choice</string>
```
nielsvanvelzen, line 47:
```suggestion
    <string name="external_player_system_choice">System default</string>
```

nielsvanvelzen (changes requested):
Code review only, did not test it yet.

app/src/main/java/org/jellyfin/mobile/bridge/ExternalPlayer.kt
Maxr1998: You still need the package name for the full class name however..

app/src/main/java/org/jellyfin/mobile/settings/SettingsFragment.kt
Maxr1998: It shows nothing as selected then, actually.

app/src/main/res/values/strings.xml
Maxr1998: Yeah, I wouldn't include any info about ads at all, honestly.

app/src/main/java/org/jellyfin/mobile/settings/SettingsFragment.kt
nielsvanvelzen: Not sure if that's the preferred behavior

app/src/main/java/org/jellyfin/mobile/bridge/ExternalPlayer.kt
nielsvanvelzen: Docs on constructor:

> cls
> String: The name of the class inside of pkg that implements the component. Can not be null. This value cannot be null.

Maxr1998: That only means that it's inside the package, you may still need to specify the full class name (with package).

CarlosOlivo: From [MX API - Intent definition](https://sites.google.com/site/mxvpen/api#h.p_ID_481:~:text=Class%20name%20for)

> Class name for free edition is **com.mxtech.videoplayer.ad.ActivityScreen** and pro edition is **com.mxtech.videoplayer.ActivityScreen**.

From [VLC Android Player Intents](https://wiki.videolan.org/Android_Player_Intents/#Sample_code:~:text=If%20you%20specifically%20want%20start%20VideoPlayerActivity%2C%20you%20can%20set)

>  vlcIntent.setComponent(new ComponentName("org.videolan.vlc", "**org.videolan.vlc.gui.video.VideoPlayerActivity**"));

app/src/main/java/org/jellyfin/mobile/settings/SettingsFragment.kt
CarlosOlivo: I'm not sure how to handle this case either, I made some changes but I don't know if it's the right way.

app/src/main/res/values/strings.xml
```diff
@@ -35,6 +35,17 @@
...
+    <string name="external_player_mpv">MPV Player</string>
+    <string name="external_player_mpv_description">A simple, small and light-weight free (as in freedom) video player. It supports a wide variety of media file formats.</string>
+    <string name="external_player_mx_player_free">MX Player Free</string>
+    <string name="external_player_mx_player_free_description">Powerful video player with advanced hardware acceleration and subtitle support.</string>
+    <string name="external_player_mx_player_pro">MX Player Pro</string>
+    <string name="external_player_mx_player_pro_description">The paid version of MX Player which provides an uninterrupted video experience.</string>
+    <string name="external_player_vlc_player">VLC Player</string>
+    <string name="external_player_vlc_player_description">Free and open source cross-platform multimedia player that plays most multimedia files.</string>
+    <string name="external_player_system_default">System default</string>
+    <string name="external_player_system_default_description">Let the system handle the player choice, some players may not be fully compatible.</string>
```
Maxr1998, line 48: Could you move this to the end of the file, please? As those are shown in the settings..

app/src/main/java/org/jellyfin/mobile/settings/ExternalPlayerType.java
```diff
@@ -0,0 +1,19 @@
...
+import androidx.annotation.StringDef;
+
+@StringDef({
+        ExternalPlayerType.MPV_PLAYER,
+        ExternalPlayerType.MX_PLAYER_FREE,
+        ExternalPlayerType.MX_PLAYER_PRO,
+        ExternalPlayerType.VLC_PLAYER,
+        ExternalPlayerType.SYSTEM_DEFAULT
+})
+public @interface ExternalPlayerType {
```
Maxr1998, line 13:
```suggestion
public @interface ExternalPlayerPackage {
```

app/src/main/java/org/jellyfin/mobile/bridge/ExternalPlayer.kt
```diff
@@ -196,4 +202,16 @@ class ExternalPlayer(private val fragment: WebViewFragment) : KoinComponent {
             }
         }
     }
+
+    /**
+     * To ensure that the correct activity is called.
+     */
+    private fun getComponent(packageName: String): ComponentName? {
```
Maxr1998, line 209:
```suggestion
    private fun getComponent(@ExternalPlayerPackage/Type packageName: String): ComponentName? {
```

app/src/main/java/org/jellyfin/mobile/settings/SettingsFragment.kt
```diff
@@ -63,10 +65,22 @@ class SettingsFragment : Fragment() {
...
+            SelectionItem(ExternalPlayerType.MPV_PLAYER, R.string.external_player_mpv, R.string.external_player_mpv_description),
+            SelectionItem(ExternalPlayerType.MX_PLAYER_FREE, R.string.external_player_mx_player_free, R.string.external_player_mx_player_free_description),
+            SelectionItem(ExternalPlayerType.MX_PLAYER_PRO, R.string.external_player_mx_player_pro, R.string.external_player_mx_player_pro_description),
+            SelectionItem(ExternalPlayerType.VLC_PLAYER, R.string.external_player_vlc_player, R.string.external_player_vlc_player_description),
+        ).filter { isPackageInstalled(it.key) }.plus(SelectionItem(ExternalPlayerType.SYSTEM_DEFAULT, R.string.external_player_system_default, R.string.external_player_system_default_description))
+        if (!isPackageInstalled(appPreferences.externalPlayerApp)) appPreferences.externalPlayerApp = ExternalPlayerType.SYSTEM_DEFAULT
+        externalPlayerChoicePreference = singleChoice(Constants.PREF_EXTERNAL_PLAYER_APP, externalPlayerOptions) {
+            titleRes = R.string.external_player_app
+            enabled = appPreferences.videoPlayerType == VideoPlayerType.EXTERNAL_PLAYER
+        }
```
Maxr1998, line 83: Could you put the external player settings below the ones for ExoPlayer please? That way it matches the order of the video player type preference.
Please also apply this to the rest of the code.

app/src/main/java/org/jellyfin/mobile/settings/ExternalPlayerType.java
```diff
@@ -0,0 +1,19 @@
...
+    String MPV_PLAYER = "is.xyz.mpv";
+    String MX_PLAYER_FREE = "com.mxtech.videoplayer.ad";
+    String MX_PLAYER_PRO = "com.mxtech.videoplayer.pro";
+    String VLC_PLAYER = "org.videolan.vlc";
+    String SYSTEM_DEFAULT = "";
```
Maxr1998, line 18:
```suggestion
    String SYSTEM_DEFAULT = "~system~";
```

*Note: '~' makes this an invalid Java package name, so you can use that.*

Maxr1998 (changes requested):

Maxr1998:
Thanks for the changes, I'll give it another thorough review tomorrow :+1:

app/src/main/java/org/jellyfin/mobile/bridge/ExternalPlayer.kt
nielsvanvelzen: Right, the Android documentation was a bit unclear to me.

nielsvanvelzen (approved):

app/src/main/java/org/jellyfin/mobile/utils/SystemUtils.kt
```diff
@@ -115,6 +118,12 @@ private fun Context.downloadFile(request: DownloadManager.Request, @DownloadMeth
 
 fun Activity.isAutoRotateOn() = Settings.System.getInt(contentResolver, ACCELEROMETER_ROTATION, 0) == 1
 
+fun Fragment.isPackageInstalled(@ExternalPlayerPackage packageName: String) = try {
```
Maxr1998, line 121: I don't think the annotation is necessary here.
```suggestion
fun Fragment.isPackageInstalled(packageName: String) = try {
```

Maxr1998 (changes requested):

app/src/main/java/org/jellyfin/mobile/utils/SystemUtils.kt
CarlosOlivo: Any specific reason? Or just because it is redundant? I thought it was good because due to the changes in Android 11 it is necessary to keep control of all the apps that we query in one place due to the new privacy limitations, unless in the future you want to add other apps apart from `@ExternalPlayerPackage`, in any case, if you confirm that you want this change, I will apply it.

Maxr1998: > Any specific reason? Or just because it is redundant? I thought it was good because due to the changes in Android 11 it is necessary to keep control of all the apps that we query in one place due to the new privacy limitations, unless in the future you want to add other apps apart from `@ExternalPlayerPackage`, in any case, if you confirm that you want this change, I will apply it.

I thought the function could be more generic, but the access control of Android 11 is a very good point, haven't thought of that. Let's merge this then. :+1:

Maxr1998 (approved):

PR #146, opened 2020-09-28
Set appMode for webui

nielsvanvelzen:
The webui needs the `appMode` property set to either `cordova` or `android` for some behavior. We didn't do that.

Relevant code in web: https://github.com/jellyfin/jellyfin-web/blob/692d404285c19e1700a74fec23b4804273d12ac3/src/components/castSenderApi.js#L3

As you can see, if the app mode is not set it will always load the cast_sender JavaScript from Google, this JS would then overwrite our own injected code. By setting the appMode it doesn't do that.

It doesn't fix most chromecast issues though.

PR #154, opened 2020-09-29
Revert "Set appMode for webui"

### LINKED ISSUE #153: Setting appMode to 'android' prevents server connection on fresh installs (rc.7)
**Describe the bug**
I try to enter the server but it does not enter. The three versions of the rc.-7 version (release, unsigned and debug) do not work. When you put the url of the server and click on connect, the app pretends to be entering the server, but then it returns to the same host screen.

**To Reproduce**
1. Enter the url of the server
2. Click on Connect
3. Jellyfin logo
4. Return to the same screen to enter the Host

**System:**
 - App version: [2.0.0-rc.7]
 - Android version: [Android 9
 - Device manufacturer and model: [10.6.4]

Maxr1998:
Reverts #146
Closes #153

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/utils/WebAppUtils.kt
+++ b/app/src/main/java/org/jellyfin/mobile/utils/WebAppUtils.kt
@@ -7,14 +7,11 @@ import java.io.IOException
 
 const val JS_INJECTION_CODE = """
 !function() {
-    window['appMode'] = 'android';
-    
     var scripts = [
         '/native/nativeshell.js',
         '/native/EventEmitter.js',
         '/native/chrome.cast.js',
     ];
-    
     scripts.forEach(function(src) {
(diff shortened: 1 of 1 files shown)


nielsvanvelzen:
I think we should actually look into what additional changes we should make to fix #153 instead of removing the appmode, as it does make some changes that are required in the webui.

Maxr1998:
> I think we should actually look into what additional changes we should make to fix #153 instead of removing the appmode, as it does make some changes that are required in the webui.

We could first try whether `appMode='cordova'` works better, yes. But if it's an obscure bug that can't easily be fixed, I wouldn't spend too much time on it..
The cast code has issues regardless of the appMode.

nielsvanvelzen:
Setting the appmode to cordova is the same as android

nielsvanvelzen (approved):
Let's just revert it for now

PR #165, opened 2020-10-04
Rework app / webapp communication

Maxr1998:
Put Kotlin channel inside wrapper class, use channel for communication from NativeInterface, RemotePlayerService and ExternalPlayer.


app/src/main/java/org/jellyfin/mobile/webapp/RemoteVolumeProvider.kt
```diff
@@ -0,0 +1,25 @@
...
+
+import android.media.AudioManager
+import android.media.VolumeProvider
+import org.jellyfin.mobile.utils.Constants
+
+class RemoteVolumeProvider(private val webappFunctionChannel: WebappFunctionChannel) : VolumeProvider(VOLUME_CONTROL_ABSOLUTE, 100, 0) {
+    override fun onAdjustVolume(direction: Int) {
+        when (direction) {
+            AudioManager.ADJUST_RAISE -> {
+                webappFunctionChannel.triggerInputManagerAction(Constants.INPUT_MANAGER_COMMAND_VOL_UP)
```
nielsvanvelzen, line 11: Fix your TODO's

app/src/main/java/org/jellyfin/mobile/webapp/RemoteVolumeProvider.kt
```diff
@@ -0,0 +1,25 @@
...
+            }
+            AudioManager.ADJUST_LOWER -> {
+                webappFunctionChannel.triggerInputManagerAction(Constants.INPUT_MANAGER_COMMAND_VOL_DOWN)
+                currentVolume -= 2 // TODO: have web notify app with new volume instead
+            }
+        }
+    }
+
+    override fun onSetVolumeTo(volume: Int) {
+        webappFunctionChannel.callPlaybackManager("sendCommand({Name:'SetVolume', Arguments:{Volume:$volume}})")
```
nielsvanvelzen, line 22:
```suggestion
        webappFunctionChannel.callPlaybackManager("sendCommand({ Name: 'SetVolume', Arguments:{ Volume: $volume }})")
```

Would it be worth adding additional methods to the function channel to generate these commands so we don't have hardcoded strings like this one everywhere?

app/src/main/java/org/jellyfin/mobile/webapp/RemoteVolumeProvider.kt
```diff
@@ -0,0 +1,25 @@
+package org.jellyfin.mobile.webapp
+
+import android.media.AudioManager
+import android.media.VolumeProvider
+import org.jellyfin.mobile.utils.Constants
+
+class RemoteVolumeProvider(private val webappFunctionChannel: WebappFunctionChannel) : VolumeProvider(VOLUME_CONTROL_ABSOLUTE, 100, 0) {
```
nielsvanvelzen, line 7:
```suggestion
class RemoteVolumeProvider(
    private val webappFunctionChannel: WebappFunctionChannel
) : VolumeProvider(VOLUME_CONTROL_ABSOLUTE, 100, 0) {
```

I prefer multi-line constructors if they become too long

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/mobile/webapp/RemoteVolumeProvider.kt
Maxr1998: They were already there in master xD

app/src/main/java/org/jellyfin/mobile/webapp/RemoteVolumeProvider.kt
Maxr1998: Yeah, I think that makes sense..

app/src/main/java/org/jellyfin/mobile/webapp/RemoteVolumeProvider.kt
nielsvanvelzen: Is it something we can fix or is it not possible without changing webui?

nielsvanvelzen (approved):

app/src/main/java/org/jellyfin/mobile/webapp/RemoteVolumeProvider.kt
Maxr1998: It needs changes to the web UI, unfortunately. But then again, we'll probably have a native UI before we get those changes, lol.

PR #175, opened 2020-10-07
Initial database setup

Maxr1998:
Required to persist server data at a later date, unused for now.

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/model/sql/entity/ServerEntity.kt
+++ b/app/src/main/java/org/jellyfin/mobile/model/sql/entity/ServerEntity.kt
@@ -1,0 +1,24 @@
+package org.jellyfin.mobile.model.sql.entity
+
+import androidx.room.ColumnInfo
+import androidx.room.Entity
+import androidx.room.Index
+import androidx.room.PrimaryKey
+import org.jellyfin.mobile.model.sql.entity.ServerEntity.Key.HOSTNAME
+import org.jellyfin.mobile.model.sql.entity.ServerEntity.Key.TABLE_NAME
+
+@Entity(tableName = TABLE_NAME, indices = [Index(value = arrayOf(HOSTNAME), unique = true)])
+data class ServerEntity(
+    @PrimaryKey(autoGenerate = true) @ColumnInfo(name = SERVER_ID) val id: Long,
+    @ColumnInfo(name = HOSTNAME) val hostname: String,
+    @ColumnInfo(name = LAST_USED_TIMESTAMP) val lastUsedTimestamp: Long,
+) {
+    constructor(hostname: String) : this(0, hostname, System.currentTimeMillis())
+
+    companion object Key {
+        const val TABLE_NAME = "Server"
+        const val SERVER_ID = "id"
+        const val HOSTNAME = "hostname"
+        const val LAST_USED_TIMESTAMP = "last_used_timestamp"
+    }
+}
--- a/app/src/main/java/org/jellyfin/mobile/model/sql/entity/UserEntity.kt
+++ b/app/src/main/java/org/jellyfin/mobile/model/sql/entity/UserEntity.kt
@@ -1,0 +1,25 @@
+package org.jellyfin.mobile.model.sql.entity
+
+import androidx.room.ColumnInfo
+import androidx.room.Entity
+import org.jellyfin.mobile.model.sql.entity.UserEntity.Key.SERVER_ID
+import org.jellyfin.mobile.model.sql.entity.UserEntity.Key.TABLE_NAME
+import org.jellyfin.mobile.model.sql.entity.UserEntity.Key.USER_ID
+
+@Entity(tableName = TABLE_NAME, primaryKeys = [SERVER_ID, USER_ID])
+data class UserEntity(
+    @ColumnInfo(name = SERVER_ID) val serverId: Long,
+    @ColumnInfo(name = USER_ID) val id: String,
+    @ColumnInfo(name = ACCESS_TOKEN) val accessToken: String?,
+    @ColumnInfo(name = LAST_LOGIN_TIMESTAMP) val lastLoginTimestamp: Long,
+) {
+    constructor(serverId: Long, userId: String, accessToken: String?) : this(serverId, userId, accessToken, System.currentTimeMillis())
+
+    companion object Key {
+        const val TABLE_NAME = "User"
+        const val SERVER_ID = "server_id"
+        const val USER_ID = "user_id"
+        const val ACCESS_TOKEN = "access_token"
+        const val LAST_LOGIN_TIMESTAMP = "last_login_timestamp"
+    }
+}
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -5,6 +5,7 @@ plugins {
     id("com.android.application")
     kotlin("android")
     kotlin("android.extensions")
+    kotlin("kapt")
     id("de.mannodermaus.android-junit5")
     id("com.github.ben-manes.versions") version Dependencies.Versions.dependencyUpdates
 }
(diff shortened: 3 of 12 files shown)


nielsvanvelzen:
According to [the documentation](https://developer.android.com/training/data-storage/room/migrating-db-versions#export-schema) we should add the schema to git. I also think we should add migration tests right from the start,

Maxr1998:
Ok, I added them. Migration tests currently don't make a lot of sense, as we don't have any migrations yet.

nielsvanvelzen:
Is there no need for an "initial migration"?

Maxr1998:
Nope.

app/src/main/java/org/jellyfin/mobile/model/sql/entity/ServerEntity.kt
```diff
@@ -0,0 +1,24 @@
+package org.jellyfin.mobile.model.sql.entity
+
+import androidx.room.ColumnInfo
+import androidx.room.Entity
+import org.jellyfin.mobile.model.sql.entity.ServerEntity.Key.HOSTNAME
+import org.jellyfin.mobile.model.sql.entity.ServerEntity.Key.TABLE_NAME
+
+@Entity(tableName = TABLE_NAME, primaryKeys = [HOSTNAME])
+data class ServerEntity(
```
nielsvanvelzen, line 9: We should use a unique identifier for the server so we can reference it from other tables in the future (like stored users). Should **not** be the server_id from the public server info.

app/src/main/java/org/jellyfin/mobile/model/sql/entity/ServerEntity.kt
```diff
@@ -0,0 +1,24 @@
...
+    @ColumnInfo(name = HOSTNAME) val hostname: String,
+    @ColumnInfo(name = USER_ID) val userId: String,
+    @ColumnInfo(name = ACCESS_TOKEN) val accessToken: String?,
+    @ColumnInfo(name = LAST_ACTIVE_TIME) val lastActiveTime: Long,
```
nielsvanvelzen, line 13: User info should use a separate table for multi-user support

app/src/main/java/org/jellyfin/mobile/model/sql/entity/ServerEntity.kt
```diff
@@ -0,0 +1,24 @@
...
+    @ColumnInfo(name = LAN_HOSTNAME) val lanHostname: String? = null,
```
nielsvanvelzen, line 14: I think `alternativeHostname` sounds better, since it doesn't imply that it's for LAN only.

nielsvanvelzen: (also, order is important. Should place hostname and this one next to each other for readability)

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/mobile/model/sql/entity/ServerEntity.kt
```diff
@@ -0,0 +1,25 @@
...
+import androidx.room.Entity
+import androidx.room.Index
+import androidx.room.PrimaryKey
+import org.jellyfin.mobile.model.sql.entity.ServerEntity.Key.HOSTNAME
+import org.jellyfin.mobile.model.sql.entity.ServerEntity.Key.TABLE_NAME
+
+@Entity(tableName = TABLE_NAME, indices = [Index(value = arrayOf(HOSTNAME), unique = true)])
+data class ServerEntity(
+    @PrimaryKey(autoGenerate = true)
+    @ColumnInfo(name = SERVER_ID) val id: Long,
```
nielsvanvelzen, line 13:
```suggestion
    @PrimaryKey(autoGenerate = true) @ColumnInfo(name = SERVER_ID) val id: Long,
```

or 

```suggestion
    @PrimaryKey(autoGenerate = true)
    @ColumnInfo(name = SERVER_ID)
    val id: Long,
```

Right now it looks like primarykey doesn't belong to anything

app/src/main/java/org/jellyfin/mobile/model/sql/entity/ServerEntity.kt
```diff
@@ -0,0 +1,25 @@
...
+    @ColumnInfo(name = HOSTNAME) val hostname: String,
+    @ColumnInfo(name = LAST_USED_TIMESTAMP) val lastUsedTimestamp: Long,
```
nielsvanvelzen, line 15: Why add the type to the name?

```suggestion
    @ColumnInfo(name = LAST_USED_TIMESTAMP) val lastUsed: Long,
```

Also, "lastActive" might be a better name

app/src/main/java/org/jellyfin/mobile/model/sql/entity/UserEntity.kt
```diff
@@ -0,0 +1,25 @@
...
+
+import androidx.room.ColumnInfo
+import androidx.room.Entity
+import org.jellyfin.mobile.model.sql.entity.UserEntity.Key.SERVER_ID
+import org.jellyfin.mobile.model.sql.entity.UserEntity.Key.TABLE_NAME
+import org.jellyfin.mobile.model.sql.entity.UserEntity.Key.USER_ID
+
+@Entity(tableName = TABLE_NAME, primaryKeys = [SERVER_ID, USER_ID])
+data class UserEntity(
+    @ColumnInfo(name = SERVER_ID) val serverId: Long,
```
nielsvanvelzen, line 11:
```suggestion
    @ColumnInfo(name = SERVER) val server: Long,
```

IMO if we refer to a table we can just use the name of the table since we'll always point to the primary key of the referred table.

app/src/main/java/org/jellyfin/mobile/model/sql/entity/UserEntity.kt
```diff
@@ -0,0 +1,25 @@
...
+    @ColumnInfo(name = USER_ID) val id: String,
+    @ColumnInfo(name = ACCESS_TOKEN) val accessToken: String?,
+    @ColumnInfo(name = LAST_LOGIN_TIMESTAMP) val lastLoginTimestamp: Long,
```
nielsvanvelzen, line 14: See comment on lastUsedTimestamp

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/mobile/model/sql/entity/ServerEntity.kt
Maxr1998: Dunno, I prefer to state that it's a timestamp.

app/src/main/java/org/jellyfin/mobile/model/sql/entity/UserEntity.kt
Maxr1998: I disagree, since this isn't an instance of `ServerEntity`. It only makes sense in classes like `ServerUser`, where it's actually mapped to that type.

nielsvanvelzen (approved):

CarlosOlivo:
I just noticed that this PR adds a new warning in Azure, I don't know if this affects something in the future. It does not generate the warning for me but I am using the embedded JDK shipped with Android Studio as recommended.

`warning: Current JDK version 1.8.0_265-b01 has a bug (https://bugs.openjdk.java.net/browse/JDK-8007720) that prevents Room from being incremental. Consider using JDK 11+ or the embedded JDK shipped with Android Studio 3.5+.[WARN] Incremental annotation processing requested, but support is disabled because the following processors are not incremental: androidx.room.RoomProcessor (DYNAMIC).`

https://dev.azure.com/jellyfin-project/jellyfin/_build/results?buildId=16226&view=logs&j=ca395085-040a-526b-2ce8-bdc85f692774&t=85c38824-6449-5223-6cf4-ccf7ebb27e39&l=65

Maxr1998:
Since it only affects CI and only increases build time, it's not too big of an issue. Still, paging @dkanada and @joshuaboniface for some input. Could we possible update to OpenJDK 11 in the pipeline?

h1dden-da3m0n:
Disclaimer: very basic research (could be wrong)

This seems to be a simple change/addition of a `input` to the Gradle@v2 tasks, see [MS-Docs](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/build/gradle?view=azure-devops). (to use JDK 11 instead of 8)

I'll fire you a quick PR for that in a moment and you can then decide to use it :wink:

PR #177, opened 2020-10-08
Add support for external subtitles in some external players

CarlosOlivo:
Works at least in MPV & MX Player
VLC Player not supported: https://trac.videolan.org/vlc/ticket/18736

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/bridge/ExternalPlayer.kt
+++ b/app/src/main/java/org/jellyfin/mobile/bridge/ExternalPlayer.kt
@@ -3,6 +3,7 @@ package org.jellyfin.mobile.bridge
 import android.app.Activity
 import android.content.Context
 import android.content.Intent
+import android.net.Uri
 import android.webkit.JavascriptInterface
 import android.widget.Toast
 import org.jellyfin.mobile.AppPreferences
@@ -25,8 +26,6 @@ class ExternalPlayer(private val fragment: WebViewFragment) : KoinComponent {
 
     private val appPreferences: AppPreferences by inject()
     private val webappFunctionChannel: WebappFunctionChannel by inject()
-    private var mediaSource: JellyfinMediaSource? = null
-    private var playerIntent: Intent? = null
 
     @JavascriptInterface
     fun isEnabled() = appPreferences.videoPlayerType == VideoPlayerType.EXTERNAL_PLAYER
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/jellyfin/mobile/bridge/ExternalPlayer.kt
```diff
@@ -42,6 +43,15 @@ class ExternalPlayer(private val fragment: WebViewFragment) : KoinComponent {
                     putExtra("position", mediaSource?.mediaStartMs)
                     putExtra("return_result", true)
                     putExtra("secure_uri", true)
+                    if (mediaSource?.subtitleTracksGroup?.selectedTrack != -1) {
+                        val selectedTrack = mediaSource?.subtitleTracksGroup?.tracks?.get(mediaSource?.subtitleTracksGroup?.selectedTrack as Int)?.url
+                        if (selectedTrack != null) {
```
nielsvanvelzen, line 48:
```suggestion
                    val selectedTrack = mediaSource?.subtitleTracksGroup?.tracks?.getOrNull(mediaSource?.subtitleTracksGroup?.selectedTrack as Int)?.url
                    if (selectedTrack != null) {
```

I think you can do something like this

app/src/main/java/org/jellyfin/mobile/bridge/ExternalPlayer.kt
```diff
@@ -42,6 +43,13 @@ class ExternalPlayer(private val fragment: WebViewFragment) : KoinComponent {
                     putExtra("position", mediaSource?.mediaStartMs)
                     putExtra("return_result", true)
                     putExtra("secure_uri", true)
+                    val selectedTrack = mediaSource?.subtitleTracksGroup?.tracks?.getOrNull(mediaSource?.subtitleTracksGroup?.selectedTrack as Int)?.url
```
Maxr1998, line 46: Instead of using so many null-safe accessors, couldn't you make `mediaSource` a local non-null variable?

Maxr1998 (changes requested):

app/src/main/java/org/jellyfin/mobile/bridge/ExternalPlayer.kt
CarlosOlivo: I'm not sure if I should, mediaSource is originally a JSON object and I can't be sure that it will always have the necessary properties.

Maxr1998: That's checked internally of JellyfinMediaSource, so no worries there. And the mediaSource itself still doesn't need to be a nullable var.

CarlosOlivo: Roger that.

Maxr1998 (approved):

Issue #173, opened 2020-10-07
Unresolvable hostname crashes client


jcdick1:
**Describe the bug**
If the saved server hostname is unresolvable, the client crashes.

**To Reproduce**
These are the steps to reproduce in my situation
1. Have a JF server with a LAN hostname that is different than the hostname that is used for remote viewing, i.e. behind a reverse proxy
2. While connected to local network, enter the LAN server hostname and log in to the server.  Home page is presented and playback is available.
3. Close client, and use app management to terminate it.
4. Switch WiFI off so that the client is connecting over cell data to force remote access.
5. Open client.  There is immediately an upside down Android logo or other crash indicator.  No option to change servers.
6. Close and terminate the client app
7. Reactivate wifi
8 Reopen app, and the home page is accessible again.
9. Use "select server" to change the hostname saved in the app to the remote access hostname.
10. Close the app
11. Turn off Wifi to trigger remote access
12. Reopen the app, and the homepage becomes accessible via the remote hostname

**Expected behavior**
Some notice of server unavailable or method to switch servers

**Logs**

**Screenshots**
![Screenshot_20201007-092257](https://user-images.githubusercontent.com/49927327/95365027-2e9c2100-0897-11eb-83e3-1c5fc69a25db.png)

**System (please complete the following information):**
 - App version: 2.0.0
 - Android version: Android 11
 - Device manufacturer and model: Google Pixel 3a
 - Jellyfin server version: 10.6.4
 - The type of video player implementation used: Exoplayer

CarlosOlivo:
#44

nielsvanvelzen:
Closing in favor of #44

Issue #131, opened 2020-09-16
"Web page not available" shown if server is off


yashx:
**Describe the bug**

If I open the app when the server is off. I get a screen showing "Web page not available". I was able to fix it by starting the server and then opening the app or by clearing app data. 

**To Reproduce**

1. Start Jellyfin server
2. Connect using the Android App
3. Close the app
4. Stop the server
5. Open the app

**Expected behavior**

Show "connect to server" screen and give a reason if it can't connect 

**Screenshots**

https://imgur.com/a/Nsl8Zbb

**System:**
 - App version: 2.0.0-rc.5
 - Android version: Android 10
 - Device manufacturer and model: Xiaomi Pocophone F1
 - Jellyfin server version: 10.6.4

nielsvanvelzen:
Duplicate of #44

Issue #44, opened 2020-08-24
Permanently blocked at loading screen


willymilkytron:
After signing in and out from a server the app is permanently blocked in a loading screen with no way to login, logout or change server. Only way to unblock is to clear app data from android app manager.

![Screenshot_1598255472](https://user-images.githubusercontent.com/62460473/91018112-73be0980-e5ef-11ea-851c-464cd7dbb5fa.png)

Maxr1998:
Is this on the new client? I can't reproduce it on my setup.

willymilkytron:
Yes, this is the new client

willymilkytron:
https://imgur.com/a/Bf3aZBs

Maxr1998:
I tried to reproduce this with the exact same steps, but couldn't. Maybe it's specific to your setup. Could you capture and upload a log for me, so that I can see whether it's related to the app or the web client?

nielsvanvelzen:
I too, was unable to reproduce this issue. Can you give us more details like your server version?

willymilkytron:
Server version is 10.6.0

willymilkytron:
I just reproduced it on a clean install (throwaway docker container from image jellyfin/jellyfin:10.6.0)

CarlosOlivo:
I don't know if it's the same error, but I found one that has a similar behavior

1. Start jellyfin server
2. Start jellyfin app & login
3. Kill jellyfin app
4. Kill jellyfin server
5. Start jellyfin app

**Expected behavior**: Initial screen to select server, since the last server is not available as in version v2.0.0-rc.3

![image](https://user-images.githubusercontent.com/25919231/92612692-bafafa00-f27f-11ea-84b2-1821e89828e7.png)

**Error**: Black screen with WebView error in jellyfin debug aka master branch
`09-09 09:10:50.422 E/MainActivity$initialize(16951): Received WebView error at http://192.168.100.2:8096/: net::ERR_CONNECTION_TIMED_OUT`

![image](https://user-images.githubusercontent.com/25919231/92612727-c2220800-f27f-11ea-80d2-31a6a5a1dd4b.png)

nielsvanvelzen:
I tried starting the app with my server off and it does indeed cause issues. On my emulator it will show the issue from @willymilkytron. I think it doesn't do that for you @CarlosOlivo because for you the webview didn't load anything from cache.

[...]

PR #181, opened 2020-10-09
Handle WebView error on unavailable instance

CarlosOlivo:
This assumes that if an error occurs when loading the url of the saved instance, it is not available. Which would redirect the user to the connection fragment instead of showing a WebView error. Similar to previous behavior in v2.0.0-rc.3

Note: This only handles the case when even cache loading fails, #44 still needs to be fixed.

See https://github.com/jellyfin/jellyfin-android/issues/44#issuecomment-689607259 #131 #173

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
+++ b/app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
@@ -157,6 +157,8 @@ class WebViewFragment : Fragment() {
             override fun onReceivedError(view: WebView, request: WebResourceRequest, error: WebResourceErrorCompat) {
                 val description = if (WebViewFeature.isFeatureSupported(WebViewFeature.WEB_RESOURCE_ERROR_GET_DESCRIPTION)) error.description else null
                 Timber.e("Received WebView error at %s: %s", request.url.toString(), description)
+                if (request.url.toString() == appPreferences.instanceUrl)
+                    runOnUiThread { onErrorReceived() }
             }
         }
         webChromeClient = WebChromeClient()


app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
```diff
@@ -157,6 +157,8 @@ class WebViewFragment : Fragment() {
             override fun onReceivedError(view: WebView, request: WebResourceRequest, error: WebResourceErrorCompat) {
                 val description = if (WebViewFeature.isFeatureSupported(WebViewFeature.WEB_RESOURCE_ERROR_GET_DESCRIPTION)) error.description else null
                 Timber.e("Received WebView error at %s: %s", request.url.toString(), description)
+                if ("${request.url}" == appPreferences.instanceUrl)
```
nielsvanvelzen, line 160: Why not?
```suggestion
                if (request.url == appPreferences.instanceUrl)
```

nielsvanvelzen: or `request.url.toString()`

CarlosOlivo: Operator '`==`' cannot be applied to '`Uri!`' and '`String?`'
`request.url` is an `Uri`, I can change to `request.url.toString()` or keep `"${request.url}"`.

CarlosOlivo: > or `request.url.toString()`

Ok

nielsvanvelzen (approved):

Maxr1998:
I'm not 100% sure why I removed that anymore, but I think I did because my code was a bit oversensitive, showing the error page when only something like a script or JavaScript call failed..

But well, we can merge this and give it a try, if there are to many false positives, we can always revert it again.

Maxr1998 (approved):

PR #182, opened 2020-10-09
Audio device profile optimizations + exoplayer update

vitorsemeano:
This PR adds some optimizations to audio capabilities of the device. 

I added max channels and max sample rate to the device profile. 

I also added profiles for aac codec.

Also updated the version of exoplayer to 2.12.0

### DIFF
--- a/app/src/main/assets/native/exoplayer.js
+++ b/app/src/main/assets/native/exoplayer.js
@@ -256,14 +256,44 @@ define(['events', 'appSettings', 'loading', 'playbackManager'], function (events
                             var audioCodec = codecs.audioCodecs[index];
                             audioCodecs.push(audioCodec.codec);
 
+                            var profiles = audioCodec.profiles.join('|');
+                            var maxChannels = audioCodec.maxChannels;
+                            var maxSampleRate = audioCodec.maxSampleRate;
+
+                            var conditions = [{
+                                Condition: 'LessThanEqual',
+                                Property: 'AudioBitrate',
+                                Value: audioCodec.maxBitrate
+                            }];
+
+                            if (profiles) {
+                                conditions.push({
+                                    Condition: 'EqualsAny',
+                                    Property: 'AudioProfile',
+                                    Value: profiles
+                                });
+                            }
+
+                            if (maxChannels) {
+                                conditions.push({
+                                    Condition: 'LessThanEqual',
+                                    Property: 'AudioChannels',
+                                    Value: maxChannels
+                                });
+                            }
+
+                            if (maxSampleRate) {
+                                conditions.push({
+                                    Condition: 'LessThanEqual',
+                                    Property: 'AudioSampleRate',
+                                    Value: maxSampleRate
+                                });
+                            }
+
                             profile.CodecProfiles.push({
                                 Type: 'Audio',
                                 Codec: audioCodec.codec,
-                                Conditions: [{
-                                    Condition: 'LessThanEqual',
-                                    Property: 'AudioBitrate',
-                                    Value: audioCodec.maxBitrate
-                                }]
+                                Conditions: conditions
                             });
                         }
                     }
--- a/app/src/main/java/org/jellyfin/mobile/player/ExoPlayerCodec.kt
+++ b/app/src/main/java/org/jellyfin/mobile/player/ExoPlayerCodec.kt
@@ -93,7 +93,7 @@ class ExoPlayerCodec(codecCapabilities: CodecCapabilities) {
             put("levels", JSONArray(levels))
             put("maxBitrate", maxBitrate)
             put("maxChannels", maxChannels)
-            put("sampleRates", maxSampleRate)
+            put("maxSampleRate", maxSampleRate)
         }
     } catch (e: JSONException) {
         null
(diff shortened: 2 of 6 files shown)


app/src/main/java/org/jellyfin/mobile/player/ExoPlayerCodec.kt
```diff
@@ -17,6 +18,8 @@ class ExoPlayerCodec(codecCapabilities: CodecCapabilities) {
     private val profiles: MutableList<String> = ArrayList()
     private val levels: MutableList<Int> = ArrayList()
     private val maxBitrate: Int
+    private var maxChannels: Int = 0
```
Maxr1998, line 21:
```suggestion
    private val maxChannels: Int
```

app/src/main/java/org/jellyfin/mobile/player/ExoPlayerCodec.kt
```diff
@@ -17,6 +18,8 @@ class ExoPlayerCodec(codecCapabilities: CodecCapabilities) {
...
+    private var maxSampleRate: Int? = null
```
Maxr1998, line 22:
```suggestion
    private val maxSampleRate: Int?
```

app/src/main/java/org/jellyfin/mobile/player/ExoPlayerCodec.kt
```diff
@@ -32,6 +35,13 @@ class ExoPlayerCodec(codecCapabilities: CodecCapabilities) {
                 isAudio = true
                 codec = audioCodec
                 maxBitrate = codecCapabilities.audioCapabilities.bitrateRange.upper
+                maxChannels = codecCapabilities.audioCapabilities.maxInputChannelCount
+
+                val sampleRates = codecCapabilities.audioCapabilities.supportedSampleRateRanges
+
+                if (sampleRates.isNotEmpty()) {
+                    maxSampleRate = sampleRates.last().upper
+                }
```
Maxr1998, line 44:
```suggestion
                maxSampleRate = if (sampleRates.isNotEmpty()) sampleRates.last().upper else null
```

app/src/main/assets/native/exoplayer.js
```diff
@@ -256,14 +256,44 @@ define(['events', 'appSettings', 'loading', 'playbackManager'], function (events
                             var audioCodec = codecs.audioCodecs[index];
                             audioCodecs.push(audioCodec.codec);
 
+                            var profiles = audioCodec.profiles.join('|');
+                            var maxChannels = audioCodec.maxChannels;
```
nielsvanvelzen, line 260: No need to assign this to a variable first IMO. Can just use `audioCodec.maxChannels` in the 2 places that use it. Same for maxSampleRate.

Maxr1998 (changes requested):

app/src/main/assets/native/exoplayer.js
nielsvanvelzen: profiles can stay as-is since it manipulates the data (.join function)

app/src/main/java/org/jellyfin/mobile/player/ExoPlayerFormats.kt
```diff
@@ -271,6 +271,23 @@ object ExoPlayerFormats {
             else -> null
         }
     }
+
+    fun getAudioProfile(codec: String, profile: Int): String? =  when(codec) {
```
nielsvanvelzen, line 275:
```suggestion
    fun getAudioProfile(codec: String, profile: Int): String? = when(codec) {
```

app/src/main/java/org/jellyfin/mobile/player/ExoPlayerFormats.kt
```diff
@@ -271,6 +271,23 @@ object ExoPlayerFormats {
...
+        "aac" -> getAACProfile(profile)
+        else -> null
+    }
+
+    private fun getAACProfile(profile: Int): String? = when(profile) {
+        MediaCodecInfo.CodecProfileLevel.AACObjectELD -> "ELD"
```
nielsvanvelzen, line 281: Maybe add an import for `MediaCodecInfo.CodecProfileLevel.*`

nielsvanvelzen (changes requested):

nielsvanvelzen:
Unexpectedly @Maxr1998 and I were reviewing at the same time haha 😂

app/src/main/assets/native/exoplayer.js
vitorsemeano: hm imo i prefer this approach, because if you need to change that name, you would need to change in 2 lines  (for this case) instead of just one. But if you insist, i can place in those 2 lines.

app/src/main/java/org/jellyfin/mobile/player/ExoPlayerFormats.kt
Maxr1998: @nielsvanvelzen I'm not sure if that's such a good idea. I think an import for `MediaCodecInfo.CodecProfileLevel` would be enough, requiring to write e.g. `CodecProfileLevel.AACObjectELD` afterwards.

Maxr1998:
@nielsvanvelzen I'd appreciate if you could take another look, from my side it's good, also tested the changes locally. Would rebase the PR once you reviewed, and then we can merge it.

app/src/main/assets/native/exoplayer.js
Maxr1998: I also think it's preferable, and it's only a small code style thing anyway.

nielsvanvelzen (approved):

Maxr1998 (approved):

PR #190, opened 2020-10-12
Add zoom feature to ExoPlayer

Maxr1998:
This allows to zoom in on movies with e.g. a wide aspect ratio (21:9) that are embedded into a 16:9 frame, as phones nowadays often have taller screens (ultra-wide in landscape).

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/player/PlayerActivity.kt
+++ b/app/src/main/java/org/jellyfin/mobile/player/PlayerActivity.kt
@@ -19,6 +19,7 @@ import androidx.core.view.isVisible
 import androidx.core.view.postDelayed
 import androidx.core.view.updatePadding
 import com.google.android.exoplayer2.Player
+import com.google.android.exoplayer2.ui.AspectRatioFrameLayout
 import com.google.android.exoplayer2.ui.PlayerView
 import org.jellyfin.mobile.AppPreferences
 import org.jellyfin.mobile.R
@@ -215,14 +216,14 @@ class PlayerActivity : AppCompatActivity() {
 
     @SuppressLint("ClickableViewAccessibility")
     private fun setupGestureDetector() {
-        // Handle tap when controls are locked
+        // Handles taps when controls are locked
         val unlockDetector = GestureDetector(this, object : GestureDetector.SimpleOnGestureListener() {
             override fun onSingleTapConfirmed(e: MotionEvent?): Boolean {
                 peekUnlockButton()
                 return true
             }
         })
-        // Handle double tap gesture on controls
+        // Handles double tap to seek and brightness/volume gestures
         val gestureDetector = GestureDetector(this, object : GestureDetector.SimpleOnGestureListener() {
             override fun onDoubleTap(e: MotionEvent): Boolean {
                 val viewWidth = playerView.measuredWidth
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/jellyfin/mobile/player/PlayerActivity.kt
```diff
@@ -317,9 +318,30 @@ class PlayerActivity : AppCompatActivity() {
...
         })
+        // Handles scale/zoom gesture
+        val zoomGestureDetector = ScaleGestureDetector(this, object : ScaleGestureDetector.OnScaleGestureListener {
+            override fun onScaleBegin(detector: ScaleGestureDetector): Boolean = true
+
+            override fun onScale(detector: ScaleGestureDetector): Boolean {
+                val scaleFactor = detector.scaleFactor
+                if (abs(scaleFactor - 1f) > 0.01f) {
+                    playerView.resizeMode = if (scaleFactor > 1) AspectRatioFrameLayout.RESIZE_MODE_ZOOM else AspectRatioFrameLayout.RESIZE_MODE_FIT
+                }
```
nielsvanvelzen, line 329:
```suggestion
                if (abs(detector.scaleFactor - 1f) > 0.01f) {
                    playerView.resizeMode = if (detector.scaleFactor > 1) AspectRatioFrameLayout.RESIZE_MODE_ZOOM else AspectRatioFrameLayout.RESIZE_MODE_FIT
                }
```

app/src/main/java/org/jellyfin/mobile/player/PlayerActivity.kt
```diff
@@ -317,9 +318,30 @@ class PlayerActivity : AppCompatActivity() {
...
+                return true
+            }
+
+            override fun onScaleEnd(detector: ScaleGestureDetector) {}
+        }).apply { isQuickScaleEnabled = false }
```
nielsvanvelzen, line 334:
```suggestion
        }).apply { 
            isQuickScaleEnabled = false
        }
```

app/src/main/java/org/jellyfin/mobile/player/PlayerActivity.kt
```diff
@@ -317,9 +318,30 @@ class PlayerActivity : AppCompatActivity() {
...
 
         playerView.setOnTouchListener { _, event ->
-            if (playerView.useController) gestureDetector.onTouchEvent(event) else unlockDetector.onTouchEvent(event)
+            if (playerView.useController) {
+                when (event.pointerCount) {
+                    1 -> gestureDetector.onTouchEvent(event)
+                    // Apparently the scale gesture detector doesn't enforce a maximum finger count.
+                    // If anyone scaled with more than 5 fingers, I'd be worried.
+                    in 2..5 -> zoomGestureDetector.onTouchEvent(event)
```
nielsvanvelzen, line 342: I think we should only allow zooming with 2 fingers

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/mobile/player/PlayerActivity.kt
Maxr1998: Maybe :sweat_smile:

app/src/main/java/org/jellyfin/mobile/player/PlayerActivity.kt
Maxr1998: I had that before, but I didn't like the layout. Maybe I just set it on the property instead of using `apply`?

app/src/main/java/org/jellyfin/mobile/player/PlayerActivity.kt
Maxr1998: Does that make such a difference? :thinking:
Actually, the extra variable ensures we get the same value for both accesses, and need less function calls.

nielsvanvelzen: It shouldn't matter that much, it's more like a personal preference.

nielsvanvelzen (approved):

PR #189, opened 2020-10-11
Handle fragment destroy properly to prevent leaks

CarlosOlivo:
From https://developer.android.com/topic/libraries/view-binding#fragments

> Note: Fragments outlive their views. Make sure you clean up any references to the binding class instance in the fragment's onDestroyView() method.

PR #201, opened 2020-10-18
Use view binding in PlayerActivity

Maxr1998:


CarlosOlivo:
#189

Maxr1998:
> #189 

Not needed (yet) because this is still an Activity. I will switch this to a fragment in another PR, wanted to split this up a bit.

nielsvanvelzen (approved):

PR #204, opened 2020-10-21
Update Activity-ktx to 1.2.0-alpha01

CarlosOlivo:
[Activity-ktx Version 1.2.0-alpha01](https://developer.android.com/jetpack/androidx/releases/activity#1.2.0-alpha01)

Bug Fixes

- Fixed a regression introduced in Activity 1.1.0 when running on older versions of the platform where onBackPressed() would cause an IllegalStateException due to a bug in the android.app.FragmentManager. ([b/146290338](https://issuetracker.google.com/issues/146290338))

### DIFF
--- a/buildSrc/src/main/kotlin/Dependencies.kt
+++ b/buildSrc/src/main/kotlin/Dependencies.kt
@@ -13,7 +13,7 @@ object Dependencies {
         const val koin = "2.1.6"
         const val appCompat = "1.1.0"
         const val androidxCore = "1.4.0-alpha01"
-        const val activity = "1.1.0"
+        const val activity = "1.2.0-alpha01"
         const val fragment = "1.2.5"
         const val exoPlayer = "2.12.0"
 


Maxr1998:
Interesting, I've had this happen to me as well, but (at least partially) worked around this in #194.
I wish the stable versions of AndroidX libraries were actually stable one time, or such fixes were backported to new stable releases so that you don't have to sign up for possible additional bugs in the alphas..

nielsvanvelzen:
If we're using an unstable version of the library I'd say we use the latest; 1.2.0-beta01

https://developer.android.com/jetpack/androidx/releases/activity#1.2.0-beta01

CarlosOlivo:
Between versions `1.2.0-alpha01` and `1.2.0-beta01` there are several changes of behavior and breaking changes, also if the beta is used it would be necessary to update to the beta/alpha versions of `appcomponent` and `fragment` needed by the lint check

nielsvanvelzen (approved):
It's fine for now but we should update it ASAP when 1.2.x stable releases

Maxr1998 (approved):

PR #205, opened 2020-10-23
Improved connect logging

nielsvanvelzen:
Added logging to hopefully help debug connectivity issues. Also makes the version check a bit more strict and requires 10.6.0 >=

image

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/fragment/ConnectFragment.kt
+++ b/app/src/main/java/org/jellyfin/mobile/fragment/ConnectFragment.kt
@@ -196,11 +196,20 @@ class ConnectFragment : Fragment() {
 
         val isValidInstance = when {
             version.size != 3 -> false
-            version[0] == PRODUCT_NAME_SUPPORTED_SINCE.first && version[1] < PRODUCT_NAME_SUPPORTED_SINCE.second -> true // Valid old version
+            // Major version is invalid
+            version[0] != PRODUCT_NAME_SUPPORTED_SINCE.first -> false
+            // Minor version is too old
+            version[1] < PRODUCT_NAME_SUPPORTED_SINCE.second -> false
             else -> true // FIXME: check ProductName once API client supports it
         }
+
         Timber.i("Server at $urls with version ${serverInfo.version} valid: $isValidInstance")
 
-        return if (isValidInstance) httpUrl else null
+        if (!isValidInstance) {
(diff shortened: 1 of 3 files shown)


nielsvanvelzen:
We should add a version check on app launch too in the future, as right now you can add a server with version 10.6.x but when you update the app and it requires 10.7.x it still connects happily to a server already configured.

app/src/main/java/org/jellyfin/mobile/fragment/ConnectFragment.kt
```diff
@@ -190,10 +196,20 @@ class ConnectFragment : Fragment() {
...
-            version[0] == PRODUCT_NAME_SUPPORTED_SINCE.first && version[1] < PRODUCT_NAME_SUPPORTED_SINCE.second -> true // Valid old version
+            // Major version is invalid
+            version[0] != PRODUCT_NAME_SUPPORTED_SINCE.first -> false
+            // Minor version is too old
+            version[1] < PRODUCT_NAME_SUPPORTED_SINCE.second -> false
             else -> true // FIXME: check ProductName once API client supports it
         }
 
-        return if (isValidInstance) httpUrl else null
+        Timber.i("Server at $urls with version ${serverInfo.version} valid: $isValidInstance")
```
nielsvanvelzen, line 206:
```suggestion
        Timber.i("Server at $httpUrl with version ${serverInfo.version} valid: $isValidInstance")
```

oops

Maxr1998:
I don't know whether we want to leak the server url in the production version, I'd prefer a `Timber.d` for the logging calls. Rest looks good, gonna review once I'm at the PC.

nielsvanvelzen:
We could mask the hostname but in the end the user is responsible for making sure to not leak the address. The apiclient, webview and a bunch of other components also print it.

Maxr1998:
Also in the non-debug version?

nielsvanvelzen:
Yup

Maxr1998 (approved):
LGTM, if you fix the thing you commented on, I'll merge.

nielsvanvelzen:
I was too lazy to open Android Studio so can just squash merge it

Maxr1998:
Hehe, ok, gonna merge once the pipeline finishes :+1:

Maxr1998:
Effin' Azure :angry:
It builds fine for me locally, so I'll merge this anyways.

PR #213, opened 2020-10-28
Do not detect gestures in top of screen

jorritsmit:
I had the issue that when i want to pull down the android top bar (e.g. to look at the time), i always lower brightness or volume. This pull request should fix that. The top 48dp (twice the top bar size)


Maxr1998 (approved):
I rebased the changes onto the latest master and moved the constant into the separate Constants file. Thanks for your contribution!

PR #221, opened 2020-10-31
Migrate server and user info to database

Maxr1998:
Got a bit bigger than I would've liked, but I had to change quite a few things to get this in. Also fixed some other things in the `MediaService` that I noticed. E.g. the `sendError()` would always crash the app in this form, as it's only used for custom commands within `MediaBrowserCompatService`. Instead, we just return a null result.

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/MainActivity.kt
+++ b/app/src/main/java/org/jellyfin/mobile/MainActivity.kt
@@ -9,23 +9,24 @@ import android.os.IBinder
 import android.view.OrientationEventListener
 import androidx.appcompat.app.AppCompatActivity
 import androidx.coordinatorlayout.widget.CoordinatorLayout
-import okhttp3.HttpUrl.Companion.toHttpUrlOrNull
-import org.jellyfin.apiclient.interaction.ApiClient
+import androidx.lifecycle.lifecycleScope
+import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.launch
 import org.jellyfin.mobile.cast.Chromecast
 import org.jellyfin.mobile.cast.IChromecast
 import org.jellyfin.mobile.fragment.ConnectFragment
 import org.jellyfin.mobile.fragment.WebViewFragment
 import org.jellyfin.mobile.player.PlayerFragment
-import org.jellyfin.mobile.utils.PermissionRequestHelper
-import org.jellyfin.mobile.utils.SmartOrientationListener
-import org.jellyfin.mobile.utils.lazyView
-import org.jellyfin.mobile.utils.replaceFragment
+import org.jellyfin.mobile.utils.*
+import org.jellyfin.mobile.viewmodel.MainViewModel
+import org.jellyfin.mobile.viewmodel.ServerState
 import org.jellyfin.mobile.webapp.RemotePlayerService
 import org.koin.android.ext.android.inject
+import org.koin.android.viewmodel.ext.android.viewModel
 import org.koin.androidx.fragment.android.setupKoinFragmentFactory
 
 class MainActivity : AppCompatActivity() {
-    val apiClient: ApiClient by inject()
+    private val mainViewModel: MainViewModel by viewModel()
     val appPreferences: AppPreferences by inject()
     val chromecast: IChromecast = Chromecast()
     private val permissionRequestHelper: PermissionRequestHelper by inject()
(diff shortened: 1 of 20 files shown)


app/src/main/java/org/jellyfin/mobile/MainActivity.kt
```diff
@@ -53,12 +54,19 @@ class MainActivity : AppCompatActivity() {
...
-                    replaceFragment<WebViewFragment>()
-                } else {
-                    replaceFragment<ConnectFragment>()
+        lifecycleScope.launch {
+            mainViewModel.serverState.collect { state ->
+                with(supportFragmentManager) {
+                    when (state) {
+                        ServerState.Pending -> {
+                            // TODO add loading indicator
+                        }
```
Maxr1998, line 63: This can be ignored for now.

app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
```diff
@@ -24,27 +24,34 @@ import androidx.webkit.WebViewClientCompat
...
+import kotlin.coroutines.suspendCoroutine
 
 class WebViewFragment : Fragment() {
     val apiClient: ApiClient by inject()
-    private val appPreferences: AppPreferences by inject()
+    private val serverController: ServerController by inject()
     private val webappFunctionChannel: WebappFunctionChannel by inject()
     private val externalPlayer by lazy { ExternalPlayer(this) }
 
+    private var serverId: Long = 0
```
nielsvanvelzen, line 53: Why no lateinit for this one?

app/src/main/java/org/jellyfin/mobile/media/MediaService.kt
```diff
@@ -160,20 +159,18 @@ class MediaService : MediaBrowserServiceCompat() {
                                 quality = 90
                             })
 
-                            val description: MediaDescriptionCompat =
-                                MediaDescriptionCompat.Builder().apply {
-                                    setMediaId(MediaItemType.Library.toString() + "|" + item.id)
-                                    setTitle(item.name)
-                                    setIconUri(Uri.parse(itemImageUrl))
-                                }.build()
+                            val description: MediaDescriptionCompat = MediaDescriptionCompat.Builder().apply {
```
nielsvanvelzen, line 162:
```suggestion
                            val description = MediaDescriptionCompat.Builder().apply {
```

app/src/main/java/org/jellyfin/mobile/media/MediaService.kt
```diff
@@ -160,20 +159,18 @@ class MediaService : MediaBrowserServiceCompat() {
...
 
                             MediaItem(description, MediaItem.FLAG_BROWSABLE)
                         }
 
-                    result.sendResult(views.toMutableList())
-
+                    result.sendResult(views)
                 } else {
-                    result.sendError(null)
+                    result.sendResult(emptyList())
```
nielsvanvelzen, line 173: Why remove the error?

app/src/main/java/org/jellyfin/mobile/media/MediaService.kt
```diff
@@ -251,7 +248,7 @@ class MediaService : MediaBrowserServiceCompat() {
                     if (response != null) {
                         processItems(response.items)
                     } else {
-                        result.sendError(null)
+                        result.sendResult(null)
```
nielsvanvelzen, line 251: Why remove the error?

app/src/main/java/org/jellyfin/mobile/media/MediaService.kt
```diff
@@ -406,9 +403,7 @@ class MediaService : MediaBrowserServiceCompat() {
                     /**
                      * Unhandled view
                      */
-                    else -> {
-                        result.sendError(null)
-                    }
+                    else -> result.sendResult(null)
```
nielsvanvelzen, line 406: Why remove the error?

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/mobile/utils/Constants.kt
```diff
@@ -15,8 +15,12 @@ object Constants {
     const val SESSION_CAPABILITIES_PATH = "sessions/capabilities/full"
 
     const val FRAGMENT_CONNECT_EXTRA_ERROR = "error"
+    const val FRAGMENT_WEB_VIEW_EXTRA_SERVER_ID = "org.jellyfin.mobile.intent.extra.SERVER_ID"
```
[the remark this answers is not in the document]
Maxr1998: We should though! It's more consistent with the Android framework.

app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
Maxr1998: Cause lateinit doesn't work for primitives.

app/src/main/java/org/jellyfin/mobile/media/MediaService.kt
Maxr1998: See my comment on the PR :sweat_smile:
`sendError` isn't allowed in that function and is a wrong usage resulting in a crash. I'm actually fixing a bug here.

app/src/main/java/org/jellyfin/mobile/MainActivity.kt
Maxr1998: Nope, compose handles it differently. But I haven't decided how to do it yet. The branch is meant mostly to have the when exhaustive.

app/src/main/java/org/jellyfin/mobile/utils/Constants.kt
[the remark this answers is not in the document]
nielsvanvelzen: Please change it for the others too then

app/src/main/java/org/jellyfin/mobile/media/MediaService.kt
nielsvanvelzen: oooh nice

app/src/main/java/org/jellyfin/mobile/media/MediaService.kt
Maxr1998: Not even part of my changes, lol. Still, will fix that.

nielsvanvelzen (approved):

PR #225, opened 2020-11-04
Fixes in setting JELLYFIN_VERSION

nielsvanvelzen:
Not tested, probably doesn't work (see my comment on line 29)

### DIFF
--- a/.ci/azure-pipelines.yml
+++ b/.ci/azure-pipelines.yml
@@ -1,5 +1,7 @@
 variables:
-  - group: 'jellyfin'
+  group: 'jellyfin'
+  ${{ if startsWith(variables['Build.SourceBranch'], 'refs/tags/v') }}:
+    JELLYFIN_VERSION: $[replace(variables['Build.SourceBranch'], 'refs/tags/', '')]
 
 trigger:
   batch: true
@@ -22,11 +24,6 @@ jobs:
 
     pool:
       vmImage: 'ubuntu-latest'
-
-    variables:
-      - name: JELLYFIN_VERSION
-        value: replace(variables['Build.SourceBranch'], 'refs/tags/', '')
-        condition: startsWith(variables['Build.SourceBranch'], 'refs/tags/v')
 
     steps:
       - task: Gradle@2
@@ -92,9 +89,6 @@ jobs:
 
     pool:
       vmImage: 'ubuntu-latest'
-
-    variables:
-      JELLYFIN_VERSION: replace(variables['Build.SourceBranch'], 'refs/tags/', '')
 
     steps:
       - task: DownloadPipelineArtifact@2
--- a/.ci/azure-pipelines.yml
+++ b/.ci/azure-pipelines.yml
@@ -23,6 +23,11 @@ jobs:
     pool:
       vmImage: 'ubuntu-latest'
 
+    variables:
+      - name: JELLYFIN_VERSION
+        value: replace(variables['Build.SourceBranch'], 'refs/tags/', '')
+        condition: startsWith(variables['Build.SourceBranch'], 'refs/tags/v')
+
     steps:
       - task: Gradle@2
         displayName: 'Run Tests'
@@ -42,10 +47,6 @@ jobs:
       vmImage: 'ubuntu-latest'
 
     steps:
-      - script: 'echo "##vso[task.setvariable variable=JELLYFIN_VERSION]$(git describe --tags)"'
-        condition: startsWith(variables['Build.SourceBranch'], 'refs/tags')
-        displayName: 'Set Version Variable'
-
       - task: Gradle@2
         displayName: 'Build Debug'
         inputs:
@@ -87,15 +88,15 @@ jobs:
     displayName: 'Publish'
 
     dependsOn: Build
-    condition: startsWith(variables['Build.SourceBranch'], 'refs/tags')
+    condition: startsWith(variables['Build.SourceBranch'], 'refs/tags/v')
 
     pool:
       vmImage: 'ubuntu-latest'
 
-    steps:
-      - script: 'echo "##vso[task.setvariable variable=JELLYFIN_VERSION]$(git describe --tags)"'
-        displayName: 'Set Version Variable'
+    variables:
+      JELLYFIN_VERSION: replace(variables['Build.SourceBranch'], 'refs/tags/', '')
 
+    steps:
       - task: DownloadPipelineArtifact@2
         displayName: 'Download APKs'
         inputs:


.ci/azure-pipelines.yml
```diff
@@ -23,6 +23,11 @@ jobs:
     pool:
       vmImage: 'ubuntu-latest'
 
+    variables:
+      - name: JELLYFIN_VERSION
+        value: replace(variables['Build.SourceBranch'], 'refs/tags/', '')
+        condition: startsWith(variables['Build.SourceBranch'], 'refs/tags/v')
```
nielsvanvelzen, line 29: It's not possible to place a condition on a variable definition.

nielsvanvelzen: @EraYaN do you have some pointer for me on how to set this variable only when the condition evaluates to true? I think we should use a script block for that (like it was before) but I'm not sure what the syntax would be for setting the variable with to the result of an Azure expression.

h1dden-da3m0n: May I ask why set the Variable in any job besides `Publish` anyway? (It does not seem to be used besides in the `Publish` step job)

EraYaN: Just put the `replace(variables['Build.SourceBranch'], 'refs/tags/', '')` bit in the place it's used. That should be only executed when the condition is true.

.ci/azure-pipelines.yml
```diff
@@ -23,6 +23,11 @@ jobs:
...
+
```
h1dden-da3m0n, line 30:
```suggestion
```

.ci/azure-pipelines.yml
```diff
@@ -87,15 +88,15 @@ jobs:
...
+    condition: startsWith(variables['Build.SourceBranch'], 'refs/tags/v')
 
     pool:
       vmImage: 'ubuntu-latest'
 
-    steps:
-      - script: 'echo "##vso[task.setvariable variable=JELLYFIN_VERSION]$(git describe --tags)"'
-        displayName: 'Set Version Variable'
+    variables:
+      JELLYFIN_VERSION: replace(variables['Build.SourceBranch'], 'refs/tags/', '')
```
h1dden-da3m0n, line 97:
```suggestion
      - name: JELLYFIN_VERSION
        value: $[replace(variables['Build.SourceBranch'],'refs/tags/v','')]
```

.ci/azure-pipelines.yml
```diff
@@ -87,15 +88,15 @@ jobs:
     displayName: 'Publish'
 
     dependsOn: Build
-    condition: startsWith(variables['Build.SourceBranch'], 'refs/tags')
+    condition: startsWith(variables['Build.SourceBranch'], 'refs/tags/v')
```
h1dden-da3m0n, line 91:
```suggestion
    condition: startsWith(variables['Build.SourceBranch'], 'refs/tags')
```
the `/v` should not be required unless you intent to ever have a tag that is not meant to trigger the `publish` job

h1dden-da3m0n (commented):
This should make this CI pipeline more aligned to what the Java API-Client CI looks like :wink:

.ci/azure-pipelines.yml
nielsvanvelzen: The app is slightly different from the library. In the case of the app we need the tag set when building to include the version in the AndroidManifest, this is not needed for the library.
The build task should only have the variable set when it's building for a tag, not when building for PR's/master.

At line 97 I didn't need the condition because it's defined at the job-level. I do need a way to only create the variable on this job though.

.ci/azure-pipelines.yml
nielsvanvelzen: This was a suggestion by @EraYaN so the publishing doesn't run for non-version tags (if we ever use them). Better be safe than sorry!

.ci/azure-pipelines.yml
h1dden-da3m0n: as mentioned by @EraYaN and me in the conversation above, this should only be set in the job that needs it and that anyway is already only run on condition

nielsvanvelzen: See my other comment on why we need it for the app but not the apiclient

.ci/azure-pipelines.yml
h1dden-da3m0n: I see, let me check the Azure pipelines docu maybe i find something useful here :wink:

.ci/azure-pipelines.yml
h1dden-da3m0n: I see, but I did comment that too ;) (! suspected that be the reason behind this change, just wanted to be sure)

.ci/azure-pipelines.yml
h1dden-da3m0n: @nielsvanvelzen Is the variable already needed in the `Test` job or does the `Build` job?

nielsvanvelzen: Only in the build job and publish jobs not for test (although if it's easier to define it top-level it won't hurt)

EraYaN: You can use YAML if constructs to define it I guess

EraYaN: (Use 1 level of extra indent after the if line)

.ci/azure-pipelines.yml
```diff
@@ -42,10 +47,6 @@ jobs:
       vmImage: 'ubuntu-latest'
 
     steps:
-      - script: 'echo "##vso[task.setvariable variable=JELLYFIN_VERSION]$(git describe --tags)"'
-        condition: startsWith(variables['Build.SourceBranch'], 'refs/tags')
-        displayName: 'Set Version Variable'
-
```
h1dden-da3m0n, line 48: Well GH does not allow for suggestions on deleted lines :expressionless: 

```yaml
- displayName: 'Set Job Scoped Variable (used by gradle)'
  condition: startsWith(variables['Build.SourceBranch'], 'refs/tags/v')
  script: 'echo "##vso[task.setvariable variable=JELLYFIN_VERSION]$[replace(variables['Build.SourceBranch'], 'refs/tags/', '')]"'
```

NOTE: I am not 100% sure if this will work, but according to the Azure Pipeline docs I am confident it should. 
Reason for me thinking this should work:
* $[\<expression\>] is a runtime expression, that gets processed during runtime (see [here](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#understand-variable-syntax))
* ##vso[task.setvariable variable=JELLYFIN_VERSION] is the way already used to set the Variable for the downstream steps (see [here](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#set-a-job-scoped-variable-from-a-script))

The only alternative that comes to my mind is duplicating the `Build` job :thinking:

.ci/azure-pipelines.yml
nielsvanvelzen: I now moved the declaration of the variable top-level with the if-statement so it is available in all jobs/steps. This way there is no duplicate code when we need it in multiple jobs.

.ci/azure-pipelines.yml
h1dden-da3m0n: probably ignore this, I think @EraYaN is the correct way to solve this :P

.ci/azure-pipelines.yml
```diff
@@ -1,5 +1,8 @@
 variables:
-  - group: 'jellyfin'
+  group: 'jellyfin'
+  ${{ if startsWith(variables['Build.SourceBranch'], 'refs/tags/v') }}:
+    - name: JELLYFIN_VERSION
+      value: replace(variables['Build.SourceBranch'], 'refs/tags/', '')
```
h1dden-da3m0n, line 5: If I read the docks mentioned by @EraYaN right then you need to use this syntax here:
```suggestion
   JELLYFIN_VERSION: replace(variables['Build.SourceBranch'], 'refs/tags/', '')
```

nielsvanvelzen: I _think_ what Azure does is remove the if statement and place the yaml in the "body" where the if-statement was. I probably forgot to add back the "- " before the group definition. I'll try that first and otherwise test this change.

nielsvanvelzen: I might have fixed it now!

nielsvanvelzen: It doesn't work because I think Azure is setting the value to empty now..

h1dden-da3m0n: u might need the expression syntax on the assignment sight tbh.
e.g. `JELLYFIN_VERSION: $[replace(variables['Build.SourceBranch'], 'refs/tags/', '')]`

nielsvanvelzen: Even if that is an issue, it should've build this PR as it is not building a tag right now, but I'll try

h1dden-da3m0n: there is also [this](https://stackoverflow.com/a/57532526) xD

.ci/azure-pipelines.yml
```diff
@@ -1,5 +1,7 @@
 variables:
   - group: 'jellyfin'
+  - ${{ if startsWith(variables['Build.SourceBranch'], 'refs/tags/v') }}:
```
h1dden-da3m0n, line 3:
```suggestion
  group: 'jellyfin'
  ${{ if startsWith(variables['Build.SourceBranch'], 'refs/tags/v') }}:
```

nielsvanvelzen: I don't think we can do that since "group" is an environment variable but telling Azure to use the vars defined in a group

h1dden-da3m0n: hmmm, well my other guess, based on the log, would be this: 
```suggestion
  - group: 'jellyfin'
   ${{ if startsWith(variables['Build.SourceBranch'], 'refs/tags/v') }}:
     - JELLYFIN_VERSION: $[replace(variables['Build.SourceBranch'], 'refs/tags/', '')]
```

EraYaN: The if statement is basically a pre-processor thing, so it disappears and puts the indented block beneath it in it's place.

nielsvanvelzen: Seems to be working now but we can't now for sure until we publish the next release

.ci/azure-pipelines.yml
EraYaN: I mean it seems to work right now? no publish and a build with the 0.0.0-dev version as expected.

EraYaN (approved):

h1dden-da3m0n (approved):

Maxr1998 (approved):

PR #228, opened 2020-11-06
Fix version code calculation comment

nielsvanvelzen:
Each part of the semver should be able to use 2 digits. It didn't do that for PATCH.

Kotlin-playground test: https://pl.kotl.in/kwvS88jHC


nielsvanvelzen:
@Artiume I think you mentioned this PR per accident in the know issues list fyi

Maxr1998 (approved):
LGTM, you could maybe still remove leading zeros from the version codes.

PR #232, opened 2020-11-11
Improve and rework MediaService & Android Auto integration

### LINKED ISSUE #226: Android Auto Compliance Issues
In our latest review from Google, some issues were highlighted that are preventing publishing of the app.

Here is the list:

**No Auto functionality**
The app did not appear in the Auto screen or lens picker when we attempted to test functionality.
>_For example, your app does not contain an Intent filter for action "android.media.action.MEDIA_PLAY_FROM_SEARCH"._

**App does not perform as expected**
Your app does not perform all functions properly or as expected from a user perspective.
>_For example, audio tracks are silent when the microphone is not active._

**Driver distraction features**
Your app includes features that distract the driver.
>_For example, your app does not pause or reduce the volume of the audio being played while the microphone is active._

Maxr1998:
- Moves library browser code into extra class and package, as it will only be used for Android Auto.
- The loader function was split into multiple subfunctions and cleaned up.
- The format of 'parentIds' was improved, unexpected behaviour with queue selection fixed.
- Shuffling was temporarily removed and will be handled differently in the future.
- Also introduces preliminaries for casting and Android 11 seamless transfer, this isn't fully enabled yet however - it seems there are still some changes necessary within the AndroidX libraries.

Uses code from the android/uamp reference media player project: https://github.com/android/uamp

Closes #226.

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/media/MediaExtensions.kt
+++ b/app/src/main/java/org/jellyfin/mobile/media/MediaExtensions.kt
@@ -1,0 +1,125 @@
+@file:Suppress("NOTHING_TO_INLINE")
+
+package org.jellyfin.mobile.media
+
+import android.graphics.Bitmap
+import android.net.Uri
+import android.support.v4.media.MediaMetadataCompat
+import androidx.core.net.toUri
+
+/**
+ * Useful extensions for [MediaMetadataCompat].
+ */
+inline val MediaMetadataCompat.mediaId: String?
+    get() = getString(MediaMetadataCompat.METADATA_KEY_MEDIA_ID)
+
+inline val MediaMetadataCompat.title: String?
+    get() = getString(MediaMetadataCompat.METADATA_KEY_TITLE)
+
+inline val MediaMetadataCompat.artist: String?
+    get() = getString(MediaMetadataCompat.METADATA_KEY_ARTIST)
+
+inline val MediaMetadataCompat.duration
+    get() = getLong(MediaMetadataCompat.METADATA_KEY_DURATION)
+
+inline val MediaMetadataCompat.album: String?
+    get() = getString(MediaMetadataCompat.METADATA_KEY_ALBUM)
+
+inline val MediaMetadataCompat.author: String?
+    get() = getString(MediaMetadataCompat.METADATA_KEY_AUTHOR)
+
+inline val MediaMetadataCompat.writer: String?
+    get() = getString(MediaMetadataCompat.METADATA_KEY_WRITER)
+
+inline val MediaMetadataCompat.composer: String?
+    get() = getString(MediaMetadataCompat.METADATA_KEY_COMPOSER)
+
+inline val MediaMetadataCompat.compilation: String?
+    get() = getString(MediaMetadataCompat.METADATA_KEY_COMPILATION)
+
+inline val MediaMetadataCompat.date: String?
+    get() = getString(MediaMetadataCompat.METADATA_KEY_DATE)
+
+inline val MediaMetadataCompat.year: String?
+    get() = getString(MediaMetadataCompat.METADATA_KEY_YEAR)
+
+inline val MediaMetadataCompat.genre: String?
+    get() = getString(MediaMetadataCompat.METADATA_KEY_GENRE)
+
+inline val MediaMetadataCompat.trackNumber
+    get() = getLong(MediaMetadataCompat.METADATA_KEY_TRACK_NUMBER)
+
+inline val MediaMetadataCompat.trackCount
+    get() = getLong(MediaMetadataCompat.METADATA_KEY_NUM_TRACKS)
+
+inline val MediaMetadataCompat.discNumber
+    get() = getLong(MediaMetadataCompat.METADATA_KEY_DISC_NUMBER)
+
+inline val MediaMetadataCompat.albumArtist: String?
+    get() = getString(MediaMetadataCompat.METADATA_KEY_ALBUM_ARTIST)
+
+inline val MediaMetadataCompat.art: Bitmap
+    get() = getBitmap(MediaMetadataCompat.METADATA_KEY_ART)
+
+inline val MediaMetadataCompat.artUri: Uri
+    get() = this.getString(MediaMetadataCompat.METADATA_KEY_ART_URI).toUri()
+
+inline val MediaMetadataCompat.albumArt: Bitmap?
+    get() = getBitmap(MediaMetadataCompat.METADATA_KEY_ALBUM_ART)
+
+inline val MediaMetadataCompat.albumArtUri: Uri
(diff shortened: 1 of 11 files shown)


Maxr1998:
Note: the code for search is untested, because Google Assistant is dumb and doesn't properly recognize queries directed at Jellyfin.
I tested the rest of the app both with the Android Auto phone app and within an actual car, and thankfully didn't observe any problems (apart from search which isn't Jellyfin's fault).

nielsvanvelzen:
You can use [this app](https://github.com/googlesamples/android-media-controller) to test the integration - need to build it yourself though.

Maxr1998:
Oh wait, I already have that installed! Thanks for reminding me 😅

app/src/main/java/org/jellyfin/mobile/media/MediaService.kt
```diff
@@ -529,28 +309,122 @@ class MediaService : MediaBrowserServiceCompat() {
...
+     * Listen for events from ExoPlayer.
+     */
+    private inner class PlayerEventListener : Player.EventListener {
+        override fun onPlayerStateChanged(playWhenReady: Boolean, playbackState: Int) {
+            when (playbackState) {
+                Player.STATE_BUFFERING,
+                Player.STATE_READY -> {
+                    notificationManager.showNotificationForPlayer(currentPlayer)
+                    if (playbackState == Player.STATE_READY) {
+                        // TODO: When playing/paused save the current media item in persistent storage
```
nielsvanvelzen, line 345: TODO for now or future?

app/src/main/java/org/jellyfin/mobile/media/car/LibraryBrowser.kt
```diff
@@ -0,0 +1,392 @@
...
+            parentId = libraryId
+            includeItemTypes = arrayOf(BaseItemType.Audio.name)
+            filters = arrayOf(ItemFilter.IsPlayed)
+            sortBy = arrayOf(ItemSortBy.DatePlayed)
+            sortOrder = SortOrder.Descending
+            recursive = true
+            imageTypeLimit = 1
+            enableImageTypes = arrayOf(ImageType.Primary)
+            enableTotalRecordCount = false
+            limit = 100
```
nielsvanvelzen, line 241: What happens when I have 100+ recents/albums/artists/genres/playlists?

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/mobile/media/car/LibraryPage.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * A list of playlists
+     */
+    const val PLAYLISTS = "playlists"
+
+    // Content types for individual library items - albums, artists, playlists and a shuffle meta item
+
+    const val ALBUM = "album"
+    const val PLAYLIST = "playlist"
```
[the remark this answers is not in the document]
Maxr1998: There's  a shared comment above it :P But it's outdated, so I should fix it anyways xD

app/src/main/java/org/jellyfin/mobile/media/MediaExtensions.kt
```diff
@@ -0,0 +1,125 @@
...
+inline val MediaMetadataCompat.displayIconUri: Uri
+    get() = this.getString(MediaMetadataCompat.METADATA_KEY_DISPLAY_ICON_URI).toUri()
+
+inline val MediaMetadataCompat.mediaUri: Uri
+    get() = this.getString(MediaMetadataCompat.METADATA_KEY_MEDIA_URI).toUri()
+
+inline val MediaMetadataCompat.downloadStatus
+    get() = getLong(MediaMetadataCompat.METADATA_KEY_DOWNLOAD_STATUS)
+
+inline fun MediaMetadataCompat.Builder.setMediaId(id: String): MediaMetadataCompat.Builder =
```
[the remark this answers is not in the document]
Maxr1998: Cause values require a getter, and the Builder doesn't provider one.

app/src/main/java/org/jellyfin/mobile/media/car/LibraryPage.kt
```diff
@@ -0,0 +1,59 @@
...
+    /**
+     * List of music libraries that the user can access (referred to as "user views" in Jellyfin)
+     */
+    const val LIBRARIES = "libraries"
+
+    /**
+     * Special flag for use with [EXTRA_RECENT][androidx.media.MediaBrowserServiceCompat.BrowserRoot.EXTRA_RECENT],
+     * loads and plays the last played queue - basically resuming playback from last session
+     */
+    const val RECENT = "recent"
```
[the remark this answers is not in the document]
Maxr1998: Yeah, the naming is not ideal. `RESUME` sounds good :+1:

app/src/main/java/org/jellyfin/mobile/media/MediaService.kt
Maxr1998: Definitely future. Need a database for that first.

app/src/main/java/org/jellyfin/mobile/media/car/LibraryBrowser.kt
Maxr1998: You don't see everything ¯\_(ツ)_/¯

But it isn't only a technical limitation, it's also kinda inteded - Google recommends not overwhelming the user with too many items on the same page, that's also why they don't implement any support for paging or async list loading.

Spotify handles it pretty much the same in Android Auto.

nielsvanvelzen: Hmm weird, because that way an artist starting with "z" might be missing because you have large library

Maxr1998: That's why I sort by `DatePlayed` to still get relevant results :wink:

nielsvanvelzen: I suppose it's not that big a problem when there's search / shuffle buttons

nielsvanvelzen:
/azp run


we should look into these build issues 😒

app/src/main/java/org/jellyfin/mobile/media/MediaService.kt
nielsvanvelzen, line 345: Can you make an issue so we don't forget this TODO?

nielsvanvelzen (approved):

app/src/main/java/org/jellyfin/mobile/media/MediaService.kt
Maxr1998: Done.

PR #236, opened 2020-11-16
Fix bad device profile

nielsvanvelzen:
Caused JSON parsing errors in 10.7 >= and thus playback not working

this condition literally never worked because it used an non-existing operator and property lol


Maxr1998 (approved):
LGTM

PR #240, opened 2020-11-18
Plugins modified to support es6 migration

vitorsemeano:
This PR introduces changes to exoplayer and externalplayer plugin, modifying them into classes and expose them in window.

Each plugin will have access to a set of dependencies in its constructor (invoked by jf-web).

also the list of plugins were modified so it can find these plugins in window.

This is still not tested, so i will try to test it somehow with the changes present in https://github.com/jellyfin/jellyfin-web/pull/1994


nielsvanvelzen:
I think we should release one version of the Android app before 10.7 releases that supports both 10.6 and 10.7. A month after 10.7 we will remove the 10.6 support.

To achieve this goal I think we need to:
 - Load the .js files from a versioned directory, so we would get something like:  
   - /assets/native/10.6/exoplayer.js
   - /assets/native/10.7/exoplayer.js  
      This is not ideal because changes would need to made in both files - this is only temporary until we support 10.7>= only
- When the app starts it will always look up the version of the server it connects to and when the user connects to a server it will always look up the version of the server it connects to
  - Server 10.5<= show error
  - Server 10.6= load 10.6 assets, show toast to user to notify about 10.7
  - Server 10.7>= load 10.7 assets

This way we can keep supporting the 2 most used versions in the next few months.

CC @Maxr1998 what do you think about this plan? And should we do it for 2.2 or for 2.3?

Maxr1998:
That sounds good to me. I wouldn't be worried about duplicated files, as we'll later remove those anyways.
Re: which version, I don't really care. Depends on how long it'll take, having another version before this won't hurt though. So probably 2.3?

nielsvanvelzen:
If we do it in 2.3 we can confirm Google allows our changes for Android Auto this time. So I think that's a good approach.

Maxr1998:
> If we do it in 2.3 we can confirm Google allows our changes for Android Auto this time. So I think that's a good approach.

Yep, that's what I was thinking. Let's get that Android Auto out finally!

nielsvanvelzen:
Some quick testing for this PR:

 - https://github.com/jellyfin/jellyfin-android/blob/eff741c59786f01256a83eb8cf4070b67eb0ad5e/app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt#L132
  Need to add a new injection path here since `apploader.js` is gone. I guess it should match "main.bundle.js".

- With above change the basics work: change server / client settings buttons show up and the device information is set properly
- The video player hooks do not work

nielsvanvelzen:
I pushed some fixes that will hopefully load the plugins, need to wait on jellyfin/jellyfin-web#2139 and jellyfin/jellyfin-web#2137 to be merged so I can test though.

nielsvanvelzen:
Tested with demo.jellyfin.org/unstable and demo.jellyfin.org/stable for web player, native player and external player and everything seems to work.

The JavaScript in the 10.7 version is formatted with VSCode and I fixed some small issues like duplicate functions and switched from `new Promise()....` to async functions.

app/src/main/assets/native-10.7/ExoPlayerPlugin.js
```diff
@@ -0,0 +1,389 @@
...
+            src: this._currentSrc
+        };
+
+        this.events.trigger(this, 'stopped', [stopInfo]);
+        this._currentSrc = this._currentTime = null;
+    }
+
+    async getDeviceProfile() {
+        // using native player implementations, check if item can be played
+        // also check if direct play is supported, as audio is supported
```
Maxr1998, line 167: While we're at it, do you want to update this comment to something meaningful?

app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
```diff
@@ -129,15 +130,20 @@ class WebViewFragment : Fragment() {
                 val url = request.url
                 val path = url.path?.toLowerCase(Locale.ROOT) ?: return null
                 return when {
-                    path.endsWith(Constants.APPLOADER_PATH) -> {
+                    path.endsWith(Constants.APPLOADER_PATH) || path.endsWith(Constants.MAIN_BUNDLE_PATH) -> {
+                        if (path.endsWith(Constants.APPLOADER_PATH))
+                            assetsVersion = "10.6"
+                        else if (path.endsWith(Constants.MAIN_BUNDLE_PATH))
+                            assetsVersion = "10.7"
```
Maxr1998, line 137: Could make this a `when` as well, and have the else case throw an error.

Maxr1998 (changes requested):

app/src/main/assets/native-10.7/ExoPlayerPlugin.js
```diff
@@ -0,0 +1,389 @@
...
+    canSetAudioStreamIndex() {
+        return false;
+    }
+
+    setAudioStreamIndex(index) {
+        this.audioStreamIndex = index;
+    }
+
+    async changeSubtitleStream(index) {
+        // detach from the main ui thread
```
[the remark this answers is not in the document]
nielsvanvelzen: I removed it from others, forgot it in this one I guess

app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
nielsvanvelzen: Don't need an else clause because there's only 2 possible branches here considering we matched with one of both paths in the parent when branch

app/src/main/assets/native-10.7/ExoPlayerPlugin.js
nielsvanvelzen: I just removed it

Maxr1998 (approved):

PR #244, opened 2020-11-20
Remove the ExoPlayer cast extension

### LINKED ISSUE #243: libre-release.apk added non-libre elements
v2.2.0 just introduced [Firebase Data Transport](https://github.com/firebase/firebase-android-sdk/tree/master/transport), Google Mobile Services, Google Cast… bringing the `*libre-release.apk` even in size to `*proprietary-release.apk`. None of these were present in v2.1.3 – and none of these are "libre".

It hopefully was just an accident. Can you please "revert" that? Thanks in advance!

nielsvanvelzen:
The easy way to fix #243.

Another approach is to make the dependency on the cast extension proprietary only but this would require some changes in the MediaService because it won't compile without the library.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -118,7 +118,6 @@ dependencies {
 
     // Cast
     implementation(Dependencies.Cast.mediaRouter)
-    implementation(Dependencies.Cast.exoPlayerCastExtension)
     proprietaryImplementation(Dependencies.Cast.playServicesCast)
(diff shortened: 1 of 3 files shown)


Maxr1998:
Superseded by #250.

nielsvanvelzen:
I changed the description to automatically close this one :(

Maxr1998:
I only saw that afterwards :sweat_smile:

PR #256, opened 2020-11-23
Fix build reproducibility

nielsvanvelzen:
By enabling the ` vectorDrawables.useSupportLibrary` option the Android compiler will stop creating unnecessary png files.


nielsvanvelzen:
Fun fact: Gradle can generate 100mb >= of logs when you enable debug logs to figure out why a file is created by the build.

Maxr1998 (approved):

PR #258, opened 2020-12-03
Drop support for 10.6

Maxr1998:
We can possibly also merge this already, as potential fixes for 2.2.2 could be made to the 2.2.z branch.


nielsvanvelzen:
You can rename the assets folder back and remove the `-10.7` part (because it's probably the same source for 10.8).

nielsvanvelzen (approved):

PR #261, opened 2020-12-04
Fix JS injection breaking on page reloads in 10.7

Maxr1998:
Instead of only injecting when the main.bundle.js is requested, it attempts to inject for every *.bundle.js that is requested, and ensures one-time execution by setting a flag in window.

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
+++ b/app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
@@ -130,10 +130,10 @@ class WebViewFragment : Fragment() {
                 val url = request.url
                 val path = url.path?.toLowerCase(Locale.ROOT) ?: return null
                 return when {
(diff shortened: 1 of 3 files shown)


nielsvanvelzen:
I'm not a big fan of this approach. Opening the unstable demo and signing in loads **37** files matching "bundle.js". That means we inject a bit of JavaScript 37 times (although it only loads the nativeshell once).

Maxr1998:
Me neither, but at the same it's quite reliable from my testing and since it's async it doesn't really slow down the loading process. I could try to use the `/system/info/public` call from the JS apiclient for the hook, but I'm not sure whether the timing is suitable with that one.

Maxr1998:
I had the idea that we could use the `config.json` that's being pulled from the server as a hook, but that's already used on 10.6 - means we couldn't discern the versions if we used that. Honestly, instead of over-engineering this, I'd suggest to merge this as is and switch to the `config.json` hook when dropping 10.6 support.

nielsvanvelzen (approved):
Tested it in the emulator and it worked fine, it did inject 32 times but I guess it's not _that_ big of an issue.

PR #262, opened 2020-12-04
Add workflow config for GitHub CI

Maxr1998:
Set up GitHub CI for testing and building the app. No signing or publishing configs for now, only builds and stores the (unsigned) APKs as artifacts.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -9,7 +9,7 @@
 #
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
-org.gradle.jvmargs=-Xmx8192m
+org.gradle.jvmargs=-Xmx2G -XX:MaxPermSize=512m -XX:MaxMetaspaceSize=512m
 # AndroidX package structure to make it clearer which packages are bundled with the
 # Android operating system, and which are packaged with your app"s APK
 # https://developer.android.com/topic/libraries/support-library/androidx-rn
--- a/.ci/azure-pipelines.yml
+++ b/.ci/azure-pipelines.yml
@@ -12,11 +12,6 @@ trigger:
     include:
       - '*'
 
-pr:
-  branches:
-    include:
-      - '*'
-
 jobs:
   - job: Test
     displayName: 'Test'
--- a/.github/workflows/app-test.yaml
+++ b/.github/workflows/app-test.yaml
@@ -1,0 +1,30 @@
+name: Test App
+
+on:
+  push:
+    branches:
+      - master
+      - release-*
+  pull_request:
+
+jobs:
+  test:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Setup Java
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+      - name: Setup Gradle cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-
+      - name: Run test and assemble tasks
+        run: ./gradlew --no-daemon --build-cache test


h1dden-da3m0n:
:wave: since you removed the Azure pipeline run on PRs, may I ask when you intent to run the GHA workflow (currently you configured it to run on release publishes only?)

Maxr1998:
GitHub Actions runs for all commits right now, but it doesn't have the secrets configured yet to sign the APKs and publish them to the Jellyfin repo. We're testing GA "unofficially" for now, and will only fully migrate if also the rest of the projects decides to switch.

.github/workflows/app.yml
```diff
@@ -0,0 +1,36 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Cache Gradle wrapper and dependencies
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/wrapper/dists
+            ~/.gradle/caches
+          key: gradle-cache
```
h1dden-da3m0n, line 18: Any particular reason youu are not using the "default" `key` and `restore-key` for a Gradle build?

```yaml
- uses: actions/cache@v2
  with:
    path: |
      ~/.gradle/caches
      ~/.gradle/wrapper
    key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*') }}
    restore-keys: |
      ${{ runner.os }}-gradle-
```

h1dden-da3m0n: The reason I am asking is that the key you selected may work but you run into the risk of hitting the [Cache Limit](https://github.com/actions/cache#cache-limits) somewhere down the line.

.github/workflows/app.yml
```diff
@@ -0,0 +1,36 @@
+name: Test and build Jellyfin Android
+on:
+  push:
+  release:
+    types: [ published ]
```
h1dden-da3m0n, line 5: *Okay now in comment form*

This indeed will trigger on all pushes on all branches, yet I am pretty sure NOT on PRs, which is my confusion since you disabled the Azure pipeline for those.
Additionally you set this to run on Release publish (which confuses me a bit why, but is fine)

h1dden-da3m0n:
> GitHub Actions runs for all commits right now, but it doesn't have the secrets configured yet to sign the APKs and publish them to the Jellyfin repo on releases. We're testing GA "unofficially" for now, and will only fully migrate if also the rest of the org decides to switch.



> GitHub Actions runs for all commits right now, but it doesn't have the secrets configured yet to sign the APKs and publish them to the Jellyfin repo on releases. We're testing GA "unofficially" for now, and will only fully migrate if also the rest of the org decides to switch.

WAIT, do I read that right: you are playing with the idea to move from Azure Pipelines to GH Actions?

nielsvanvelzen:
We're slowly moving things to GitHub actions in multiple repositories.

h1dden-da3m0n:
I see, I'll try and chime in where I can I love doing CI :grin: (but  poor joshuaboniface and his sanity, if his comments regarding CI migrations go in other PRs)

Thank you for the info and I'll stopping going off-topic with this

.github/workflows/app.yml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Publish compiled APKs as artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: jellyfin-apks
+          path: app/build/outputs/apk/**/jellyfin-android-*.apk
+      - name: Publish version.txt as artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: version.txt
+          path: app/build/version.txt
```
nielsvanvelzen, line 36: I'd rather create 1 artifact with the apk files and version.txt together. Basically the "complete package" to upload to repo.jellyfin.org (like https://repo.jellyfin.org/releases/client/android/versions/v2.2.2/)

.github/workflows/app.yml
```diff
@@ -0,0 +1,36 @@
...
+          path: |
+            ~/.gradle/wrapper/dists
+            ~/.gradle/caches
+          key: gradle-cache
+      - name: Set environment variable with version
+        if: ${{ startsWith(github.ref, 'refs/tags/v') }}
+        run: |
+          echo 'JELLYFIN_VERSION<<EOF' >> $GITHUB_ENV
+          echo $GITHUB_REF | sed 's|refs/tags/||' >> $GITHUB_ENV
+          echo 'EOF' >> $GITHUB_ENV
```
nielsvanvelzen, line 24: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable

```suggestion
        run:  echo "JELLYFIN_VERSION=${$GITHUB_REF | sed 's|refs/tags/||'}" >> $GITHUB_ENV
```
_not tested_

.github/workflows/app.yml
nielsvanvelzen: We should follow the guide for gradle for the caching: https://docs.github.com/en/free-pro-team@latest/actions/guides/building-and-testing-java-with-gradle#caching-dependencies

.github/workflows/app.yml
```diff
@@ -0,0 +1,36 @@
+name: Test and build Jellyfin Android
+on:
+  push:
+  release:
+    types: [ published ]
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
nielsvanvelzen, line 8: I'd prefer to pin it to a version because Microsoft will definitely break the image like they always do. In addition, we should use a step to install JDK 11 or higher instead of the default JDK from the image.

https://docs.github.com/en/free-pro-team@latest/actions/guides/building-and-testing-java-with-gradle#specifying-the-jvm-version-and-architecture

nielsvanvelzen (changes requested):

.github/workflows/app.yml
h1dden-da3m0n: Alternatively you could set a output for this step, but it does not realy matter if you use step outputs or envs IMO

nielsvanvelzen: We need `JELLYFIN_VERSION` as environment variable in the Gradle build scripts so outputs would only make it more complicated because you'd basically do:
```
export JELLYFIN_VERSION = "${{ steps.variables.outputs.SomeVarName }}"
```
before running `gradlew`.

h1dden-da3m0n: :man_facepalming: I forgot about that requirement, sorry about that

nielsvanvelzen: We do have some slightly unusual requirements in the build so I won't blame you for forgetting that haha

.github/workflows/app.yml
Maxr1998: That should be possible, I just took the easiest approach here.

.github/workflows/app.yml
h1dden-da3m0n: May I ask if it where possible to pass the `JELLYFIN_VERSION` as a gradle build propertie, just so its use is more transperant in the CI?
e.g. `./gradlew test assemble versionTxt -Pjellyfin.version=...`

nielsvanvelzen: I think that might help understanding people how the CI works so I'd say let's try it

.github/workflows/app.yml
Maxr1998: Reworked the caching mechanism now.

.github/workflows/app.yml
Maxr1998: Done.

.github/workflows/app.yml
Maxr1998: I thought about doing this, but the problem is that you'd always set jellyfin.version then and would explicitly need to filter out e.g. empty strings. It makes the code more complicated as you need to specifically handle what the `if` in the env task did before, so I'd prefer to keep it as is.

dkanada:
@h1dden-da3m0n you're not off topic in a CI pull request haha. Also, feel free to migrate whatever pipelines you can find in *any* of our repositories. As long as they're 1:1 replacements we will gladly accept GitHub Actions.

h1dden-da3m0n:
> @h1dden-da3m0n you're not off topic in a CI pull request haha. Also, feel free to migrate whatever pipelines you can find in _any_ of our repositories. As long as they're 1:1 replacements we will gladly accept GitHub Actions.

Did not expect getting a reply on this :grinning:, anyway I currently helping out applying CI on all the plugin repos as well as other bits and pieces in those repos.
With that said, I will keep around for sure and help out where I can :wink:

.github/workflows/app.yml
```diff
@@ -0,0 +1,45 @@
...
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: ${{ runner.os }}-java-${{ matrix.java }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
+          restore-keys: |
+            ${{ runner.os }}-java-${{ matrix.java }}-gradle-
+            ${{ runner.os }}-java-
+      - name: Set environment variable with version
+        if: ${{ startsWith(github.ref, 'refs/tags/v') }}
+        run: echo "JELLYFIN_VERSION=${$GITHUB_REF | sed 's|refs/tags/||'}" >> $GITHUB_ENV
```
h1dden-da3m0n, line 32: are you sure this syntax is correct?
```suggestion
        run: echo "JELLYFIN_VERSION=$(echo $GITHUB_REF | sed 's|refs/tags/||')" >> $GITHUB_ENV
```

However, an alternative that requires escaping though:
```suggestion
        run: echo "JELLYFIN_VERSION=${GITHUB_REF//refs\/tags\//}" >> $GITHUB_ENV
```

Something useful i love to use when it comes to bash scripts: [Bash cheatsheet](https://devhints.io/bash#parameter-expansions)

.github/workflows/app.yml
```diff
@@ -0,0 +1,45 @@
...
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+      - name: Cache Gradle wrapper and dependencies
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: ${{ runner.os }}-java-${{ matrix.java }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
```
h1dden-da3m0n, line 26: is `${{ matrix.java }}` still set? It looks to me as if u don't have a matrix obj in this workflow and only build with JDK 11 anyway?

.github/workflows/app.yml
```diff
@@ -0,0 +1,45 @@
...
+          restore-keys: |
+            ${{ runner.os }}-java-${{ matrix.java }}-gradle-
```
h1dden-da3m0n, line 28: same here

h1dden-da3m0n: and sorry I should have done this as a review :facepalm:

.github/workflows/app.yml
Maxr1998: Now that you say it - no xD
I copied Niels' suggestion and didn't think about it. I like the alternative though. Gonna apply that one.

.github/workflows/app.yml
Maxr1998: Lazy copy-pasting on my end, and you expose me just like that. :eyes:
Thanks for the heads up, gonna fix!

.github/workflows/app.yml
Maxr1998: @nielsvanvelzen btw, could we also remove the 'v' prefix here instead of inside the `VersionUtils`?

.github/workflows/app.yml
Maxr1998: No worries :sweat_smile:

.github/workflows/app.yml
nielsvanvelzen: See my other comment

gradle.properties
```diff
@@ -9,7 +9,7 @@
 #
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
-org.gradle.jvmargs=-Xmx8192m
+org.gradle.jvmargs=-Xmx6144m -XX:MaxPermSize=512m -XX:MaxMetaspaceSize=512m
```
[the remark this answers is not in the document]
Maxr1998: I wanted to give it as much as possible (and originally set the 8 GB for only myself), but I reduced it because GitHub runners only have 7 GB of RAM. But you have a point, we shouldn't expect that all contributors have so much RAM - but I'd advocate for 4 GB instead.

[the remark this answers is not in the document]
Maxr1998: Oh, and the metaspacesize is from the official Gradle Docs, so I added it - I'd suggest adding that to the SDK as well.

.github/workflows/app.yml
```diff
@@ -0,0 +1,45 @@
+name: Test and build Jellyfin Android
```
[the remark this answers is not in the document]
Maxr1998: I prefer "Test app". but you're right.

gradle.properties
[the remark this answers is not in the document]
nielsvanvelzen: The SDK uses hundreds of Kotlin data classes and compiles fine with 2gb, I don't think 4gb is necessary. We can always increase is later if desired though

.github/workflows/app.yml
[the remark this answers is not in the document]
nielsvanvelzen: Should be `App Test` for ordering in the Actions tab. The format is "[Group] [Action]":

- Gradle update
- Gradle validate
- App Publish
- App Test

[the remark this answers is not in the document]
nielsvanvelzen: Should also rename the file to `app-test.yaml` with that.

.github/workflows/app.yml
```diff
@@ -0,0 +1,30 @@
...
+
+on:
+  push:
+    branches:
+      - master
+      - release-*
+  pull_request:
+
+jobs:
+  build:
```
nielsvanvelzen, line 11:
```suggestion
  test:
```

.github/workflows/app.yml
```diff
@@ -0,0 +1,30 @@
...
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-
+      - name: Run test and assemble tasks
+        run: ./gradlew --no-daemon --build-cache test assemble
```
nielsvanvelzen, line 30:
```suggestion
        run: ./gradlew --no-daemon --build-cache test
```
AFAIK assemble should be in the publish task and doesn't provide much extra on top of test except for packaging the .apk file. The test job will compile all code already.

Maxr1998: It does do minification and packaging steps, which might show errors in specific cases.. but I guess just using test should be enough.

h1dden-da3m0n (dismissed):

nielsvanvelzen (approved):

PR #267, opened 2020-12-12
Fix support for playlists in external player

### LINKED ISSUE #265: Media playback does not always respect External Player preference
**Describe the bug**
Some, but not all media will default to Web Player even though Client Settings is configured to play via External Player, in my case MX Player. However, if Client Settings is configured to play via ExoPlayer, it will work properly. Additionally, this behavior is not consistent as it happens only in certain scenarios:

1. Does not happen to all media. I can't quite pinpoint why it does/doesn't work as there is no discernible pattern. However, it is not random as if it works for Media A, it always does and if doesn't work for Media B, it never does.
2. Seems to only affect TV Series but not Movies
3. Only certain conditions will trigger said behavior - A) the Play button on the bottom right of the Series thumbnail in Home page B) Play button in the Series main page. Playing the individual episode from the episode list works fine.
4. It **always** works properly when the affected media is resumed from Next Up or Continue Watching list

**To Reproduce**

I cannot determine why only certain media is affected, but if you do have affected media:
1. Configure Client Settings to play using External Player. I'm using MX Player
2. Go to Home page and tap the Play icon on series thumbnail in Latest TV Shows
3. Media will be played using Web Player, ignoring the Client Settings
4. Play media long enough so that it will be in a resumable state in Continue Watching
5. Tap on Play icon on series thumbnail again, but this time from Continue Watching instead of Latest TV Shows
6. Media will be played using MX Player, according to Client Settings

**Expected behavior**
Media should be played according to Client Settings preferences at all times

**Logs**
No error, media still plays fine

 - App version: 2.2.0
 - Android version: Android 10
 - Device manufacturer and model: Pocophone F1
 - Jellyfin server version: 10.6.4
 - The type of video player implementation used: External Player (MX Player)

CarlosOlivo:
Only allow intros, a single item to play and stop as there is no way to stop a playlist from an external player.

Closes #265

### DIFF
--- a/app/src/main/assets/native-10.7/ExternalPlayerPlugin.js
+++ b/app/src/main/assets/native-10.7/ExternalPlayerPlugin.js
@@ -24,6 +24,7 @@ export class ExternalPlayerPlugin {
         this._paused = true;
         this._volume = 100;
         this._currentSrc = null;
+        this._isIntro = false;
 
         this._externalPlayer = window['ExternalPlayer'];
     }
@@ -33,8 +34,7 @@ export class ExternalPlayerPlugin {
     }
 
     canPlayItem(item, playOptions) {
-        var mediaSource = item.MediaSources && item.MediaSources[0];
-        return this._externalPlayer.isEnabled() && mediaSource && mediaSource.SupportsDirectStream;
+        return this._externalPlayer.isEnabled();
     }
 
     supportsPlayMethod(playMethod, item) {
@@ -49,6 +49,7 @@ export class ExternalPlayerPlugin {
         this._currentTime = (options.playerStartPositionTicks || 0) / 10000;
         this._paused = false;
         this._currentSrc = options.url;
+        this._isIntro = options.item && options.item.ProviderIds && options.item.ProviderIds.hasOwnProperty("prerolls.video");
         this._externalPlayer.initPlayer(JSON.stringify(options));
     }
 
@@ -97,6 +98,7 @@ export class ExternalPlayerPlugin {
             src: this._currentSrc
         };
 
+        this.playbackManager._playNextAfterEnded = this._isIntro;
         this.events.trigger(this, 'stopped', [stopInfo]);
         this._currentSrc = this._currentTime = null;
     }
--- a/app/src/main/assets/native-10.6/externalplayer.js
+++ b/app/src/main/assets/native-10.6/externalplayer.js
@@ -23,14 +23,14 @@ define(['appSettings', 'events', 'playbackManager'], function (appSettings, even
         self._paused = true;
         self._volume = 100;
         self._currentSrc = null;
+        self._isIntro = false;
 
         self.canPlayMediaType = function (mediaType) {
             return mediaType === 'Video';
         };
 
         self.canPlayItem = function (item, playOptions) {
-            var mediaSource = item.MediaSources && item.MediaSources[0];
-            return window.ExternalPlayer.isEnabled() && mediaSource && mediaSource.SupportsDirectStream;
+            return window.ExternalPlayer.isEnabled();
         };
 
         self.supportsPlayMethod = function (playMethod, item) {
@@ -46,6 +46,7 @@ define(['appSettings', 'events', 'playbackManager'], function (appSettings, even
                 self._currentTime = (options.playerStartPositionTicks || 0) / 10000;
                 self._paused = false;
                 self._currentSrc = options.url;
+                self._isIntro = options.item && options.item.ProviderIds && options.item.ProviderIds.hasOwnProperty("prerolls.video");
                 window.ExternalPlayer.initPlayer(JSON.stringify(options));
                 resolve();
             });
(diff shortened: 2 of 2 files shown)


nielsvanvelzen (approved):

Maxr1998 (approved):
LGTM, but have you checked how this works with something like the intros plugin?

dkanada:
@CarlosOlivo if you do end up testing the intros plugin, please let me know if the settings can be changed from the defaults on the latest version. I'm getting reports about that, but I've checked on my server and it seems to be working.

CarlosOlivo:
> LGTM, but have you checked how this works with something like the intros plugin?

Well... fixed! (?)

> @CarlosOlivo if you do end up testing the intros plugin, please let me know if the settings can be changed from the defaults on the latest version. I'm getting reports about that, but I've checked on my server and it seems to be working.

I do not know about the previous versions since it is the first time I have tried the intros plugin but at least in version 1.2.1.0 it allows me to change the default settings.

nielsvanvelzen:
Won't this fix cause the same issue to happen when pressing back during an intro?

CarlosOlivo:
> Won't this fix cause the same issue to happen when pressing back during an intro?

As long as at least a millisecond spends playing the intro, the intro is marked as seen and plays the next item.

nielsvanvelzen:
From my testing this does not work with intros. It will play the intro and after that nothing happens, no movie is showing up for me.

CarlosOlivo:
> From my testing this does not work with intros. It will play the intro and after that nothing happens, no movie is showing up for me.
> 
> Tested with 10.7 between RC1/RC2 in

Oh sorry, I forgot to add a `this` in 10.7, try again.

---

Also RIP Azure pipeline

```
> Could not resolve all artifacts for configuration ':buildSrc:classpath'.
   > Could not resolve org.gradle.kotlin:plugins:1.3.6.
     Required by:
         project :buildSrc > org.gradle.kotlin.kotlin-dsl:org.gradle.kotlin.kotlin-dsl.gradle.plugin:1.3.6
      > Could not resolve org.gradle.kotlin:plugins:1.3.6.
         > Could not get resource 'https://plugins.gradle.org/m2/org/gradle/kotlin/plugins/1.3.6/plugins-1.3.6.module'.
            > Could not GET 'https://jcenter.bintray.com/org/gradle/kotlin/plugins/1.3.6/plugins-1.3.6.module'. Received status code 502 from server: Bad Gateway
```

nielsvanvelzen:
Pressing the back button during the intro will still open the movie so you'd need to press back again. I guess that's something we'll have to live with until playlist management is implemented natively. @Maxr1998 what do you think?

Maxr1998:
Guess that'd be acceptable, yes. I haven't really looked into this PR yet, the only concern I currently have is that it might add a lot of complexity to the migration process when developing the native app.

app/src/main/assets/native-10.7/ExternalPlayerPlugin.js
```diff
@@ -49,6 +49,9 @@ export class ExternalPlayerPlugin {
         this._currentTime = (options.playerStartPositionTicks || 0) / 10000;
         this._paused = false;
         this._currentSrc = options.url;
+        this._isIntro = typeof options.item != "undefined" &&
+            typeof options.item.ProviderIds != "undefined" &&
+            options.item.ProviderIds.hasOwnProperty("prerolls.video");
```
Maxr1998, line 54: No JS wizard, but couldn't this be simplified a bit? Does it really need the `typeof` here?

CarlosOlivo: This is the "safe" approach

I found something (Kotlin?), although it is relatively new, and some old webview may not be compatible.

https://github.com/tc39/proposal-optional-chaining
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

`options?.item?.ProviderIds?.["prerolls.video"] != undefined`

More options, choose one if you like: https://stackoverflow.com/questions/2631001/test-for-existence-of-nested-javascript-object-key

Or ask for your nearest JS Wizard (?)

nielsvanvelzen: Ex-JS Wizard here 🧙 

I think we could use JavaScript quirks to simplify to:

```suggestion
        this._isIntro = options.item && options.item.ProviderIds && options.item.ProviderIds.hasOwnProperty('prerolls.video');
```

Basically we check if the members are true-ish, `undefined` or `null` is falsy so this check makes sure  the members exists before calling the hasOwnProperty function.

CarlosOlivo: Ah yes, the magic world of JS, that worked before 

https://github.com/jellyfin/jellyfin-android/blob/6cc40f6571f554859446ad35cb4e58138a0d492d/app/src/main/assets/native-10.7/ExternalPlayerPlugin.js#L36-L37

Anything else?

Maxr1998 (approved):

app/src/main/assets/native-10.7/ExternalPlayerPlugin.js
Maxr1998: Nope, should be fine now.

PR #275, opened 2020-12-18
Add support for AV1

Maxr1998:
(Only available on Android 10+)


Maxr1998:
Hilariously, it straight up didn't work in my first test because my Pixel 5 seemingly maxes out at 1080p for AV1 support. We should be able to filter that out in the future through better `android.media.MediaCodecInfo.CodecCapatibilities` matching.

nielsvanvelzen (approved):

PR #286, opened 2021-01-17
Improve fullscreen, orientation and PiP handling

### LINKED ISSUE #174: Navigation bar does not hide itself after initial playback
**Describe the bug**

When exiting out of the app and going back into it, or turning off display and unlocking the device, the navigation bar that is hidden when initial playback starts never gets hidden again.

**To Reproduce**
1. Start playback
2. Exit app
3. Return to app

**Expected behavior**
For the navbar to hide itself again.

 - App version: 2.0.0
 - Android version: Android 10
 - Device manufacturer and model: Google Pixel 2
 - Jellyfin server version: not relevent
 - The type of video player implementation used: Happens on ExoPlayer and web players impls

ThreeFive-O:
Fixes #174.

See: https://developer.android.com/training/system-ui/navigation
> When the user reopens the activity, onCreate() won't get called, so the system bars will remain visible. If you want system UI changes to persist as the user navigates in and out of your activity, set UI flags in onResume() or onWindowFocusChanged().

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/player/PlayerFragment.kt
+++ b/app/src/main/java/org/jellyfin/mobile/player/PlayerFragment.kt
@@ -4,7 +4,6 @@ import android.annotation.SuppressLint
 import android.app.PictureInPictureParams
 import android.content.pm.ActivityInfo
 import android.content.res.Configuration
-import android.graphics.Color
 import android.media.AudioManager
 import android.os.Build
 import android.os.Bundle
@@ -13,7 +12,6 @@ import android.view.*
 import android.view.WindowManager.LayoutParams.*
 import android.widget.*
 import androidx.core.content.getSystemService
-import androidx.core.content.withStyledAttributes
 import androidx.core.view.ViewCompat
 import androidx.core.view.isVisible
 import androidx.core.view.postDelayed
@@ -100,11 +98,8 @@ class PlayerFragment : Fragment() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
-        // Set orientation to landscape and enable fullscreen initially, set status bar color
-        with(requireActivity()) {
-            requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_SENSOR_LANDSCAPE
-            enableFullscreen()
-        }
+        // Request landscape orientation for playback start
+        requireActivity().requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_SENSOR_LANDSCAPE
 
         // Observe ViewModel
         viewModel.player.observe(this) { player ->
@@ -180,6 +175,15 @@ class PlayerFragment : Fragment() {
         orientationListener.enable()
     }
 
+    override fun onResume() {
+        super.onResume()
+
+        // When returning from another app, the last active mode has to be set again
+        with(requireActivity()) {
+            if (isLandscape()) enableFullscreen() else disableFullscreen()
+        }
+    }
+
     fun lockScreen() {
         playerView.useController = false
         orientationListener.disable()
@@ -192,7 +196,7 @@ class PlayerFragment : Fragment() {
             requireActivity().requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_UNSPECIFIED
         }
         orientationListener.enable()
-        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.O || (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O && !requireActivity().isInPictureInPictureMode)) {
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.N || !requireActivity().isInPictureInPictureMode) {
             playerView.useController = true
             playerView.apply {
                 if (!isControllerVisible) showController()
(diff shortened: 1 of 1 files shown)


Maxr1998 (commented):
Good catch, thanks for the fix. Could you move the `onResume` function below `onStart` please? That order makes more sense imo considering the activity lifecycle.

ThreeFive-O:
@Maxr1998 Thank you for the feedback. Fixed the feedback and the code comment which was outdated.

One thing I noticed is that onResume doesn't fully fix the case when returning (maximizing) from Picture in Picture (PiP) mode. In most cases the navigation bar is still visible, but can be hidden by tapping the player window. In rare cases the navigation bar hides when returning from PiP.
Returning from PiP calls onResume, but there must be some other problem involved.
I'm on Android Pie on a Samsung phone.

Any idea?

Maxr1998:
I've observed that as well when first implementing PiP, but never found a solution/fix for the issue. I'm thinking it's a bug in the Android framework or ExoPlayer library, but I can't really tell.

YouTube opens the video in portrait mode when returning from PiP, so the bug doesn't apply there. Maybe we should do the same to cover it, it also makes sense since you mostly watch through PiP while in portrait.

ThreeFive-O:
@Maxr1998 It took me a while to figure out the fullscreen logic correctly, but I guess with the latest commit I addressed all issues.

The behavior of the latest commit is as follows:

- When starting the black we request landscape orientation. The configuration change handler will take care about fullscreen mode.
- When playback is stopped and app is paused (because of an app switch) the `onResume` handler will restore the latest viewing mode (fullscreen when it was fullscreen; normal mode when it was normal mode)
- When switching to PiP mode the configuration changes for enabling/disabling fullscreen mode are not handled. This fixes the rogue orientation change in the configuration change handle.

Please give it a try.

app/src/main/java/org/jellyfin/mobile/player/PlayerFragment.kt
```diff
@@ -180,6 +177,18 @@ class PlayerFragment : Fragment() {
...
 
+    override fun onResume() {
+        super.onResume()
+
+        // When returning from another app, the last active mode has to be set again
+        with(requireActivity()) {
+            if(isLandscape(resources.configuration))
+                enableFullscreen()
+            else
+                disableFullscreen()
```
Maxr1998, line 188:
```suggestion
            if (isLandscape()) enableFullscreen() else disableFullscreen()
```

app/src/main/java/org/jellyfin/mobile/player/PlayerFragment.kt
```diff
@@ -204,6 +213,9 @@ class PlayerFragment : Fragment() {
      * Handle current orientation and update fullscreen state and switcher icon
      */
     private fun updateFullscreenStateAndSwitcher(configuration: Configuration) {
+        // Do not handle any orientation changes while being in Picture-in-Picture mode
+        if(requireActivity().isInPictureInPictureMode) return
```
Maxr1998, line 217:
```suggestion
        if (requireActivity().isInPictureInPictureMode)
            return
```
You could also combine all the calls to `requireActivity()` in this method to use a `with()` block.

Maxr1998 (changes requested):
Thanks for your other set of changes, looks good to me! Just a small formatting nitpicks to fix, then it's ready to merge!

app/src/main/java/org/jellyfin/mobile/player/PlayerFragment.kt
ThreeFive-O: Wrapped and done!

app/src/main/java/org/jellyfin/mobile/player/PlayerFragment.kt
```diff
@@ -204,12 +210,18 @@ class PlayerFragment : Fragment() {
...
+        with(requireActivity()) {
+            // Do not handle any orientation changes while being in Picture-in-Picture mode
+            if (isInPictureInPictureMode)
+                return
+
+            if (isLandscape(configuration)) enableFullscreen() else disableFullscreen()
+            val fullscreenDrawable = when {
+                !isFullscreen() -> R.drawable.ic_fullscreen_enter_white_32dp
+                else -> R.drawable.ic_fullscreen_exit_white_32dp
+            }
```
nielsvanvelzen, line 222:
```suggestion
            val fullscreenDrawable = if (isFullscreen())
                R.drawable.ic_fullscreen_exit_white_32dp
            else
                R.drawable.ic_fullscreen_enter_white_32dp
```

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/mobile/player/PlayerFragment.kt
Maxr1998: I'd personally keep it as is, even though the when isn't necessary here, it looks cleaner and makes it easier to see the two used drawable resources.

nielsvanvelzen: In that case I'd invert the condition, instead of checking for `!isFullscreen()` use `isFullscreen()`. More readable that way.

nielsvanvelzen (approved):

ThreeFive-O:
Thank you @Maxr1998 and @nielsvanvelzen for your code reviews and improvements.

Maxr1998:
Added another small fix (the `isPictureInPictureMode` call could have crashed below N) and rebased/squashen all commits onto the lastest master.

Maxr1998 (approved):

nielsvanvelzen (approved):

nielsvanvelzen (approved):

PR #287, opened 2021-01-18
Miscellaneous casting fixes

nielsvanvelzen:
- Added a few logging statements
- Made the execCast and castCallback async so the JS engine will behave slightly different in the way everything is called
  - This fixes the unstable client never loading
- Change how the chrome cast plugin is loaded to MITM the google url instead
  - Because jellyfin-web does [some magic with that url](https://github.com/jellyfin/jellyfin-web/blob/eb1e501fad14b15ef3ea7d102c8caff46a4963ea/src/components/castSenderApi.js)

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
+++ b/app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
@@ -130,10 +130,6 @@ class WebViewFragment : Fragment() {
                 val url = request.url
                 val path = url.path?.toLowerCase(Locale.ROOT) ?: return null
                 return when {
-                    path.endsWith(Constants.CAST_SDK_PATH) -> {
-                        // Load the chrome.cast.js library instead
-                        webView.context.loadAsset("native-${assetsVersion}/chrome.cast.js")
-                    }
                     path.endsWith(Constants.APPLOADER_PATH) || path.endsWith(Constants.ANY_BUNDLE_PATH) -> {
                         assetsVersion = when {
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
```diff
@@ -130,6 +130,10 @@ class WebViewFragment : Fragment() {
                 val url = request.url
                 val path = url.path?.toLowerCase(Locale.ROOT) ?: return null
                 return when {
+                    path.endsWith(Constants.CAST_SDK_PATH) -> {
+                        // Load the chrome.cast.js library instead
+                        webView.context.loadAsset("native-${assetsVersion}/chrome.cast.js")
+                    }
```
Maxr1998, line 136: Could you put this below the `path.contains("native")` line?

app/src/main/java/org/jellyfin/mobile/utils/Constants.kt
```diff
@@ -13,6 +13,7 @@ object Constants {
     const val ANY_BUNDLE_PATH = ".bundle.js" // 10.7 >=
     const val SELECT_SERVER_PATH = "selectserver.html"
     const val SESSION_CAPABILITIES_PATH = "sessions/capabilities/full"
+    const val CAST_SDK_PATH = "cast_sender.js"
```
Maxr1998, line 16: This one would be below the `ANY_BUNDLE_PATH` then.

Maxr1998 (approved):
LGTM, however I would reorder the statements a bit.

Maxr1998 (approved):

PR #297, opened 2021-02-07
Native player improvements

Maxr1998:
Various improvements for the native player, including better handling of full-screen for portrait mode videos, correct aspect ratio for PiP, smarter subtitle handling and some bugfixes.


app/src/main/java/org/jellyfin/mobile/player/source/ExoPlayerTrack.kt
```diff
@@ -9,6 +10,10 @@ sealed class ExoPlayerTrack(track: JSONObject) {
     val title: String = track.optString("DisplayTitle", "")
 
     class Video(track: JSONObject) : ExoPlayerTrack(track) {
+        val width: Int = track.optInt("Width", -1)
```
nielsvanvelzen, line 13: We should move those strings to constants

app/src/main/java/org/jellyfin/mobile/player/PlaybackMenus.kt
```diff
@@ -128,6 +149,12 @@ class PlaybackMenus(
         }
     }
 
+    private fun updateSubtitlesButton() {
+        subtitlesButton.isEnabled = subtitleCount > 0
```
nielsvanvelzen, line 153: I think we want to hide the button when there's no subtitles instead of disabling it. Other apps do the same thing.

Maxr1998: That's a good point, yeah. Gonna do that.

app/src/main/java/org/jellyfin/mobile/player/source/ExoPlayerTrack.kt
Maxr1998: I don't know, we'll rip this out anyways one we get to the stage where we handle video playback with the new apiclient..

nielsvanvelzen (approved):

PR #300, opened 2021-02-10
Drop AndroidX Webkit dependency

CarlosOlivo:
Related to #96

Apparently the AndroidX WebKit dependency has compatibility problems with certain old versions of WebView and the jellyfin-web combo, although I can't find the origin/source/reason in the Logcat, at least the WebView versions >= 74 don't throw this exception.

I also notice a slight performance improvement when using just the WebViewClient, maybe it's my imagination.

Also, it will eventually be removed when the switch is made to the native client, who agrees?

- AVD settings
Android 7.1.1 (API 25) x86_64
WebView com.android.chrome version 69.0.3497.100 (code 349710067)

- With androidx.webkit.WebViewClientCompat (Version 1.4.0)
xCqdeavBLS
```
2021-02-09 11:49:38.500 10152-10152/org.jellyfin.mobile.debug E/RedScreenOfDeath: ══════════ Exception caught by Red Screen Of Death library ═════════
2021-02-09 11:49:38.500 10152-10152/org.jellyfin.mobile.debug E/RedScreenOfDeath: IllegalArgumentException
    java.lang.IllegalArgumentException: reasonPhrase can't be empty.
        at android.webkit.WebResourceResponse.setStatusCodeAndReasonPhrase(WebResourceResponse.java:138)
        at android.webkit.WebResourceResponse.<init>(WebResourceResponse.java:76)
        at xl.a(PG:292)
        at afG.handleMessage(PG:67)
        at android.os.Handler.dispatchMessage(Handler.java:102)
        at android.os.Looper.loop(Looper.java:154)
        at android.app.ActivityThread.main(ActivityThread.java:6119)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:886)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:776)
2021-02-09 11:49:38.500 10152-10152/org.jellyfin.mobile.debug E/RedScreenOfDeath: ════════════════════════════════════════════════════════════════════
```

- With android.webkit.WebViewClient
WdXmB7x0jc
```
2021-02-10 00:15:21.085 7609-7609/org.jellyfin.mobile.debug E/WebViewFragment$initialize: Received WebView HTTP 400 error: null
```

Also closes #212

PR #326, opened 2021-03-23
Check WebView support and version

Maxr1998:
Closes #300.

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/utils/Constants.kt
+++ b/app/src/main/java/org/jellyfin/mobile/utils/Constants.kt
@@ -9,6 +9,7 @@ object Constants {
     const val APP_INFO_VERSION: String = BuildConfig.VERSION_NAME
 
     // Webapp constants
+    const val MINIMUM_WEB_VIEW_VERSION = 88
     const val WEB_CONFIG_PATH = "config.json"
     const val CAST_SDK_PATH = "cast_sender.js"
     const val SESSION_CAPABILITIES_PATH = "sessions/capabilities/full"
--- a/app/src/main/java/org/jellyfin/mobile/utils/WebViewUtils.kt
+++ b/app/src/main/java/org/jellyfin/mobile/utils/WebViewUtils.kt
@@ -1,0 +1,68 @@
+/**
+ * Taken and adapted from https://github.com/tachiyomiorg/tachiyomi/blob/master/app/src/main/java/eu/kanade/tachiyomi/util/system/WebViewUtil.kt
+ *
+ * Copyright 2015 Javier Tomás
+ *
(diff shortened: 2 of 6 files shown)


app/src/main/java/org/jellyfin/mobile/utils/WebViewUtils.kt
```diff
@@ -0,0 +1,69 @@
...
+
+    // Revert to original UA string
+    settings.userAgentString = originalUA
+
+    return defaultUserAgentString
+}
+
+@SuppressLint("SetJavaScriptEnabled")
+fun WebView.setDefaultSettings() {
+    with(settings) {
```
nielsvanvelzen, line 65:
```suggestion
fun WebView.setDefaultSettings() = with(settings) {
```

app/src/main/java/org/jellyfin/mobile/utils/Constants.kt
```diff
@@ -9,6 +9,7 @@ object Constants {
     const val APP_INFO_VERSION: String = BuildConfig.VERSION_NAME
 
     // Webapp constants
+    const val MINIMUM_WEB_VIEW_VERSION = 88
```
nielsvanvelzen, line 12: Any particular reason version 88 was chosen?

app/src/main/res/values/strings.xml
```diff
@@ -1,4 +1,11 @@
 <resources xmlns:tools="http://schemas.android.com/tools" tools:ignore="MissingTranslation">
+    <string name="dialog_web_view_not_supported">WebView not supported</string>
+    <string name="dialog_web_view_not_supported_message">WebViews are not supported by this device.\nPlease install a compatible WebView implementation if possible or switch to a supported device.</string>
```
nielsvanvelzen, line 3:
```suggestion
    <string name="dialog_web_view_not_supported_message">WebViews are not supported by this device.\nInstall a compatible WebView implementation or switch to a supported device.</string>
```

No need to beg the user

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/mobile/utils/Constants.kt
Maxr1998: Tachiyomi uses the same, so it seemed sane.

app/src/main/res/values/strings.xml
Maxr1998: lol, kk

nielsvanvelzen (approved):

nielsvanvelzen (dismissed):

nielsvanvelzen (approved):

PR #330, opened 2021-03-28
ci(publish): add initial publish app GH workflow

h1dden-da3m0n:
### Description

This is an initial attempt on offering a replacement for the sol remaining Azure CI workflow.
Any feedback is welcome and the are still at least 2 steps missing from the original workflow (namely: `CopyFilesOverSSH`, `SSH`, mostly since I don't know what they are exactly used for or how to replicate them).

### Changes

* Add publish GitHub workflow 

### Issues

* n/a (move from Azure CI to GH Actions)


.github/workflows/publish.yaml
```diff
@@ -0,0 +1,68 @@
...
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-
+
+      - name: Set variable for version
+        id: variable
+        run: echo "::set-output name=AppVersion::${GITHUB_REF//refs\/tags\//}"
```
nielsvanvelzen, line 31: Should just set the env var so all gradle tasks pick it up automatically, like so:

https://github.com/jellyfin/jellyfin-sdk-kotlin/blob/5d1d7e3600d8f9d2f98dd5dbc2a68f4090f4f416/.github/workflows/sdk-publish.yaml#L22-L25

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,68 @@
+name: Publish App
```
nielsvanvelzen, line 1:
```suggestion
name: App Publish
```

Also, file name should be "app-publish.yaml"

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+on:
+  push:
+    tags: [ 'v*' ]
```
nielsvanvelzen, line 5:
```suggestion
    tags:
      - v*
```

nielsvanvelzen (changes requested):
(Only reviewed until the signing part)

Ideally the publishing workflow also publishes test versions (on the master branch) that are not signed. The Kotlin SDK works similarly. See https://github.com/jellyfin/jellyfin-sdk-kotlin/blob/5d1d7e3600d8f9d2f98dd5dbc2a68f4090f4f416/.github/workflows/sdk-publish.yaml

.github/workflows/publish.yaml
h1dden-da3m0n: Thy for the reference and given the different application than I assumed I agree that the env var will work better.
I used this method originally to make the use of the version more transparent (like we discussed in the test PR ages ago :sweat_smile:).

h1dden-da3m0n: Allow me to ask this though: Why do you need the `tr / -` pipe in the sdk repo?

nielsvanvelzen: The `tr` part is not necessary but I kept it just to be safe IIRC

h1dden-da3m0n: mind if I omit it? additionally should I too remove the `v` prefix or let the VersionUtils.kt handle it?

nielsvanvelzen: I'd rather keep it in case there was a reason that I forgot :p

Also the v prefix can be used to detect version tags in the CI but it should **always** pass the exact tag to the Gradle build scripts. The Gradle build scripts will handle it.

h1dden-da3m0n: Okay then I'll use your version of the export 1:1. Regarding the tag I agree, however was unsure since the old script handled only the `refs/tags` bit so wanted to be sure :+1:

h1dden-da3m0n:
okay, I am not 100% on the repo upload script NGL. Please let me know what you think of that, otherwise this *should* cover the requirements to this workflow.
There are a few secrets that would be required to be added to the repo should the workflow in its current form be accepted, but I expect this to still go throug an iteration or two.

Note I plan to squash the workflow commits in the end unless you want the full history, but this should be discussed at the end.

h1dden-da3m0n:
okay I opted to remove my shell script in favor of two actions, I found them to be more straight forward (however I can always revert if you like the script more)
There are several secrets that this workflow would require to be added to this repo before it would work and I still need to squash the workflow commits, but functionality wise it should be equivalent to the Azure CI pipeline.

.github/workflows/app-publish.yaml
```diff
@@ -0,0 +1,121 @@
+name: App Publish
+
+on:
+  push:
+    tags:
+      - v*
+    branches:
+      - master
+      - release*
```
nielsvanvelzen, line 9:
```suggestion
      - master
```

No need to run on release branches because we (normally) only push when we're going to tag it.

.github/workflows/app-publish.yaml
```diff
@@ -0,0 +1,121 @@
...
+
+      - name: Setup Gradle cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-
```
nielsvanvelzen, line 32: Don't use caching for the publish task, always build clean so the cache can't cause or suppress any issues.

.github/workflows/app-publish.yaml
```diff
@@ -0,0 +1,121 @@
...
+
+      - name: Set JELLYFIN_VERSION
+        if: ${{ startsWith(github.ref, 'refs/tags/v') }}
+        run: echo "JELLYFIN_VERSION=$(echo ${GITHUB_REF#refs/tags/v} | tr / -)" >> $GITHUB_ENV
+
+      - name: Build Release
+        run: ./gradlew --no-daemon --build-cache assembleRelease versionTxt
```
nielsvanvelzen, line 39:
```suggestion
      - name: Build Release
        run: ./gradlew --no-daemon --info assembleRelease versionTxt
```

.github/workflows/app-publish.yaml
```diff
@@ -0,0 +1,121 @@
...
+
+      - name: Split APK release types
+        run: |
+          mkdir app/build/outputs/apk/libre_release app/build/outputs/apk/proprietary_release;
```
nielsvanvelzen, line 43:
```suggestion
          mkdir build/jellyfin-publish/release-libre build/jellyfin-publish/release-proprietary;
```

I'd rather use a (custom) build directory in the root

nielsvanvelzen: The version.txt should be moved to `build/jellyfin-publish/`, and that directory will be the artifact path

.github/workflows/app-publish.yaml
```diff
@@ -0,0 +1,121 @@
...
+          retention-days: 30
+          if-no-files-found: error
+          path: |
+            app/build/outputs/apk/libre_release/*
+            app/build/outputs/apk/proprietary_release/*
+            app/build/version.txt
+
+  publish-release:
+    runs-on: ubuntu-20.04
+    needs: build-publish
```
nielsvanvelzen, line 59:
```suggestion
    needs: build
```
The job is named `build`

.github/workflows/app-publish.yaml
```diff
@@ -0,0 +1,121 @@
...
+
+      - name: Sign Proprietary Release APK
+        uses: r0adkll/sign-android-release@v1
+        id: sign_proprietary
+        with:
+          releaseDirectory: app/build/outputs/apk/proprietary_release
+          signingKeyBase64: ${{ secrets.SIGNING_KEY }}
+          alias: ${{ secrets.ALIAS }}
+          keyStorePassword: ${{ secrets.KEY_STORE_PASSWORD }}
+          keyPassword: ${{ secrets.KEY_PASSWORD }}
```
nielsvanvelzen, line 88: We might want to move the signing part to Gradle, We'd need to apply the `signingConfigs` in the `android {}` block for that though.

.github/workflows/app-publish.yaml
```diff
@@ -0,0 +1,121 @@
...
+
+      - name: Upload Release Artifacts
+        uses: ncipollo/release-action@v1.8.2
```
nielsvanvelzen, line 91: Can you change it to use `alexellis/upload-assets@0.2.2`? Helps with consistency between the various Android/Kotlin projects (Kotlin SDK uses that one)

.github/workflows/app-publish.yaml
```diff
@@ -0,0 +1,121 @@
...
+          if-no-files-found: error
+          path: |
+            app/build/outputs/apk/libre_release/*
+            app/build/outputs/apk/proprietary_release/*
+            app/build/version.txt
+
+  publish-release:
+    runs-on: ubuntu-20.04
+    needs: build-publish
+    if: ${{ startsWith(github.ref, 'refs/tags/v') }}
```
nielsvanvelzen, line 60:
```suggestion
    environment: release
    if: ${{ startsWith(github.ref, 'refs/tags/v') }}
```

The secrets will be added to the release environment

nielsvanvelzen (changes requested):

.github/workflows/app-publish.yaml
h1dden-da3m0n: NGL, with all the various versions of Actions you can find to do one thing this srsly gets rather convoluted and difficult to track which version is 1. maintained and 2. security patched.
The reason why I went with this one was due to its wider feature set (and since the official GH `actions/upload-release-asset` was deprecated :disappointed: )

However, I agree if you use the other action in the Kotlin SDK repo I will switch to that one :wink:

.github/workflows/app-publish.yaml
h1dden-da3m0n: :thinking: I understand what you mean, however have to disclose I would have no idea on how to do that. My experience with Gradle so far is rather limited, but I will see what I can do :+1:

nielsvanvelzen: No problem, we can always change this part later

.github/workflows/app-publish.yaml
h1dden-da3m0n: Sounds reasonable, I have not thought about it this way NGL

.github/workflows/app-publish.yaml
h1dden-da3m0n: Okay I read up on this a bit and have a couple of questions:

* given the logic of this, I assume the jks for the gradle based solution still needs to be base64-ed and serialized back to a jks before the build?
* given what I have found the signing would happen as part of the release build, if the jks, its passwd and the alias where present and unsigned if they are missing. However, I assume you thought of a custom gradle signing task that does nothing but sign the output?

Maxr1998: That's correct, you'd deserialize the keystore, put it to the right location, set the alias and passwords from the secrets, and the signing is done by the build tools. I did exactly that a while ago, see [here](https://github.com/Maxr1998/MaxLock/blob/master/buildSrc/src/main/kotlin/SigningConfigHelper.kt#L24) for reference.

h1dden-da3m0n: thank you for that reference, I shall give it a try to integrate it here.

h1dden-da3m0n: Okay, 412a4d58414d5f70f58206b9bcd679f1964a0416 now has a very 'scripted' variation of the sign config contained in the build.gradle.kts. However, I guess you imagined something more elegant than that, so please let me know.

h1dden-da3m0n: Okay, ended up installing Android Studio because IDEA refused to import properly and added a very 'scripted' signing config in a95b2e4c75adf799615286cfec5941dbacb563e4. I assume you where thinking of a more elegant solution, so let me know if I should change something or just remove the commit an squash the rest of the workflow commits.

.github/workflows/app-build.yml
```diff
@@ -0,0 +1,42 @@
...
+        run: |
+          mkdir -p build/jellyfin-publish/release-libre build/jellyfin-publish/release-proprietary;
+          mv app/build/outputs/apk/*/*/jellyfin-android-*-libre-release-unsigned.apk build/jellyfin-publish/release-libre/;
+          mv app/build/outputs/apk/*/*/jellyfin-android-*-proprietary-release-unsigned.apk build/jellyfin-publish/release-proprietary/;
+          mv app/build/version.txt build/jellyfin-publish/;
+
+      - uses: actions/upload-artifact@v2
+        with:
+          name: build-publish-artifact
+          retention-days: 30
```
nielsvanvelzen, line 40:
```suggestion
          name: build-artifacts
          retention-days: 14
```

.github/workflows/app-publish.yaml
```diff
@@ -0,0 +1,73 @@
...
+
+      - name: Set JELLYFIN_VERSION
+        run: echo "JELLYFIN_VERSION=$(echo ${GITHUB_REF#refs/tags/v} | tr / -)" >> $GITHUB_ENV
+
+      - name: Build Signed Release
+        env:
+          KEYSTORE: ${{ secrets.KEYSTORE }}
+          KEYSTORE_PASSWORD: ${{ secrets.KEYSTORE_PASSWORD }}
+          KEY_ALIAS: ${{ secrets.KEY_ALIAS }}
+          KEY_PASSWORD: ${{ secrets.KEY_PASSWORD }}
```
nielsvanvelzen, line 31: - [ ] These need to be added (preferably on organization level so we can re-use it in the ATV app later).

_(mark when added)_

.github/workflows/app-publish.yaml
```diff
@@ -0,0 +1,73 @@
...
+
+      - name: Upload to repo.jellyfin.org
+        uses: burnett01/rsync-deployments@4.1
+        with:
+          switches: -rltgoDzvO --delete --exclude='*' --include='**/*.apk' --include='*.txt'
+          path: build/jellyfin-publish/
+          remote_path: /srv/repository/releases/client/android/versions/v${{ env.JELLYFIN_VERSION }}
+          remote_host: ${{ secrets.DEPLOY_HOST }}
+          remote_user: ${{ secrets.DEPLOY_USER }}
+          remote_key: ${{ secrets.DEPLOY_KEY }}
```
nielsvanvelzen, line 59: - [ ] These need to be added (preferably on organization level so we can re-use it in the ATV app later).

_(mark when added)_

nielsvanvelzen (changes requested):

.github/workflows/app-build.yml
```diff
@@ -0,0 +1,42 @@
+name: App Build
+
+on:
+  push:
+    branches:
+      - master
```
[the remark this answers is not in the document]
h1dden-da3m0n: I mean the test workflow already runs on PRs IIRC,  however I see the appeal of having a per-compiled artefact to test b4 merging a pr

.github/workflows/app-build.yml
```diff
@@ -0,0 +1,42 @@
...
+      - name: Setup Java
+        uses: actions/setup-java@v2
+        with:
+          distribution: 'adopt'
+          java-version: 11
+
+      - name: Set dev JELLYFIN_VERISON
+        run: |
+          VERSION=$(curl --silent "https://api.github.com/repos/${{ github.repository }}/releases/latest" | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/')
+          echo "JELLYFIN_VERSION=$(echo ${VERSION#v} | tr / -)-dev" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
h1dden-da3m0n: I can remove this again, no problem, but I will wait for @Maxr1998's comment regarding it since (IIRC) he liked the idea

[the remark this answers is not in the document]
Maxr1998: I don't have an opinion on this, so if Niels wants it removed, we should do that ^^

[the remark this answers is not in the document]
h1dden-da3m0n: gotcha and it went pooof :ghost:

.github/workflows/app-build.yml
```diff
@@ -0,0 +1,36 @@
...
+          java-version: 11
+
+      - name: Build Unsigned Release
+        run: ./gradlew --no-daemon --info assembleRelease
+
+      - name: Split APK release types
+        run: |
+          mkdir -p build/jellyfin-publish/release-libre build/jellyfin-publish/release-proprietary;
+          mv app/build/outputs/apk/*/*/jellyfin-android-*-libre-release-unsigned.apk build/jellyfin-publish/release-libre/;
+          mv app/build/outputs/apk/*/*/jellyfin-android-*-proprietary-release-unsigned.apk build/jellyfin-publish/release-proprietary/;
```
nielsvanvelzen, line 29: We need the debug versions here or just all .apk files. An unsigned release can't be installed on a device.

h1dden-da3m0n: would that not require a "debug-keystore" to be passed into the build step of the `app-build` workflow?

nielsvanvelzen (approved):
Note: do not merge until secrets are added.

.github/workflows/app-publish.yaml
anthonylavado: What are KEY_ALIAS and KEY_PASSWORD? I have added KEYSTORE and KEYSTORE_PASSWORD for this repo and Android TV.

anthonylavado: After further review, KEY_ALIAS has been added. I believe KEY_PASSWORD is unused here.

.github/workflows/app-publish.yaml
```diff
@@ -0,0 +1,73 @@
...
+
+on:
+  push:
+    tags:
+      - v*
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+    environment: release
```
nielsvanvelzen, line 11: environment is not used now, sorry :p

nielsvanvelzen (changes requested):

nielsvanvelzen (approved):

Maxr1998 (approved):
LGTM. Waiting for all secrets to be added before merging.

nielsvanvelzen (approved):

PR #335, opened 2021-04-07
ci(update-action): update actions/setup-java to v2

h1dden-da3m0n:
### Description

As mentioned on Matrix, `actions/setup-java` got updated to v2 and has possible breaking changes by defaulting to AzulJDK.
So this PR updates it to v2 while maintaining the previous behavior (by using AdoptOpenJDK).

### Changes

* update actions/setup-java to v2
* add dependabot for github-actions dependencies

### Issues

* n/a

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
+    labels:
+      - ci
+      - github-actions
+    commit-message:
+      prefix: ci
--- a/.github/workflows/app-test.yaml
+++ b/.github/workflows/app-test.yaml
@@ -14,8 +14,9 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v2
       - name: Setup Java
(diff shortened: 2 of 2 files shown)


Maxr1998 (approved):

nielsvanvelzen:
According to the setup-java docs the V1 release defaulted to Zulu: https://github.com/actions/setup-java#v2-vs-v1 and the V2 version requires setting the distribution. So even if dependabot would've updated it for us we would immediately see the CI failing.

Can I interest you in updating the kotlin-sdk repo too? 😄

h1dden-da3m0n:
> According to the setup-java docs the V1 release defaulted to Zulu: https://github.com/actions/setup-java#v2-vs-v1 and the V2 version requires setting the distribution. So even if dependabot would've updated it for us we would immediately see the CI failing.
> 
> Can I interest you in updating the kotlin-sdk repo too? smile

Hmmm, okay from what I read (arguably in another post, should have gone to the source :facepalm: ). 
Anyway I will sure open a PR there too (for some reason was still under the impression that the SDK was still using Azure DevOps :sweat_smile: )

PR #349, opened 2021-04-27
Update dependencies, drop JCenter

Maxr1998:


build.gradle.kts
```diff
@@ -4,21 +4,20 @@ buildscript {
     repositories {
         google()
         mavenCentral()
-        jcenter()
+        gradlePluginPortal()
```
nielsvanvelzen, line 7: I'd prefer the order:
1. maven central
2. gradle plugin portal (do we need this one?)
3. google

build.gradle.kts
```diff
@@ -4,21 +4,20 @@ buildscript {
...
         classpath(kotlin("gradle-plugin", kotlinVersion))
-        classpath("de.mannodermaus.gradle.plugins:android-junit5:1.6.2.0")
+        classpath("de.mannodermaus.gradle.plugins:android-junit5:1.7.1.1")
     }
 }
 
 allprojects {
     repositories {
         google()
         mavenCentral()
```
nielsvanvelzen, line 20: I'd prefer the order:

1. maven central
2. google

build.gradle.kts
Maxr1998: Apparently it's unneeded, it was for an older version of the JUnit plugin after removing JCenter..

nielsvanvelzen (approved):

PR #360, opened 2021-05-04
Use eps1lon/actions-label-merge-conflict

nielsvanvelzen:
jellyfin/jellyfin-sdk-kotlin#235

### DIFF
--- a/.github/workflows/merge-conflict-labeler.yaml
+++ b/.github/workflows/merge-conflict-labeler.yaml
@@ -1,4 +1,4 @@
-name: Merge conflict label
+name: Merge conflict labeler
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


.github/workflows/merge-conflict-labeler.yaml
```diff
@@ -1,16 +1,17 @@
 name: Merge conflict label
```
nielsvanvelzen, line 1:
```suggestion
name: Merge conflict labeler
```

nielsvanvelzen: good idea, sure is nice to be consistent with the other repos

Maxr1998 (approved):

PR #361, opened 2021-05-05
Dependency updates + Detekt

Maxr1998:


app/build.gradle.kts
```diff
@@ -1,15 +1,30 @@
...
     kotlin("android")
     kotlin("kapt")
     id("kotlin-parcelize")
+    id("io.gitlab.arturbosch.detekt") version Dependencies.Versions.detekt
     id("de.mannodermaus.android-junit5")
     id("com.github.ben-manes.versions") version Dependencies.Versions.dependencyUpdates
 }
 
+detekt {
+    buildUponDefaultConfig = true
```
nielsvanvelzen, line 16: Adding `ignoreFailures = true` will make sure Detekt reports all issues - handy for when we add linting to CI.

app/build.gradle.kts
```diff
@@ -1,15 +1,30 @@
...
+    allRules = false
+    config = files("${rootProject.projectDir}/detekt.yml")
+
+    reports {
+        html.enabled = true
+        xml.enabled = false
+        txt.enabled = true
```
nielsvanvelzen, line 23: Do we need txt reports? :p

app/build.gradle.kts
```diff
@@ -145,31 +157,45 @@ dependencies {
...
 }
 
-tasks.withType<Test> {
-    useJUnitPlatform()
-    testLogging {
-        outputs.upToDateWhen { false }
-        showStandardStreams = true
+tasks {
+    withType<Detekt> {
+        jvmTarget = JavaVersion.VERSION_1_8.toString()
```
nielsvanvelzen, line 162: I don't think this is needed?

app/build.gradle.kts
Maxr1998: They're nice to read and the overhead is zero, so sure.

app/build.gradle.kts
Maxr1998: Theoretically, I want to have all failure fixed. I'll start working on it once ExoPlayer is merged.

app/build.gradle.kts
nielsvanvelzen: Fair enough

app/build.gradle.kts
Maxr1998: Detekt recommends it in the setup. Idk.

app/build.gradle.kts
nielsvanvelzen: Yes but we want all the issues reported to GitHub so therefor it should be ignored.

app/build.gradle.kts
nielsvanvelzen: I didn't need it for other projects I've used Detekt and it works fine for those

app/build.gradle.kts
Maxr1998: Ok :+1:

app/build.gradle.kts
Maxr1998: Afaik it serves as a hint to detekt. So not 'needed', but might be helpful?

nielsvanvelzen: https://detekt.github.io/detekt/gradle.html#using-type-resolution

The default `detekt` task already uses Java 1.8

Maxr1998: I can't read that from there, but I'll remove it ig.

nielsvanvelzen (approved):

PR #381, opened 2021-05-11
Update SDK and rework server connection

### LINKED ISSUE #366: Error when trying to enter the server in the beta of Jellyfin for Android
**Describe the bug**
When putting the address of the server an error appears saying that I update to a more updated version of my server. But I have version 10.7.5, the last one released. And the exact same thing happens to many users. I hope they can fix it as quickly as possible

**Screenshots**
photo4945336013492955494

 - App version: [2.3.0-beta1]
 - Android version: [Android 10]
 - Device manufacturer and model: [LG k50S]
 - Jellyfin server version: [10.7.5]

Maxr1998:
Fixes #366


app/src/main/java/org/jellyfin/mobile/fragment/ConnectFragment.kt
```diff
@@ -149,13 +150,18 @@ class ConnectFragment : Fragment() {
         chooseServerButton.isVisible = false
     }
 
-    private fun showConnectionError(@StringRes errorString: Int = R.string.connection_error_cannot_connect) {
-        connectionErrorText.setText(errorString)
-        connectionErrorText.isVisible = true
+    private fun showConnectionError(error: String? = null) {
+        connectionErrorText.apply {
+            if (error != null) text = error else setText(R.string.connection_error_cannot_connect)
```
nielsvanvelzen, line 155:
```suggestion
           text =  if (error != null) error else getString(R.string.connection_error_cannot_connect)
```

nielsvanvelzen: Makes it easier to read imo

app/src/main/java/org/jellyfin/mobile/fragment/ConnectFragment.kt
```diff
@@ -164,46 +170,45 @@ class ConnectFragment : Fragment() {
...
-            version == null -> false
-            // Version too old
-            version < Jellyfin.minimumVersion -> false
-            // Incorrect product name
-            systemInfo.productName != "Jellyfin Server" -> false
-            else -> true
+        val server = greatServer ?: goodServers.firstOrNull()
+        if (server != null) {
+            val systemInfo = requireNotNull(server.systemInfo)
+            Timber.i("Found valid server at ${server.address} with version ${systemInfo.version}")
```
nielsvanvelzen, line 196: Log the score to, it's useful to know if a server was GOOD or GREAT

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/mobile/fragment/ConnectFragment.kt
Maxr1998: Not the same though :wink:
But in our case, probably should be.

nielsvanvelzen: oh yeah `?:` is a thing

nielsvanvelzen (approved):

PR #384, opened 2021-05-12
Replace GH_TOKEN with JF_BOT_TOKEN

ferferga:
``GH_TOKEN`` seems to be used internally by GitHub and it's making some actions we're using in some repos fail ([Source](https://github.com/alex-page/github-project-automation-plus/issues/39))

*(PRs made from users of the org are always successful, but PRs that aren't don't work, like Dependabot PRs. This can be checked by looking at the history of server repos and Jellyfin Vue)*

The org-wide secret should be renamed to something else to avoid confusion and further replacements or messes in GH's side. This PR replaces ``GH_TOKEN`` with ``JF_BOT_TOKEN``, which should be pretty generic.

### DIFF
--- a/.github/workflows/merge-conflict-labeler.yaml
+++ b/.github/workflows/merge-conflict-labeler.yaml
@@ -14,4 +14,4 @@ jobs:
       - uses: eps1lon/actions-label-merge-conflict@v2.0.1
         with:
           dirtyLabel: merge conflict
-          repoToken: ${{ secrets.GH_TOKEN }}
+          repoToken: ${{ secrets.JF_BOT_TOKEN }}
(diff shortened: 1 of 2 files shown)


nielsvanvelzen:
While you're at it, can you add the token to gradlew-update.yaml too?

ferferga:
@nielsvanvelzen Done!

.github/workflows/merge-conflict-labeler.yaml
```diff
@@ -8,10 +8,10 @@ on:
 
 jobs:
   triage:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
nielsvanvelzen, line 11:
```suggestion
    runs-on: ubuntu-20.04
```

Please keep the version specific image

nielsvanvelzen (changes requested):

Maxr1998 (approved):

nielsvanvelzen (approved):

PR #389, opened 2021-05-14
Extended webview checks

### LINKED ISSUE #383: Webview is too old, but I just installed ROM and GApps (2.3.0-beta1)
The Webview version Jellyfin-Android requires is 88. My system only provides com.android.webview 83.

The "check for updates" button redirects to the Play store, but the store doesn't seem to find the app.

Also I tried installing the webview I could find on the Play store, it's com.google.android.webview 90, but it does not help the issue.

FYI, the rom is POSP and the gapps package is NikGApps. My android knowledge isn't great, IDK if the webview you're expecting is provided by the system, the default browser, or the gapps.

nielsvanvelzen:
Fixes #383

Adds an additional button to open the webview settings, it allows changing the used webview more easily. User needs to manually open the app again after changing it.
For the second alert (outdated webview) pressing on the background will cancel the dialog and never show it again. Adding an additional button was too much work because I'd need to make my own layout for it.

image
image
_note: this one actually says "open settings" instead of "settings" but I was too lazy to re-capture_

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/MainActivity.kt
+++ b/app/src/main/java/org/jellyfin/mobile/MainActivity.kt
@@ -59,7 +59,7 @@ class MainActivity : AppCompatActivity() {
         bindService(Intent(this, RemotePlayerService::class.java), serviceConnection, Service.BIND_AUTO_CREATE)
 
         // Check WebView support
-        if (!appPreferences.ignoreWebViewChecks && !isWebViewSupported()) {
+        if (!isWebViewSupported()) {
             AlertDialog.Builder(this).apply {
                 setTitle(R.string.dialog_web_view_not_supported)
                 setMessage(R.string.dialog_web_view_not_supported_message)
--- a/app/src/main/java/org/jellyfin/mobile/MainActivity.kt
+++ b/app/src/main/java/org/jellyfin/mobile/MainActivity.kt
@@ -4,9 +4,12 @@ import android.app.Service
 import android.content.ComponentName
 import android.content.Intent
 import android.content.ServiceConnection
+import android.os.Build
 import android.os.Bundle
 import android.os.IBinder
+import android.provider.Settings
 import android.view.OrientationEventListener
+import android.widget.Toast
 import androidx.appcompat.app.AlertDialog
 import androidx.appcompat.app.AppCompatActivity
 import androidx.lifecycle.lifecycleScope
@@ -56,13 +59,20 @@ class MainActivity : AppCompatActivity() {
         bindService(Intent(this, RemotePlayerService::class.java), serviceConnection, Service.BIND_AUTO_CREATE)
 
         // Check WebView support
-        if (!isWebViewSupported()) {
+        if (!appPreferences.ignoreWebViewChecks && !isWebViewSupported()) {
             AlertDialog.Builder(this).apply {
                 setTitle(R.string.dialog_web_view_not_supported)
                 setMessage(R.string.dialog_web_view_not_supported_message)
                 setCancelable(false)
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+                    setNeutralButton(R.string.dialog_button_open_settings) { _, _ ->
+                        startActivity(Intent(Settings.ACTION_WEBVIEW_SETTINGS))
+                        Toast.makeText(context, R.string.toast_reopen_after_change, Toast.LENGTH_LONG).show()
+                        finishAfterTransition()
+                    }
+                }
                 setNegativeButton(R.string.dialog_button_close_app) { _, _ ->
-                    finish()
+                    finishAfterTransition()
                 }
             }.show()
             return
@@ -103,7 +113,11 @@ class MainActivity : AppCompatActivity() {
         requestCode: Int,
         permissions: Array<out String>,
         grantResults: IntArray
-    ) = permissionRequestHelper.handleRequestPermissionsResult(requestCode, permissions, grantResults)
+    ) {
+        super.onRequestPermissionsResult(requestCode, permissions, grantResults)
+
+        permissionRequestHelper.handleRequestPermissionsResult(requestCode, permissions, grantResults)
+    }
 
     override fun onSupportNavigateUp(): Boolean {
         onBackPressed()
--- a/app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
+++ b/app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
@@ -113,8 +113,6 @@ class WebViewFragment : Fragment(), NativePlayerHost {
             }
         }
 
-        appPreferences.ignoreWebViewChecks = false
-
         // Setup WebView
         webView.initialize()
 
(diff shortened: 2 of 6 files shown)


app/src/main/java/org/jellyfin/mobile/MainActivity.kt
```diff
@@ -103,7 +113,11 @@ class MainActivity : AppCompatActivity() {
         requestCode: Int,
         permissions: Array<out String>,
         grantResults: IntArray
-    ) = permissionRequestHelper.handleRequestPermissionsResult(requestCode, permissions, grantResults)
+    ) {
+        super.onRequestPermissionsResult(requestCode, permissions, grantResults)
+
+        permissionRequestHelper.handleRequestPermissionsResult(requestCode, permissions, grantResults)
+    }
```
Maxr1998, line 120: Ah lol, I wanted to fix this *for ages*

app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
```diff
@@ -109,6 +113,8 @@ class WebViewFragment : Fragment(), NativePlayerHost {
             }
         }
 
+        appPreferences.ignoreWebViewChecks = false
+
```
Maxr1998, line 117: Debugging left-overs?

```suggestion
```

app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
```diff
@@ -126,15 +132,14 @@ class WebViewFragment : Fragment(), NativePlayerHost {
     }
 
     private fun WebView.initialize() {
-        if (isOutdated()) { // Check WebView version
+        if (!appPreferences.ignoreWebViewChecks && isOutdated()) { // Check WebView version
             AlertDialog.Builder(requireContext()).apply {
                 setTitle(R.string.dialog_web_view_outdated)
                 setMessage(R.string.dialog_web_view_outdated_message)
-                setCancelable(false)
```
Maxr1998, line 133: I'm not sure it should be that easy. People will not read and/or accidentally dismiss the prompt..

app/src/main/java/org/jellyfin/mobile/MainActivity.kt
nielsvanvelzen: We should actually move it to contracts but not in the scope of this PR

app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
nielsvanvelzen: Yes!

app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
nielsvanvelzen: Adding 4 buttons requires a custom layout tho

nielsvanvelzen: Maybe replace "close app" with "ignore"?

Maxr1998: Yes, good idea.

nielsvanvelzen:
image
Updated screenshot

Maxr1998:
Maybe the "ignore" should be at the bottom?

nielsvanvelzen:
image

app/src/main/java/org/jellyfin/mobile/MainActivity.kt
```diff
@@ -56,13 +59,20 @@ class MainActivity : AppCompatActivity() {
         bindService(Intent(this, RemotePlayerService::class.java), serviceConnection, Service.BIND_AUTO_CREATE)
 
         // Check WebView support
-        if (!isWebViewSupported()) {
+        if (!appPreferences.ignoreWebViewChecks && !isWebViewSupported()) {
```
Maxr1998, line 62: It already checks for ignore here, but there's no way to ignore from the "unsupported" dialog below.. seems like an oversight?

nielsvanvelzen: I should definitely not work on the app when I'm half asleep 😅

CarlosOlivo:
You forgot to handle [the note](https://developer.android.com/reference/android/provider/Settings#ACTION_WEBVIEW_SETTINGS), also necessary for devices without gapps (or alternative store like Aurora Store || F-Droid).

> In some cases, a matching Activity may not exist, so ensure you safeguard against this.

```
android.content.ActivityNotFoundException: No Activity found to handle Intent { act=android.intent.action.VIEW dat=market://details?id=com.google.android.webview (has extras) }
	at android.app.Instrumentation.checkStartActivityResult(Instrumentation.java:2067)
	at android.app.Instrumentation.execStartActivity(Instrumentation.java:1727)
	at android.app.Activity.startActivityForResult(Activity.java:5320)
	at androidx.activity.ComponentActivity.startActivityForResult(ComponentActivity.java:574)
	at android.app.Activity.startActivityForResult(Activity.java:5278)
	at androidx.activity.ComponentActivity.startActivityForResult(ComponentActivity.java:560)
	at android.app.Activity.startActivity(Activity.java:5664)
	at androidx.core.content.ContextCompat.startActivity(ContextCompat.java:251)
	at androidx.fragment.app.FragmentHostCallback.onStartActivityFromFragment(FragmentHostCallback.java:166)
	at androidx.fragment.app.Fragment.startActivity(Fragment.java:1377)
	at androidx.fragment.app.Fragment.startActivity(Fragment.java:1365)
	at org.jellyfin.mobile.fragment.WebViewFragment.initialize$lambda-9$lambda-6(WebViewFragment.kt:161)
	at org.jellyfin.mobile.fragment.WebViewFragment.lambda$Pt7vAIcI7Sig4ZNp7P1oGVz8n9g(Unknown Source:0)
	at org.jellyfin.mobile.fragment.-$$Lambda$WebViewFragment$Pt7vAIcI7Sig4ZNp7P1oGVz8n9g.onClick(Unknown Source:6)
	at androidx.appcompat.app.AlertController$ButtonHandler.handleMessage(AlertController.java:167)
	at android.os.Handler.dispatchMessage(Handler.java:106)
	at android.os.Looper.loop(Looper.java:223)
	at android.app.ActivityThread.main(ActivityThread.java:7656)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:592)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:947)
```

nielsvanvelzen:
Thanks @CarlosOlivo, added an additional check to see if the intent can be resolved 👍

app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
```diff
@@ -126,35 +130,48 @@ class WebViewFragment : Fragment(), NativePlayerHost {
...
+                        authority(context.packageName)
+                    }.build()
+
+                    val marketIntent = Intent(Intent.ACTION_VIEW).apply {
+                        data = marketUri
+                        putExtra(Intent.EXTRA_REFERRER, referrerUri)
+                    }
+
+                    // Only show button if the intent can be resolved
+                    if(marketIntent.resolveActivity(context.packageManager) != null) {
```
Maxr1998, line 157:
```suggestion
                    if (marketIntent.resolveActivity(context.packageManager) != null) {
```

Maxr1998 (approved):

Maxr1998 (approved):

PR #398, opened 2021-05-22
fix incorrect rsync options

h1dden-da3m0n:
### Description

With the deployment/release of the Android app we ran into a few issues with the new GitHub actions based workflow, most of which got already fixed but apparently I messed up the `rsync`options in my initial PR :man_facepalming:.
This PR aims to fix that and hopefully get the workflow 100% operational for the next release.

### Changes

* remove obsolete/release breaking `rsync` options

### Issues

* n/a?

### DIFF
--- a/.github/workflows/app-publish.yaml
+++ b/.github/workflows/app-publish.yaml
@@ -50,8 +50,8 @@ jobs:
       - name: Upload to repo.jellyfin.org
         uses: burnett01/rsync-deployments@4.1
         with:
-          switches: -rltgoDzvO --delete --exclude='*' --include='**/*.apk' --include='*.txt'
-          path: build/jellyfin-publish/
+          switches: -vrptz
+          path: build/jellyfin-publish/**/
           remote_path: /srv/repository/releases/client/android/versions/v${{ env.JELLYFIN_VERSION }}
           remote_host: ${{ secrets.DEPLOY_HOST }}
           remote_user: ${{ secrets.DEPLOY_USER }}


.github/workflows/app-publish.yaml
```diff
@@ -50,7 +50,7 @@ jobs:
       - name: Upload to repo.jellyfin.org
         uses: burnett01/rsync-deployments@4.1
         with:
-          switches: -rltgoDzvO --delete --exclude='*' --include='**/*.apk' --include='*.txt'
+          switches: -avzrtO --delete
```
Maxr1998, line 53:
```suggestion
          switches: -avz
```

- `-a` implies `-rt`
- Why the `-O`? Shouldn't be needed
- `--delete` might be unsafe and isn't necessary - if we retag a release, the filenames shouldn't change
- On the `-a`: do we really need to preserve owner, group, ...? `-rpt` might be enough (ignoring owner, group, links and device/special files).

Maxr1998 (changes requested):

.github/workflows/app-publish.yaml
h1dden-da3m0n: I could have sworn that I read a does not imply `r`which is why its there

joshuaboniface: I agree with removing `--delete`. This could have dangerous potential implications based on the directory. The collection task should *move* (`mv`) the files to their final locations thus removing them and leaving the directory empty anyways.

joshuaboniface: `-a` does imply `-r` ;-)

Maxr1998: > I could have sworn that I read `-a` does not imply `-r`which is why its there

[`-a` is equivalent to that neat string `-rlptgoD`](https://linux.die.net/man/1/rsync). *I also had to look it up earlier.*

Maxr1998 (approved):
LGTM. That should work now, hopefully. :crossed_fingers:

PR #410, opened 2021-05-29
Update SDK snapshot version to master-SNAPSHOT

nielsvanvelzen:
**Changes**
- Update SDK snapshot version to master-SNAPSHOT
  - See jellyfin/jellyfin-sdk-kotlin#264

**Issues**

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -19,7 +19,7 @@ allprojects {
         google()
         maven("https://s01.oss.sonatype.org/content/repositories/snapshots/") {
             content {
-                includeGroup(Dependencies.Groups.jellyfin)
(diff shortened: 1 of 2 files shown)


build.gradle.kts
```diff
@@ -19,7 +19,7 @@ allprojects {
         google()
         maven("https://s01.oss.sonatype.org/content/repositories/snapshots/") {
             content {
-                includeGroup(Dependencies.Groups.jellyfin)
+                includeVersionByRegex(Dependencies.Groups.jellyfin, ".*", ".*-SNAPSHOT")
```
nielsvanvelzen, line 22: Are you sure this won't be a bad regex because the `.` in this constant is not escaped?????????????

nielsvanvelzen: yes should be fine

Maxr1998 (approved):

PR #436, opened 2021-06-15
Handle WebView error when loadUrl failed

### LINKED ISSUE #422: After Sign out and opening the app again I got blank screen. I turned Wifi off, then it refreshed page.
I think I signed out from my Server (local network) then some hours later I opened the app again but it only showed a blank white screen. When I turned off Wifi it refreshed the page so that I could enter my newly created Server address and port.

CarlosOlivo:
And replace blank screen with Jellyfin logo

# Before

### Server unavailable
iAGQWFOboB

### Server available
IIpiZYXb5Y

# After

### Server unavailable
jzZUYpOdhB

### Server available
WpzF63DRLK

Closes #422

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
+++ b/app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
@@ -210,6 +210,7 @@ class WebViewFragment : Fragment(), NativePlayerHost {
             override fun onReceivedSslError(view: WebView, handler: SslErrorHandler?, error: SslError) {
                 super.onReceivedSslError(view, handler, error)
                 Timber.e("Received SSL error: %s", error.toString())
+
                 if (error.url == view.url) onErrorReceived()
             }
         }
@@ -290,7 +291,6 @@ class WebViewFragment : Fragment(), NativePlayerHost {
 
     fun onConnectedToWebapp() {
         connected = true
-        webView.animate().alpha(1f).start()
         (activity as? MainActivity)?.requestNoBatteryOptimizations()
     }
 
--- a/app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
+++ b/app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
@@ -207,9 +207,9 @@ class WebViewFragment : Fragment(), NativePlayerHost {
                 if (request.url == Uri.parse(view.url)) onErrorReceived()
             }
 
-            override fun onReceivedSslError(view: WebView, handler: SslErrorHandler?, error: SslError) {
-                super.onReceivedSslError(view, handler, error)
+            override fun onReceivedSslError(view: WebView, handler: SslErrorHandler, error: SslError) {
                 Timber.e("Received SSL error: %s", error.toString())
+                handler.cancel()
 
                 if (error.url == view.url) onErrorReceived()
             }
--- a/app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
+++ b/app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
@@ -3,6 +3,7 @@ package org.jellyfin.mobile.fragment
 import android.content.Intent
 import android.graphics.Rect
 import android.net.Uri
+import android.net.http.SslError
 import android.os.Build
 import android.os.Bundle
 import android.provider.Settings
@@ -11,6 +12,7 @@ import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import android.webkit.ConsoleMessage
+import android.webkit.SslErrorHandler
 import android.webkit.WebChromeClient
 import android.webkit.WebResourceRequest
 import android.webkit.WebResourceResponse
@@ -195,14 +197,20 @@ class WebViewFragment : Fragment(), NativePlayerHost {
                 val errorMessage = errorResponse.data?.run { bufferedReader().use(Reader::readText) }
                 Timber.e("Received WebView HTTP %d error: %s", errorResponse.statusCode, errorMessage)
 
-                if (request.url == Uri.parse(server.hostname)) onErrorReceived()
+                if (request.url == Uri.parse(view.url)) onErrorReceived()
             }
 
             override fun onReceivedError(view: WebView, request: WebResourceRequest, error: WebResourceErrorCompat) {
                 val description = if (WebViewFeature.isFeatureSupported(WebViewFeature.WEB_RESOURCE_ERROR_GET_DESCRIPTION)) error.description else null
                 Timber.e("Received WebView error at %s: %s", request.url.toString(), description)
 
-                if (request.url == Uri.parse(server.hostname)) onErrorReceived()
+                if (request.url == Uri.parse(view.url)) onErrorReceived()
+            }
+
+            override fun onReceivedSslError(view: WebView, handler: SslErrorHandler?, error: SslError) {
+                super.onReceivedSslError(view, handler, error)
+                Timber.e("Received SSL error: %s", error.toString())
+                if (error.url == view.url) onErrorReceived()
             }
         }
         webChromeClient = object : WebChromeClient() {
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
```diff
@@ -199,14 +199,14 @@ class WebViewFragment : Fragment(), NativePlayerHost {
                 val errorMessage = errorResponse.data?.run { bufferedReader().use(Reader::readText) }
                 Timber.e("Received WebView HTTP %d error: %s", errorResponse.statusCode, errorMessage)
 
-                if (request.url == Uri.parse(server.hostname)) onErrorReceived()
+                if (request.url.equals(Uri.parse(view.url))) onErrorReceived()
```
Maxr1998, line 202: Same thing in Kotlin
```suggestion
                if (request.url == Uri.parse(view.url)) onErrorReceived()
```

app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
```diff
@@ -199,14 +199,14 @@ class WebViewFragment : Fragment(), NativePlayerHost {
...
             }
 
             override fun onReceivedError(view: WebView, request: WebResourceRequest, error: WebResourceErrorCompat) {
                 val description = if (WebViewFeature.isFeatureSupported(WebViewFeature.WEB_RESOURCE_ERROR_GET_DESCRIPTION)) error.description else null
                 Timber.e("Received WebView error at %s: %s", request.url.toString(), description)
 
-                if (request.url == Uri.parse(server.hostname)) onErrorReceived()
+                if (request.url.equals(Uri.parse(view.url))) onErrorReceived()
```
Maxr1998, line 209: See above
```suggestion
                if (request.url == Uri.parse(view.url)) onErrorReceived()
```

app/src/main/res/layout/activity_main.xml
```diff
@@ -6,6 +6,13 @@
     android:layout_height="match_parent"
     tools:viewBindingIgnore="true">
 
+    <ImageView
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        android:adjustViewBounds="true"
+        android:scaleType="fitCenter"
+        android:src="@drawable/ic_launcher_foreground" />
```
Maxr1998, line 14: It's a bit huge imo, would be better to give it a specific size.

Maxr1998 (changes requested):
Some feedback points.
Also, have you checked for any other side effects of changing the url to match for error cases?
Since the alpha is defined in the layout, how does the app handle configuration changes now?

app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
CarlosOlivo:
```
/**
* Compares this Uri to another object for equality. Returns true if the
* encoded string representations of this Uri and the given Uri are
* equal. Case counts. Paths are not normalized. If one Uri specifies a
* default port explicitly and the other leaves it implicit, they will not
* be considered equal.
*/
public boolean equals(Object o) {
   if (!(o instanceof Uri)) {
       return false;
   }

   Uri other = (Uri) o;

   return toString().equals(other.toString());
}
```

In this case, wouldn't it be better to skip the Uri parse and compare the strings directly?

`if (request.url.toString() == view.url) onErrorReceived()`

app/src/main/res/layout/activity_main.xml
CarlosOlivo: > It's a bit huge imo, would be better to give it a specific size.

For example?

### Original jellyfin-web splash screen
qemu-system-x86_64_HA2v5pOwUD
qemu-system-x86_64_mJvGNuxwLV

### Current PR
qemu-system-x86_64_frIjxPnNdh
qemu-system-x86_64_5AP8l2aNyi

### Current PR + Padding 64dp
qemu-system-x86_64_fHJc9BukHT
qemu-system-x86_64_3ics5wUvtV
studio64_rxE2VbQdfh


### Current PR + width 256dp
studio64_HUpXzxzZ2U
studio64_ExoVJRVHDq


Or anything else?

CarlosOlivo:
> Some feedback points.
> Also, have you checked for any other side effects of changing the url to match for error cases?

For example? Here are some error cases:
```
Received WebView error at http://10.0.2.2:8096/: net::ERR_CONNECTION_TIMED_OUT
REQUEST=http://10.0.2.2:8096/ WEBVIEW=http://10.0.2.2:8096/
MATCH

Received WebView error at http://10.0.2.2:8096/: net::ERR_INTERNET_DISCONNECTED
REQUEST=http://10.0.2.2:8096/ WEBVIEW=http://10.0.2.2:8096/
MATCH

Received WebView HTTP 404 error: null
REQUEST=http://10.0.2.2:8096// WEBVIEW=http://10.0.2.2:8096//
MATCH

Received WebView error at http://10.0.2.2:8096/web/7025.8047dd014c638223df19.bundle.js: net::ERR_FAILED
REQUEST=http://10.0.2.2:8096/web/7025.8047dd014c638223df19.bundle.js WEBVIEW=http://10.0.2.2:8096/
NO_MATCH

Received WebView HTTP 401 error: null
REQUEST=https://demo.jellyfin.org/stable/System/Info WEBVIEW=https://demo.jellyfin.org/stable/web/index.html
NO_MATCH
```

> Since the alpha is defined in the layout, how does the app handle configuration changes now?

In the same way as before, since configuration changes are handled by `MainActivity` and `FragmentManager` via `FragmentContainerView`, not `WebViewFragment`
if not so the `WebView` will be destroyed at each orientation change for example

app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
Maxr1998: I think I deliberately chose to parse it as Uri since the parsing would normalize the URLs, thus I would keep it like that.

app/src/main/res/layout/activity_main.xml
Maxr1998: The problem is that the current launcher icon includes quite a bit of padding itself. Good question, I'll probably have to play with the layout myself on-device.

Maxr1998:
> > Also, have you checked for any other side effects of changing the url to match for error cases?
> 
> For example? Here are some error cases:
> 
> ```
> Received WebView error at http://10.0.2.2:8096/: net::ERR_CONNECTION_TIMED_OUT
> REQUEST=http://10.0.2.2:8096/ WEBVIEW=http://10.0.2.2:8096/
> MATCH
> 
> Received WebView error at http://10.0.2.2:8096/: net::ERR_INTERNET_DISCONNECTED
> REQUEST=http://10.0.2.2:8096/ WEBVIEW=http://10.0.2.2:8096/
> MATCH
> 
> Received WebView HTTP 404 error: null
> REQUEST=http://10.0.2.2:8096// WEBVIEW=http://10.0.2.2:8096//
> MATCH
> 
> Received WebView error at http://10.0.2.2:8096/web/7025.8047dd014c638223df19.bundle.js: net::ERR_FAILED
> REQUEST=http://10.0.2.2:8096/web/7025.8047dd014c638223df19.bundle.js WEBVIEW=http://10.0.2.2:8096/
> NO_MATCH
> 
> Received WebView HTTP 401 error: null
> REQUEST=https://demo.jellyfin.org/stable/System/Info WEBVIEW=https://demo.jellyfin.org/stable/web/index.html
> NO_MATCH
> ```

Alright, seems to be good then. I think I missed something when looking at the code back then.

> > Since the alpha is defined in the layout, how does the app handle configuration changes now?
> 
> In the same way as before, since configuration changes are handled by `MainActivity` and `FragmentManager` via `FragmentContainerView`, not `WebViewFragment`
> if not so the `WebView` will be destroyed at each orientation change for example

You're right, that should be good.

CarlosOlivo:
Add a missing case, handle SSL errors, as both are strings and in the case of `SslError.url` it's a non-null, there should be no problems with wrong matches.

Maxr1998:
Only rebased for now.

Maxr1998 (approved):

Maxr1998:
As said in the first commit, it makes more sense to move the UI rework into a separate PR. I have some other things I want to refactor as well, so I will handle this in the next days.

PR #439, opened 2021-06-17
Improve JS injection method

CarlosOlivo:
No more undefined `NativeShell`

Before

chrome_1FT2lPrgak

```
V/WebView: ApiClient serverAddress: http://10.0.2.2:8096, http://10.0.2.2:8096/web/main.4b4b736dbefe7c3a90ec.bundle.js (2)
V/WebView: ApiClient appName: undefined, http://10.0.2.2:8096/web/main.4b4b736dbefe7c3a90ec.bundle.js (2)
V/WebView: ApiClient appVersion: undefined, http://10.0.2.2:8096/web/main.4b4b736dbefe7c3a90ec.bundle.js (2)
V/WebView: ApiClient deviceName: undefined, http://10.0.2.2:8096/web/main.4b4b736dbefe7c3a90ec.bundle.js (2)
V/WebView: ApiClient deviceId: undefined, http://10.0.2.2:8096/web/main.4b4b736dbefe7c3a90ec.bundle.js (2)
```

After

chrome_xPapohG59v

```
V/WebView: ApiClient serverAddress: http://10.0.2.2:8096, http://10.0.2.2:8096/web/main.4b4b736dbefe7c3a90ec.bundle.js?1623960060466 (2)
V/WebView: ApiClient appName: Jellyfin Android, http://10.0.2.2:8096/web/main.4b4b736dbefe7c3a90ec.bundle.js?1623960060466 (2)
V/WebView: ApiClient appVersion: 0.0.0-dev.1, http://10.0.2.2:8096/web/main.4b4b736dbefe7c3a90ec.bundle.js?1623960060466 (2)
V/WebView: ApiClient deviceName: sdk_gphone_x86, http://10.0.2.2:8096/web/main.4b4b736dbefe7c3a90ec.bundle.js?1623960060466 (2)
V/WebView: ApiClient deviceId: 9c17977ed5b293fb, http://10.0.2.2:8096/web/main.4b4b736dbefe7c3a90ec.bundle.js?1623960060466 (2)
```


Maxr1998 (dismissed):
Looks good! I will do some more on-device testing tomorrow and merge it if I don't find any side-effects.

app/src/main/assets/native/injectionScript.js
```diff
@@ -0,0 +1,16 @@
...
+    scripts.forEach(function(src) {
+        var scriptElement = document.createElement('script');
+        scriptElement.type = 'text/javascript';
+        scriptElement.src = src;
+        scriptElement.charset = 'utf-8';
+        scriptElement.setAttribute('defer', '');
+        document.body.appendChild(scriptElement);
+    });
+    document.currentScript.remove();
+}();
```
nielsvanvelzen, line 16: Made some changes  to the script, mainly using ES6 syntax since the webview requires it and it makes the code slightly easier to read. I also moved the cache buster so we're 100% sure the injected scripts are not outdated (although I think that's not possible anyway). Also moved the script elements to the head instead of body.

```suggestion
(() => {
    const scripts = [
        '/native/nativeshell.js',
        '/native/EventEmitter.js',
        document.currentScript.src
    ];

    for (const script of scripts) {
        const scriptElement = document.createElement('script');
        scriptElement.src = script.concat('?', Date.now());
        scriptElement.charset = 'utf-8';
        scriptElement.setAttribute('defer', '');

        document.head.appendChild(scriptElement);
    }
})();
```

nielsvanvelzen (changes requested):

app/src/main/assets/native/injectionScript.js
CarlosOlivo: > I also moved the cache buster so we're 100% sure the injected scripts are not outdated (although I think that's not possible anyway).

Yes, it is definitely not possible, this is because all intercepted resources are never cached.

image

While the rest of the resources are cached

image

So I'm going to leave it as it was to make the logcat slightly easier to read

CarlosOlivo:
I found that [WebViewAssetLoader.AssetsPathHandler](https://developer.android.com/reference/kotlin/androidx/webkit/WebViewAssetLoader.AssetsPathHandler#handle) exists, so I removed `WebAppUtils.kt`

app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
```diff
@@ -157,18 +156,14 @@ class WebViewFragment : Fragment(), NativePlayerHost {
...
+                    !connected && path.endsWith(Constants.ANY_BUNDLE_PATH) -> {
+                        onConnectedToWebapp()
+                        AssetsPathHandler(context).handle("native/injectionScript.js")
                     }
-                    path.contains("native") -> webView.context.loadAsset("native/${url.lastPathSegment}")
+                    path.contains("/native/") -> AssetsPathHandler(context).handle("native/${url.lastPathSegment}")
                     path.endsWith(Constants.CAST_SDK_PATH) -> {
                         // Load the chrome.cast.js library instead
-                        webView.context.loadAsset("native/chrome.cast.js")
+                        AssetsPathHandler(context).handle("native/chrome.cast.js")
```
Maxr1998, line 166: The `AssetPathHandler` should be a cached, shared instance as a property of the fragment.

Maxr1998 (changes requested):

nielsvanvelzen (dismissed):

Maxr1998 (changes requested):
If the site refreshes without the fragment getting recreated (can be simulated with a forced refresh of the page), this `!connected` check breaks the injection.

I'll try to fix it now by reworking the check, but please comment if you have other/better suggestions.

app/src/main/assets/native/injectionScript.js
```diff
@@ -0,0 +1,15 @@
...
+        '/native/nativeshell.js',
+        '/native/EventEmitter.js',
+        document.currentScript.src.concat('?', Date.now())
+    ];
+    for (const script of scripts) {
+        const scriptElement = document.createElement('script');
+        scriptElement.src = script;
+        scriptElement.charset = 'utf-8';
+        scriptElement.setAttribute('defer', '');
+        document.body.appendChild(scriptElement);
```
Maxr1998, line 12: @nielsvanvelzen I changed this back, please see my comment in the commit.

app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
```diff
@@ -157,19 +158,14 @@ class WebViewFragment : Fragment(), NativePlayerHost {
...
-                        null // continue loading normally
-                    }
-                    path.contains("native") -> webView.context.loadAsset("native/${url.lastPathSegment}")
-                    path.endsWith(Constants.CAST_SDK_PATH) -> {
-                        // Load the chrome.cast.js library instead
-                        webView.context.loadAsset("native/chrome.cast.js")
+                    path.matches(Constants.MAIN_BUNDLE_PATH_REGEX) && url.query.isNullOrEmpty() -> {
+                        onConnectedToWebapp()
+                        assetsPathHandler.handle("native/injectionScript.js")
                     }
```
Maxr1998, line 164: I'm not a fan of the regex, but this is the most robust implementation I could come up with. Feel free to suggest different methods.

app/src/main/assets/native/injectionScript.js
nielsvanvelzen: I'd still put it in the head tbh

app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
```diff
@@ -158,15 +158,14 @@ class WebViewFragment : Fragment(), NativePlayerHost {
                 val url = request.url
                 val path = url.path?.lowercase(Locale.ROOT) ?: return null
                 return when {
-                    !connected && path.endsWith(Constants.ANY_BUNDLE_PATH) -> {
+                    path.matches(Constants.MAIN_BUNDLE_PATH_REGEX) && url.query.isNullOrEmpty() -> {
```
nielsvanvelzen, line 161: I'd add a special query parameter in the injection script the check here. I can imagine jellyfin-web adding a cache buster at some point which would break this detection

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
Maxr1998: > I'd still put it in the head tbh

But why?

Maxr1998: > I'd add a special query parameter in the injection script the check here. I can imagine jellyfin-web adding a cache buster at some point which would break this detection

That'd also break appending the timestamp in the JS..

Maxr1998: Also, isn't the hash inside the filename already a cache-buster?

app/src/main/assets/native/injectionScript.js
Maxr1998: But why? I don't see a reason for changing the current behavior of Jellyfin Web.

app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
CarlosOlivo: > I'd add a special query parameter in the injection script the check here. I can imagine jellyfin-web adding a cache buster at some point which would break this detection

Don't worry, that will not happen because it is already [included](https://github.com/jellyfin/jellyfin-web/blob/6fc729f59351ef3bc313eea0e1da0dfa3236233a/webpack.common.js#L77) in the filename.

> v10.7.5 `main.4b4b736dbefe7c3a90ec.bundle.js`
> v10.7.6 `main.0f183787a2f78055e856.bundle.js`

> Also, isn't the hash inside the filename already a cache-buster?

Yep

app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
```diff
@@ -157,18 +158,14 @@ class WebViewFragment : Fragment(), NativePlayerHost {
...
                 val path = url.path?.lowercase(Locale.ROOT) ?: return null
                 return when {
-                    path.endsWith(Constants.WEB_CONFIG_PATH) -> {
-                        runOnUiThread {
-                            webView.evaluateJavascript(JS_INJECTION_CODE) {
-                                onConnectedToWebapp()
-                            }
-                        }
-                        null // continue loading normally
+                    !connected && path.endsWith(Constants.ANY_BUNDLE_PATH) -> {
```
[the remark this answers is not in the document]
CarlosOlivo: I have some ideas, I will try them first

But is there such a way to force the refresh of the page without editing the current source code? How do you simulate the refresh of the page without editing the source code?, I thought that with a pull-to-refresh gesture or some option within jellyfin-web but I don't see any

app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
Maxr1998: Did some more tweaks, should be good now.

Maxr1998 (dismissed):

app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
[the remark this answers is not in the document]
CarlosOlivo: Ok my suggestion, add [WebViewClient.shouldOverrideUrlLoading](https://developer.android.com/reference/android/webkit/WebViewClient#shouldOverrideUrlLoading(android.webkit.WebView,%20android.webkit.WebResourceRequest))

```
override fun shouldOverrideUrlLoading(view: WebView, request: WebResourceRequest): Boolean {
    if (request.url == Uri.parse(view.url)) connected = false
    return super.shouldOverrideUrlLoading(view, request)
}
```

This is called before shouldInterceptRequest so it should work. Could you try it? This way the regex is no longer needed.

CarlosOlivo:
See https://github.com/jellyfin/jellyfin-android/pull/439#discussion_r654810080 I think it is better than running a regex comparison for each request.

app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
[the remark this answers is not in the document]
Maxr1998: That should work, but it's less comprehensible.. while the Regex performs worse, I still think I prefer it..

app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
nielsvanvelzen: The fact that the hash is included in the filename doesn't mean someone can add a query string to the file in the future. I think it's important to create an implementation that is future-proof.

app/src/main/java/org/jellyfin/mobile/utils/Constants.kt
```diff
@@ -12,7 +12,7 @@ object Constants {
 
     // Webapp constants
     const val MINIMUM_WEB_VIEW_VERSION = 80
-    const val WEB_CONFIG_PATH = "config.json"
+    val MAIN_BUNDLE_PATH_REGEX = Regex(""".*/main.[a-f\d]+.bundle.js""")
```
nielsvanvelzen, line 15: How can we be sure the hash of the bundle never changes? If webpack was updated and they decided to use sha sums instead of md4 it might break.

app/src/main/assets/native/injectionScript.js
nielsvanvelzen: Scripts belong in the <head> element, although in HTML5 that isn't required anymore. Not a huge issue, don't have a strong preference for it either.

app/src/main/java/org/jellyfin/mobile/utils/Constants.kt
Maxr1998: It shouldn't, the default digest is hex, regardless of the hashing algorithm.

nielsvanvelzen: Ok sha hashes was a wrong example, but my point still stands

app/src/main/assets/native/injectionScript.js
CarlosOlivo: Just as a comment but actually `jellyfin-web` does inject the scripts into head tag
https://github.com/jellyfin/jellyfin-web/blob/0dde17fbd7b8318269de53784543cb2981715dce/src/components/castSenderApi.js#L22

app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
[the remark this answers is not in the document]
CarlosOlivo: I think that in these cases the solution should always be the one with the best performance, but it works anyway.

app/src/main/java/org/jellyfin/mobile/utils/Constants.kt
Maxr1998: I'd say it's highly unlikely for Jellyfin Web or Webpack to switch to a different digest - the web app will be EOL before that..

CarlosOlivo: Anyway, nothing is future-proof, if I already happened at least 2 times with v10.6 and v10.7, I don't doubt that it will change again for v10.8, unless the native client appears before, we have the webview hybrid client for a while...

app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
Maxr1998: Yep, that's why I changed it. I think this one can be resolved.

app/src/main/assets/native/injectionScript.js
Maxr1998: Interesting, webpack puts the bundle into the body though..

app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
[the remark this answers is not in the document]
nielsvanvelzen: I think the current way with the query parameter and regex should work fine and it keeps the injections/url handling in one place. Performance shouldn't differ too much, could maybe swap the query parameter check to be the first as that should be faster.

Maxr1998 (approved):

nielsvanvelzen (approved):

PR #441, opened 2021-06-17
Follow skip forward/back length on ExoPlayer

### LINKED ISSUE #84: Skip forward and back doesn't follow settings in exoplayer
Pressing the right side of the screen doesn't skip the set 30s but only 5s

**To Reproduce**
1. Set forward and back skip to 30 seconds in settings
2. Start video
3. Skip forward and backward by double tapping the sides of the screen

 - App version: 2.0.0-rc.2
 - Android version: Android 10
 - Device manufacturer and model: OnePlus 6t
 - Jellyfin server version: 10.6.3
 - The type of video player implementation use: integrated player (Exoplayer)

CarlosOlivo:
Closes #84

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/controller/ApiController.kt
+++ b/app/src/main/java/org/jellyfin/mobile/controller/ApiController.kt
@@ -3,17 +3,13 @@ package org.jellyfin.mobile.controller
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
 import org.jellyfin.mobile.AppPreferences
-import org.jellyfin.mobile.model.DisplayPreferences
 import org.jellyfin.mobile.model.sql.dao.ServerDao
 import org.jellyfin.mobile.model.sql.dao.UserDao
 import org.jellyfin.mobile.model.sql.entity.ServerEntity
-import org.jellyfin.mobile.utils.Constants.DEFAULT_SEEK_TIME_MS
 import org.jellyfin.sdk.api.client.ApiClient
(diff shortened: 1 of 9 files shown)


Maxr1998 (approved):

Maxr1998:
Add an extra model class to store the display preferences, that way the number only has to be parsed once. Rest works well.

app/src/main/java/org/jellyfin/mobile/controller/ApiController.kt
```diff
@@ -63,11 +72,35 @@ class ApiController(
...
+                    displayPreferencesId = "usersettings",
+                    client = "emby"
+                )
+
+                val customPrefs = displayPreferences.customPrefs
+
+                DisplayPreferences(
+                    skipBackLength = customPrefs?.get("skipBackLength")?.toLongOrNull() ?: DEFAULT_SEEK_TIME_MS,
+                    skipForwardLength = customPrefs?.get("skipForwardLength")?.toLongOrNull() ?: DEFAULT_SEEK_TIME_MS
+                )
```
nielsvanvelzen, line 94: Should use constants for these

nielsvanvelzen (changes requested):
Instead of adding a fragile HTTP interceptor it might be easier to just GET the display preferences when the player starts. That should be more reliable.

Maxr1998:
True that, I'll do that.

nielsvanvelzen (approved):

Maxr1998 (approved):

PR #442, opened 2021-06-17
Remember brightness changes on ExoPlayer

CarlosOlivo:
Closes #418

PR #444, opened 2021-06-19
Follow state changes for brightness & gesture preference

CarlosOlivo:
Complement of #442

I think I found an error in your library @Maxr1998, or I am using it incorrectly?

```
D/SettingsFragment: appPreferences.exoPlayerAllowSwipeGestures = true
D/SettingsFragment: swipeGesturesPreference.checked = false
D/SettingsFragment: swipeGesturesPreference.getBoolean(false) = false
```

Could you check why it returns different values?


Maxr1998:
> ```
> D/SettingsFragment: appPreferences.exoPlayerAllowSwipeGestures = true
> D/SettingsFragment: swipeGesturesPreference.checked = false
> D/SettingsFragment: swipeGesturesPreference.getBoolean(false) = false
> ```
> 
> Could you check why it returns different values?

That's normal, when building the preference tree, it isn't attached to the context yet, so it returns default values. Once the screen builder finishes, it resolves all data from SharedPreferences.

Maxr1998 (approved):

PR #445, opened 2021-06-19
Restructure Webapp UI

Maxr1998:
- Move CoordinatorLayout to WebViewFragment, since it's only supposed to be used there anyways
- Fade in WebView once the app has connected to the sever
- Fix a potential NPE when setting the locale

The background icon as discussed in #436 is still missing, but this should be easy to add now. I will play with that tomorrow.

cc: @CarlosOlivo

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/utils/UIExtensions.kt
+++ b/app/src/main/java/org/jellyfin/mobile/utils/UIExtensions.kt
@@ -51,12 +51,12 @@ fun View.applyWindowInsetsAsMargins() {
 fun View.fadeIn() {
     alpha = 0f
     isVisible = true
-    with(animate()) {
+    animate().apply {
         alpha(1f)
         @Suppress("MagicNumber")
         duration = 300L
         interpolator = LinearOutSlowInInterpolator()
-        start()
+        withLayer()
     }
 }
 
--- a/app/src/main/java/org/jellyfin/mobile/utils/UIExtensions.kt
+++ b/app/src/main/java/org/jellyfin/mobile/utils/UIExtensions.kt
@@ -14,7 +14,9 @@ import androidx.annotation.StringRes
 import androidx.annotation.StyleRes
 import androidx.core.view.ViewCompat
 import androidx.core.view.WindowInsetsCompat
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/jellyfin/mobile/utils/UIExtensions.kt
```diff
@@ -46,6 +48,18 @@ fun View.applyWindowInsetsAsMargins() {
...
+fun View.fadeIn() {
+    alpha = 0f
+    isVisible = true
+    with(animate()) {
+        alpha(1f)
+        @Suppress("MagicNumber")
+        duration = 300L
+        interpolator = LinearOutSlowInInterpolator()
+        start()
+    }
```
nielsvanvelzen, line 60:
```suggestion
    animate().apply {
        alpha(1f)
        @Suppress("MagicNumber")
        duration = 300L
        interpolator = LinearOutSlowInInterpolator()
        withLayer()
    }
```
- start() is not needed, will automatically start
- withLayer() enables hardware acceleration

app/src/main/res/layout/fragment_webview.xml
```diff
@@ -1,5 +1,12 @@
 <?xml version="1.0" encoding="utf-8"?>
-<WebView xmlns:android="http://schemas.android.com/apk/res/android"
+<androidx.coordinatorlayout.widget.CoordinatorLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
nielsvanvelzen, line 2: could probably use a framelayout here

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/mobile/utils/UIExtensions.kt
Maxr1998: - neat, didn't know
- that's not the default??

app/src/main/res/layout/fragment_webview.xml
Maxr1998: Nope, we need that for the snackbar.

nielsvanvelzen (approved):

PR #447, opened 2021-06-20
Improve session integration

CarlosOlivo:
- Changed `PLAYER_TIME_UPDATE_RATE` from 1 second to 10 seconds
 Too much spam before, to the level of slowing down the server (at least in my case), also is the default in [jellyfin-web](https://github.com/jellyfin/jellyfin-web/blob/0dde17fbd7b8318269de53784543cb2981715dce/src/components/playback/playbackmanager.js#L2751-L2755) 
- Stop active encoding if transcoding when playbacks ends

---

Originally 2 sessions are created because jellyfin-web creates a [singleton](https://github.com/jellyfin/jellyfin-web/blob/2edf4661d5b19dcc59414ba3c157f2ff1ca71713/src/components/ServerConnections.js#L20-L39) per device (many users - one device), while jellyfin-sdk-kotlin proposes one user per [session](https://github.com/jellyfin/jellyfin-sdk-kotlin/blob/a1d7596c4ef3bf3cb69eaab9b7872d3b0e5d350a/jellyfin-model/src/main/kotlin/org/jellyfin/sdk/model/DeviceInfo.kt#L7-L10) (one user - one device)

In the case of jellyfin-web, as it is a Singleton, once NativeShell is [injected](https://github.com/jellyfin/jellyfin-android/blob/530e01f67590077b8d1e2fc558b390e92b293b44/app/src/main/java/org/jellyfin/mobile/bridge/NativeInterface.kt#L54-L73) the deviceId should not change or it causes unexpected behaviors

So `jellyfin-web` and `jellyfin-sdk-kotlin` must come to an agreement on what to do in these cases.

---

Before

rs0PFEElVk

After

6XhogObBHn

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/api/ApiModule.kt
+++ b/app/src/main/java/org/jellyfin/mobile/api/ApiModule.kt
@@ -6,6 +6,7 @@ import org.jellyfin.sdk.api.client.ApiClient
 import org.jellyfin.sdk.api.client.KtorClient
 import org.jellyfin.sdk.api.operations.ArtistsApi
 import org.jellyfin.sdk.api.operations.GenresApi
+import org.jellyfin.sdk.api.operations.HlsSegmentApi
 import org.jellyfin.sdk.api.operations.ImageApi
 import org.jellyfin.sdk.api.operations.ItemsApi
 import org.jellyfin.sdk.api.operations.MediaInfoApi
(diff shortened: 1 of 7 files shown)


Maxr1998:
I've rebased your PR and reverted the device ID change for now, since it has undesired side-effects for multiple users on one device, I have an idea how to fix this in a separate PR however.

Maxr1998 (approved):

PR #465, opened 2021-07-12
Add aspect ratio menu to ExoPlayer

CarlosOlivo:


Maxr1998:
This seems to break the zoom gesture that's already present. I'm personally not sure if this menu is necessary at all (especially the weirder modes that stretch the content), but for feature parity I suppose it's generally a good idea.

PR #466, opened 2021-07-13
Use AuthorizationHeaderBuilder to encode device name send to web client

nielsvanvelzen:
This makes sure the SDK and webui use an identical name. Requires SDK 1.0.1 to function properly.


app/src/main/java/org/jellyfin/mobile/bridge/NativeInterface.kt
```diff
@@ -60,9 +61,7 @@ class NativeInterface(private val fragment: WebViewFragment) : KoinComponent {
             // normalize the name by removing special characters
             // and making sure it's at least 1 character long
             // otherwise the webui will fail to send it to the server
-            val name = deviceInfo.name
-                .replace("[^\\x20-\\x7E]".toRegex(), "")
-                .trim()
+            val name = AuthorizationHeaderBuilder.encodeParameterValue(deviceInfo.name)
                 .padStart(1)
```
Maxr1998, line 65: Could probably move this into a single line ¯\\\_(ツ)\_/¯
```suggestion
            val name = AuthorizationHeaderBuilder.encodeParameterValue(deviceInfo.name).padStart(1)
```

Maxr1998 (approved):

PR #482, opened 2021-08-04
Migrate to Gradle version catalogs

Maxr1998:
I guess since everything builds, it shouldn't require a very close review.. but it won't hurt either.


build.gradle.kts
```diff
@@ -19,13 +19,13 @@ allprojects {
         google()
         mavenLocal {
             content {
-                includeVersionByRegex(Dependencies.Groups.jellyfin, ".*", Dependencies.Versions.jellyfinSdkLocal)
+                includeVersionByRegex(JellyfinSdk.GROUP, ".*", JellyfinSdk.LOCAL)
```
nielsvanvelzen, line 22: Why not `libs.versions.jellyfinSdkLocal`?

gradle/libs.versions.toml
```diff
@@ -0,0 +1,135 @@
...
+androidx-test-runner = "1.4.0"
+androidx-test-espresso = "3.4.0"
+
+[libraries]
+# KotlinX
+coroutines-core = { group = "org.jetbrains.kotlinx", name = "kotlinx-coroutines-core", version.ref = "coroutines" }
+coroutines-android = { group = "org.jetbrains.kotlinx", name = "kotlinx-coroutines-android", version.ref = "coroutines" }
+
+# Core
+koin = { group = "io.insert-koin", name = "koin-android", version.ref = "koin" }
```
nielsvanvelzen, line 57: I don;t see a point in separating the version if it's only used once.

gradle/libs.versions.toml
```diff
@@ -0,0 +1,135 @@
...
+# Testing
+junit = "5.7.2"
+kotest = "4.6.1"
+mockk = "1.12.0"
+androidx-test-runner = "1.4.0"
+androidx-test-espresso = "3.4.0"
+
+[libraries]
+# KotlinX
+coroutines-core = { group = "org.jetbrains.kotlinx", name = "kotlinx-coroutines-core", version.ref = "coroutines" }
```
nielsvanvelzen, line 53: What are your rules to determine between group+name or module?

nielsvanvelzen: (IMO module is always shorter and therefor easier to read)

gradle/libs.versions.toml
```diff
@@ -0,0 +1,135 @@
...
+coroutines-android = { group = "org.jetbrains.kotlinx", name = "kotlinx-coroutines-android", version.ref = "coroutines" }
+
+# Core
+koin = { group = "io.insert-koin", name = "koin-android", version.ref = "koin" }
+androidx-core = { module = "androidx.core:core-ktx", version.ref = "androidx-core" }
+androidx-appCompat = { module = "androidx.appcompat:appcompat", version.ref = "androidx-appcompat" }
```
nielsvanvelzen, line 59:
```suggestion
androidx-appcompat = { module = "androidx.appcompat:appcompat", version.ref = "androidx-appcompat" }
```

Officially called "Appcompat" without uppercase C

gradle/libs.versions.toml
```diff
@@ -0,0 +1,135 @@
...
+androidx-activity = { module = "androidx.activity:activity-ktx", version.ref = "androidx-activity" }
+androidx-fragment = { module = "androidx.fragment:fragment-ktx", version.ref = "androidx-fragment" }
+androidDesugarLibs = { module = "com.android.tools:desugar_jdk_libs", version.ref = "androidDesugarLibs" }
+
+# Lifecycle Extensions
+androidx-lifecycleExtensions-viewModel = { module = "androidx.lifecycle:lifecycle-viewmodel-ktx", version.ref = "androidx-lifecycleExtensions" }
```
nielsvanvelzen, line 65: I'd shorten it to `androidx-lifecycle-xxx`, also viewmodel is without uppercase M. https://developer.android.com/jetpack/androidx/releases/lifecycle

gradle/libs.versions.toml
```diff
@@ -0,0 +1,135 @@
...
+androidx-lifecycleExtensions-liveData = { module = "androidx.lifecycle:lifecycle-livedata-ktx", version.ref = "androidx-lifecycleExtensions" }
```
nielsvanvelzen, line 66: no uppercase D, actually should just lowercase everything

buildSrc/src/main/kotlin/JellyfinSdk.kt
```diff
@@ -0,0 +1,7 @@
+object JellyfinSdk {
+    const val GROUP ="org.jellyfin.sdk"
```
nielsvanvelzen, line 2:
```suggestion
    const val GROUP = "org.jellyfin.sdk"
```

nielsvanvelzen (changes requested):

build.gradle.kts
Maxr1998: Good point, I'll fix that. Forgot that this worked too.

gradle/libs.versions.toml
Maxr1998: I disagree, it allows you to see all versions in a single list.

gradle/libs.versions.toml
Maxr1998: It was a bit random, I mostly split group/name when having multiple of the same group. But it's not really consistent. Not sure what'd be the recommended style (shorter is not really an argument imo).

gradle/libs.versions.toml
Maxr1998: Right, that's a copy paste error. I fixed it for some other deps.

buildSrc/src/main/kotlin/JellyfinSdk.kt
Maxr1998: Huh, how did that happen..

build.gradle.kts
Maxr1998: Nvm, that doesn't work. It fails the build.

Maxr1998:
I recommend a squash after review ^^

build.gradle.kts
nielsvanvelzen: Maybe it needs Gradle 7.2 :(

Maxr1998: I'd merge it as is then.

nielsvanvelzen (approved):

PR #496, opened 2021-08-11
Workaround for server connection failure on libre release builds

Maxr1998:


nielsvanvelzen:
Replaced with jellyfin/jellyfin-sdk-kotlin#300

PR #503, opened 2021-08-16
Properly open live streams before playback if necessary

jdlayman:
Previously the media source resolver did not check whether a media source required opening. Now, the media resolver appropriately opens the live stream for a media source, when required.

Even after the recent changes in fc05f2a5f104757438c77a861b098abc4da87fb0, the app still wouldn't open LiveTV for me using a HDHR Connect Duo. I took a peek at the web client, and realized that the app wasn't checking to see whether the media source had the requiresOpening flag set. Adding the code from my PR seems to resolve things for me.

_Note: my expertise is C and not Kotlin, so let me know if you have any feedback/recommendations. I'm not as fluent with the functional languages._

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/player/source/MediaSourceResolver.kt
+++ b/app/src/main/java/org/jellyfin/mobile/player/source/MediaSourceResolver.kt
@@ -34,6 +34,7 @@ class MediaSourceResolver(
                     audioStreamIndex = audioStreamIndex,
                     subtitleStreamIndex = subtitleStreamIndex,
                     maxStreamingBitrate = /* 1 GB/s */ 1_000_000_000,
+                    autoOpenLiveStream = true,
                 ),
             )
 


app/src/main/java/org/jellyfin/mobile/player/source/MediaSourceResolver.kt
```diff
@@ -36,10 +37,30 @@ class MediaSourceResolver(
...
+                                audioStreamIndex = audioStreamIndex,
+                                subtitleStreamIndex = subtitleStreamIndex,
+                                maxStreamingBitrate = /* 1 GB/s */ 1_000_000_000,
+                            )
+                        )
+                        liveStreamResponse.mediaSource ?: return Result.failure(PlayerException.UnsupportedContent())
+                    }
+                    null -> return Result.failure(PlayerException.UnsupportedContent())
+                }
+            }
```
Maxr1998, line 63: I'd replace the let call with an extra variable and use an if instead of when (just keep the null check as is).

Like so:
```kotlin
            val source = response.mediaSources?.let { sources ->
                sources.find { source -> source.id?.toUUIDOrNull() == itemId } ?: sources.firstOrNull()
            } ?: return Result.failure(PlayerException.UnsupportedContent())

            if (source.requiresOpening) {
                val liveStreamResponse by mediaInfoApi.openLiveStream(
                     data = OpenLiveStreamDto(
                         openToken = source.openToken,
                         itemId = itemId,
                         userId = apiClient.userId,
                         playSessionId = response.playSessionId,
                         deviceProfile = deviceProfile,
                         startTimeTicks = startTimeTicks,
                         audioStreamIndex = audioStreamIndex,
                         subtitleStreamIndex = subtitleStreamIndex,
                         maxStreamingBitrate = /* 1 GB/s */ 1_000_000_000,
                     )
                )

                liveStreamResponse.mediaSource ?: Result.failure(PlayerException.UnsupportedContent())
            } else {
                source
            }
```

Maybe even extract the code in the if block into another (private) function.

Maxr1998 (changes requested):
Good catch! We might also want to close the media source when the livestream stops.

app/src/main/java/org/jellyfin/mobile/player/source/MediaSourceResolver.kt
jdlayman: Good call on suggestions. I will make the edits and take a look at closing the stream this evening.

Maxr1998: Regarding the closing: you could add a `requiresClosing` boolean flag to the constructor of the JellyfinMediaSource and process it in the PlayerViewModel.

The open/close methods could maybe be moved into the source class too, not sure if that works though.

jdlayman: I did some more investigation this evening, and I found a more elegant solution. Adding `autoOpenLiveStream = true` to the PlaybackInfoDto for the original playback info request accomplishes the same thing server side (see: [MediaInfoController.cs](https://github.com/jellyfin/jellyfin/blob/ea7c7ea48c49750c182bf1f2c9a2496c5297aa8f/Jellyfin.Api/Controllers/MediaInfoController.cs#L200)). Additionally, while reading through the server code, it seems like the server tracks how many consumers are active and closes the live stream itself when there are no more active players. I confirmed via logs that this occurs as expected.

The only downside I could see with this approach would be if the app was resolving media sources well before playing them (for example, when adding them to a queue). However, I found that even when items are queued, the media is not resolved until they become active.

I updated my branch to reflect the changes above. I've tested them locally and they work for me.

app/src/main/java/org/jellyfin/mobile/player/source/MediaSourceResolver.kt
```diff
@@ -34,6 +34,7 @@ class MediaSourceResolver(
                     audioStreamIndex = audioStreamIndex,
                     subtitleStreamIndex = subtitleStreamIndex,
                     maxStreamingBitrate = /* 1 GB/s */ 1_000_000_000,
+                    autoOpenLiveStream = true
```
Maxr1998, line 37: Please use trailing commas when possible.
```suggestion
                    autoOpenLiveStream = true,
```

Maxr1998 (changes requested):
The `autoOpenLivestream` attribute was a good idea, didn't know that existed. I still wonder whether closing the live stream might be necessary after playback ends. Could you check if there are any remaining processes after playback with this patch?

app/src/main/java/org/jellyfin/mobile/player/source/MediaSourceResolver.kt
jdlayman: Thanks for the quick review. I added the comma as requested.

Additionally, I looked at processes on both Android and the server to confirm there are no orphaned processes after video stops. I also confirmed the tuner no longer has any active sessions.

The following log items appear in my server log, which validates the code I saw in the repository last night. As you can see, once the app notifies the server that playback has stopped, the server determines the consumer count is zero and performs the necessary cleanup of the SharedHttpStream with the tuner and cancels/deletes the transcoding session.

> [2021-08-18 01:01:06.901 +00:00] [INF] [11] Emby.Server.Implementations.Session.SessionManager: Playback stopped reported by app "Jellyfin Android" "0.0.0-dev.1" playing "KSHB-TV". Stopped at "18165" ms
[2021-08-18 01:01:06.907 +00:00] [INF] [38] Emby.Server.Implementations.Library.MediaSourceManager: Live stream "native_686b312d3504901ba64fed43701d1733_9219b79e0dbf2b42af414df829c1f675" consumer count is now 0
[2021-08-18 01:01:06.907 +00:00] [INF] [38] Emby.Server.Implementations.Library.MediaSourceManager: Closing live stream "a17c75760a04e99b68cf766e11316e1c_09efa0d56b934a82adec00a87b837fb0_native_686b312d3504901ba64fed43701d1733_9219b79e0dbf2b42af414df829c1f675"
[2021-08-18 01:01:06.907 +00:00] [INF] [38] Emby.Server.Implementations.LiveTv.TunerHosts.HdHomerun.HdHomerunHost: Closing "SharedHttpStream"
[2021-08-18 01:01:06.907 +00:00] [INF] [38] Emby.Server.Implementations.Library.MediaSourceManager: Live stream "a17c75760a04e99b68cf766e11316e1c_09efa0d56b934a82adec00a87b837fb0_native_686b312d3504901ba64fed43701d1733_9219b79e0dbf2b42af414df829c1f675" closed successfully
[2021-08-18 01:01:07.220 +00:00] [INF] [22] Emby.Server.Implementations.LiveTv.TunerHosts.HdHomerun.HdHomerunHost: Copying of "SharedHttpStream" to "/config/transcodes/e75e9b1e6025408bb7245eb5aacb31c0.ts" was canceled
[2021-08-18 01:01:07.221 +00:00] [INF] [22] Emby.Server.Implementations.LiveTv.TunerHosts.HdHomerun.HdHomerunHost: Deleting temp files ["/config/transcodes/e75e9b1e6025408bb7245eb5aacb31c0.ts"]
[2021-08-18 01:01:17.691 +00:00] [INF] [35] Jellyfin.Api.Helpers.TranscodingJobHelper: Transcoding kill timer stopped for JobId "8554f07b8e234f84b35683592e405f7d" PlaySessionId "bb5fb5c834044d7f9b571f2582e4ae8a". Killing transcoding
[2021-08-18 01:01:17.692 +00:00] [INF] [35] Jellyfin.Api.Helpers.TranscodingJobHelper: Stopping ffmpeg process with q command for "/config/transcodes/c7e30cf6f09a787f455aea00a0661fc7.m3u8"
[2021-08-18 01:01:22.692 +00:00] [INF] [35] Jellyfin.Api.Helpers.TranscodingJobHelper: Killing FFmpeg process for "/config/transcodes/c7e30cf6f09a787f455aea00a0661fc7.m3u8"
[2021-08-18 01:01:22.692 +00:00] [INF] [35] Jellyfin.Api.Helpers.TranscodingJobHelper: Deleting partial stream file(s) "/config/transcodes/c7e30cf6f09a787f455aea00a0661fc7.m3u8"
[2021-08-18 01:01:22.703 +00:00] [ERR] [22] Jellyfin.Api.Helpers.TranscodingJobHelper: FFmpeg exited with code 137

Maxr1998 (approved):

PR #509, opened 2021-08-23
Also build app bundle for Google Play in CI

Maxr1998:


nielsvanvelzen:
I'd prefer keeping a flat file structure in repo.jellyfin.org. Easier to maintain and it doesn't make sense since the other files are already in a flat structure (non-signed release, debug version).

Maxr1998:
I just thought it might be confusing for users. But in a way having both signed and unsigned is already confusing, so..

nielsvanvelzen:
Users that would be confused by repo.jellyfin.org or the GitHub release page should just use Google Play / Amazon Appstore / F-Droid.

h1dden-da3m0n:
Interesting, so GPlay needs yet another bundle type :roll_eyes: .
Anyway, based on the PR title I initially thought that you modified the CI to auto publish to GPlay (AFAIK that would be possible too?)

The changes LGTM

Maxr1998:
Well, Google Play doesn't strictly *need* them, but since we're using Play App Signing already anyway, we can also migrate to bundles and save a few bytes when users download the app.

PR #526, opened 2021-09-07
Workaround for serviceWorker cache prevent injections

### LINKED ISSUE #494: Reverse proxy black screen with 2.3.x on JF 10.7.x
### Describe the bug

Hi all, 

Referring to the reverse proxy for "Nginx from a subdomain" as stated in the documentation https://jellyfin.org/docs/general/networking/nginx.html, 2.3.0 on JF 10.7.6 can only launched the first time after login. Close the app then reopen will result in a black screen permanently unless reinstalled or data cleared.

Things that has been tested:

1. Reverse proxy works correctly on previous version before 2.3.0.
2. Reverse proxy works correctly from webUI and Tizen with no errors in developer tools.
3. Tested with the exact snippet copied from the JF reverse proxy documentation except for these changes, "set $jellyfin 127.0.0.1;" and commented out the resolver.
4. Reverse proxy tested on both port 80 and 443 (Let's Encrypt).
5. Reverse proxy tested on Note 20 Ultra Android 11 and Android Studio API 29/30 clean image.
6. I'm not attaching any logcat or debug logs because there's no specific errors comparing to connecting directly to port 8096 which works correctly. Just a black screen when using the reverse proxy. I could see the websocket activities though in the logcat, server and Android Studio debug log that's just it. Hope someone could help to replicate the reverse proxy and point out where is the issue.
7. Last edit: Not sure if it matters but the subdomain is on IPV6 fallback to IPV4.

Thank you.

All running on debian 11 (bullseye)
JF 10.7.6
Nginx 1.18.0

### Logs

_No response_

### Application version

2.3.0

Google Store

### Device information

Samsung Note 20 Ultra

### Android version

Android 11

### Jellyfin server version

10.7.6

- [ ] Web player (default)

Nazar78:
Block Service Worker from caching that causes script injections to fail resulting in black screen or unable to connect on subsequent app restarts. 

Fixes #494

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
+++ b/app/src/main/java/org/jellyfin/mobile/fragment/WebViewFragment.kt
@@ -12,8 +12,6 @@ import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import android.webkit.ConsoleMessage
-import android.webkit.ServiceWorkerClient
-import android.webkit.ServiceWorkerController
 import android.webkit.SslErrorHandler
 import android.webkit.WebChromeClient
 import android.webkit.WebResourceRequest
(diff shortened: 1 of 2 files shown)


Maxr1998:
Moved and tweaked your code a little, using the compat APIs is generally preferred to ensure the best compatibility. Thanks again for investigating and the PR!

Maxr1998 (approved):

PR #552, opened 2021-11-01
Fallback to prefer extension renderers if decoder fails

Maxr1998:
Necessary to support e.g. AAC-LTP on some devices, without unnecessarily impacting battery life for codecs supported by the hardware.

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/player/PlayerViewModel.kt
+++ b/app/src/main/java/org/jellyfin/mobile/player/PlayerViewModel.kt
@@ -15,11 +15,13 @@ import androidx.lifecycle.viewModelScope
 import com.google.android.exoplayer2.C
 import com.google.android.exoplayer2.DefaultRenderersFactory
 import com.google.android.exoplayer2.ExoPlayer
+import com.google.android.exoplayer2.PlaybackException
 import com.google.android.exoplayer2.Player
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/jellyfin/mobile/player/PlayerViewModel.kt
```diff
@@ -450,6 +458,19 @@ class PlayerViewModel(application: Application) : AndroidViewModel(application),
...
     }
 
+    override fun onPlayerError(error: PlaybackException) {
+        if (error.cause is MediaCodecDecoderException && !fallbackPreferExtensionRenderers) {
+            Timber.e(error.cause, "Decoder failed, attempting to restart playback with decoder extensions preferred")
+            playerOrNull?.run {
+                removeListener(this@PlayerViewModel)
+                release()
+            }
+            fallbackPreferExtensionRenderers = true
```
nielsvanvelzen, line 468: And when will we put this back to false? 🤔

Maxr1998: We don't, but you normally don't play a lot of stuff in a row, and if you do, it's likely the same codec (e.g. a show).

nielsvanvelzen (approved):

PR #556, opened 2021-11-06
Build script refactoring

nielsvanvelzen:
- Update Android Gradle to 7.0.3
- Move all plugin configuration to version catalog
- Move Kotlin version to version catalog
- Switch from KAPT to KSP
  - Should improve build times since KSP doesn't do java stuff
  - Also hopefully fixes the occasional "cache is corrupt manually delete build dir to fix" issue
- Update Room to 2.4.0 beta - required for KSP 1.0.0 compatibility
- Use Android 12 when compiling
  - Also use it as target
  - Added `exported=true` in manifest for mainactivity (required by android 12)

I wanted to move the "all projects repositories" from build.gradle.kts to settings.gradle.kts using dependencyResolutionManagement but that wasn't possible because it uses buildsrc constants for the SDK stuff. So I didn't move that part, otherwise the root build.gradle.kts could become pretty much empty.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -3,12 +3,12 @@ import com.github.benmanes.gradle.versions.updates.gradle.GradleReleaseChannel
 import io.gitlab.arturbosch.detekt.Detekt
 
 plugins {
-    id("com.android.application")
-    kotlin("android")
-    kotlin("kapt")
-    id("kotlin-parcelize")
+    alias(libs.plugins.android.app)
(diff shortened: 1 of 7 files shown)


nielsvanvelzen:
I did most of these changes in one of my own apps (https://github.com/nielsvanvelzen/tv-launcher) and  it makes the build files way easier to understand IMO. Although our own build scripts are a bit more complicated with the flavors/signing/testing/publishing setup

Maxr1998 (approved):

PR #567, opened 2021-11-16
Fix potential crash when loading display preferences

Maxr1998:


app/src/main/java/org/jellyfin/mobile/player/PlayerViewModel.kt
```diff
@@ -133,6 +133,8 @@ class PlayerViewModel(application: Application) : AndroidViewModel(application),
                 )
             } catch (e: ApiClientException) {
                 Timber.e(e, "Failed to load display preferences")
+            } catch (e: IllegalArgumentException /* kotlinx.serialization.SerializationException */) {
```
nielsvanvelzen, line 136: You should add a `// TODO: Remove when SDK is updated to 10.8 >= based API`

nielsvanvelzen (approved):

PR #575, opened 2021-11-27
Fix inconsistent behaviour of the previous button

em1lyy:
in the player control UI, the previous button only skips to the previous
track if the current position is smaller than
`Constants.MAX_SKIP_TO_PREV_MS`. however, the previous button in the
notification does not respect the current playback progress. this is
inconsistent and confusing behaviour.
this patch makes the previous button in the notification behave the same
as the previous button in the player control UI.
this makes the behaviour more consistent and also now the same as many
other popular media players, such as VLC, Spotify, etc.

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/player/PlayerMediaSessionCallback.kt
+++ b/app/src/main/java/org/jellyfin/mobile/player/PlayerMediaSessionCallback.kt
@@ -24,7 +24,7 @@ class PlayerMediaSessionCallback(private val viewModel: PlayerViewModel) : Media
     }
 
     override fun onSkipToPrevious() {
-        viewModel.skipToPrevious(force = true)
+        viewModel.skipToPrevious()
     }
 
     override fun onSkipToNext() {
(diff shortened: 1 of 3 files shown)


Maxr1998:
Actually, the notification interface itself is already quite inconsistent..
Depending on the amount of playlist items in the queue, it either shows fast-forward/rewind buttons which always have the behavior that you're asking for, or previous/next which just jump between tracks. I'm thinking that it'd be better to outright remove on of those options or, instead of making it respond to the playlist size, be user-configurable in the settings.

em1lyy:
Oh really? I have never actually seen the fast-forward/rewind buttons, no matter what I do, after using the Android App for quite some time. Also, looking at PlayerNotificationHelper.kt#L162 , from what I understand reading PlayerViewModel.kt, the rewind button just rewinds a few seconds and does not skip to the beginning, second of which is the behaviour I wanted. I do not want to discredit you or something, I mean you have developed this app and not me, I'm just a little confused. I do see the conditionals choosing between the actions, but when I play a single-track album, it still shows the previous and next buttons; they just don't do anything. Outright removing the rewind/fast-forward buttons would be an option, but it wouldn't solve the actual problem I was trying to address; the fact that the *previous button itself* always skips to the previous track and not in a playback progress conditional way like the web UI player does, meaning that would either be an addition or a completely separate issue/PR. I can do that if you want me to, but still, it is a different issue.

em1lyy:
Any updates on this?

Maxr1998:
Not yet. Haven't had much time to work on the app recently.

Maxr1998 (approved):
I decided that the behavior that you're suggesting does make more sense in the context of a video player. Directly jumping to the previous track could be useful music maybe, even if it's not the common behavior, but is definitely not expected for an all-media-types video player. Thus, let's get this merged.

Maxr1998 (approved):

PR #591, opened 2021-12-17
Added support for android client side logs to settings fragment

khrosales1989:
Includes support for client side logs on android within the jellyfin setting, as well as some minor updates due to having to increase the target sdk  for 3rd party library support. Also addressed small code smells throughout. Includes two options, one for generating a logfile and another for deleting logs.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -7,8 +7,10 @@ plugins {
     kotlin("android")
     kotlin("kapt")
     id("kotlin-parcelize")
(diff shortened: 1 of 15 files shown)


nielsvanvelzen (commented):
Sorry for the late reply, development of our mobile app was slow for a while.

The changes in this PR are not something we'd like to introduce to the app. Adding better crash reporting is scheduled together with some changes in Jellyfin 10.8 that allows a client to upload them to the server. They will likely use ACRA. We're not interested in running logcat (which is likely blocked on a lot of devices).

PR #592, opened 2021-12-18
addressed insets bug when returning from media

khrosales1989:
When returning from fullscreen player fragment, the mainactivity wasn't returning to nonfullscreen mode without rotating screen. Added insets listener to ensure that mainactivity correctly displays content without needing to rotate when returning from player fragment


Maxr1998:
Thanks for your contribution. I fear the updated code won't work as expected though, since it sets the padding to a constant used for inset identification, not the actual insets. I have however reworked some of the insets handling recently and will open a PR to fix those issues soon.

Maxr1998:
I can't reproduce the issue you describe on both the integrated player and the web player, thus I'm closing this PR.

PR #593, opened 2021-12-21
added support for user default aspect ratio

khrosales1989:
The user default aspect ratio is for html Html Video Player only. Allows user to specify a default aspect ratio in the setting fragment


nielsvanvelzen:
I don't see why you chose to implement this in the Android app. This should be added to jellyfin-web instead. It's a change for the web player (which we intent to phase out eventually).

khrosales1989:
I'm new to this project. I saw an open issue, not labeled, and decided to work on it.

Maxr1998:
I understand. However, since the WebPlayer won't stay in the Android app in the long run, I think it may be best to add this to Jellyfin Web instead and create a preference there.
Thus, I'm closing this PR since integrating this into the app doesn't make sense. Thanks for your understanding.

PR #597, opened 2021-12-29
Update and rename full_description.txt to fdroid_full_description.txt

hankster112:
F-Droid variant does not have Chromecast support


Maxr1998:
To my knowledge, fdroid/fastlane current doesn't support any other filenames than `full_description`, or am I missing something?

nielsvanvelzen:
Closing as stale and invalid (this is not valid a fastlane file).

PR #611, opened 2022-01-24
Refactor application packages

Maxr1998:
Ok, this is a huge one. I noticed that the current package structure was both inconsistent and confusing, thus I decided to regroup all packages and introduce new (sub-)packages. This time, classes are grouped by domain instead of superclass or function, which should make working on individual features easier.


nielsvanvelzen (approved):

nielsvanvelzen:
Time to break all the other PR's

Maxr1998:
Git fortunately handles rebasing quite well…

PR #535, opened 2021-10-01
Support video bitrate selection

Maxr1998:
Add support for bitrate selection and improves transcoding playback.

Supersedes and closes #274.

Closes #24.

PR #631, opened 2022-03-27
Update dependencies

Maxr1998:
Update various dependencies. ExoPlayer is excluded from this since I want to update it as part of #535 (some APIs were deprecated and removed, but a rework before that PR isn't worth it).


nielsvanvelzen:
Please fix the newly introduced linter errors (probably by ignoring them)

nielsvanvelzen (approved):

PR #637, opened 2022-04-04
Use specified video version in native player

### LINKED ISSUE #585: Integrated player: Duration / Runtime seems to be taken from metadata instead of the actual data
### Describe the bug

1. Add two different versions of a movie (having different runtimes).
2. Play both versions and check the runtime. Both runtimes should be equal, so one of them is off.

The only explanation I can think of is that the runtime is not actually determined by the data, but from the metadata instead.

This only happens with the integrated player. The web player and any external players are fine. When seeking, this will also cause issues as the progress is not mapped accurately to the actual position in the stream. 

### Logs

_No response_

### Application version

2.4.2

F-Droid

### Device information

Realms 7

### Android version

Android 11

### Jellyfin server version

10.7.7

- [ ] Web player (default)

mingyech:
User can specify a version of a video they want to play, but previously the native player does not play the specified version and plays the default version instead.  Use the specified version from `playOptions` if available, similar to how it is handled in external player mode. 

Fixes #585.


Maxr1998 (approved):
LGTM and worked fine in my tests. Thank you!

PR #740, opened 2022-07-18
ExoPlayer: Use cronet where available

knackebrot:
[Cronet](https://exoplayer.dev/network-stacks.html#cronet) supports HTTP/3 and is better suited for streaming over lossy networks than Android's built-in stack. It's also recommended for this use case.
This commit adds a soft dependency on Cronet from Google Play Services and falls back to the previously used implementation if it's not available.

I don't know how a soft dependency on Google Play Services affects the libre build or if 16 MB in-memory cache is acceptable. Therefore, I welcome any feedback.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -142,6 +142,7 @@ dependencies {
     implementation(libs.androidx.mediarouter)
     implementation(libs.bundles.exoplayer)
     implementation(libs.jellyfin.exoplayer.ffmpegextension)
+    implementation(libs.exoplayer.cronet)
     @Suppress("UnstableApiUsage")
     proprietaryImplementation(libs.exoplayer.cast)
     @Suppress("UnstableApiUsage")
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -136,13 +136,15 @@ dependencies {
     }
     implementation(libs.okhttp)
     implementation(libs.coil)
+    implementation(libs.cronetembedded)
+    implementation(libs.cronetokhttp)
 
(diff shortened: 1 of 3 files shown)


Maxr1998:
I like the idea and appreciate the work you put into this, but I'm equally concerned about the additional propietary Google library. I'm not decided yet on our propietary build flavor with cast support, but we definitely cannot use this in the libre build.

How about trying the embedded cronet implementation though? That seems to be fully open source, but at the cost of increasing the APK size by 8 MB (which should be ok, but I cannot decide that myself).

knackebrot:
If I understand this correctly, the issue is that cronet exoplayer extension depends on `com.google.android.gms:play-services-cronet` (https://github.com/google/ExoPlayer/blob/dev-v2/extensions/cronet/build.gradle#L23) which is released under the Android SDK License, probably with no source code available. Cronet itself is open source. We could probably replace it with https://mvnrepository.com/artifact/org.microg.gms/play-services-cronet-core which is under Apache 2 license. I don't know if you can override dependencies of dependencies in gradle or we'd have to fork the extension.

knackebrot:
Alternatively, we can switch to OkHttp ExoPlayer extension. It's already listed as a dependency in jellyfin-android, supports HTTP/2, and you can plug in the embedded cronet library without adding a dependency on Google Play Services. Would you find this viable, @Maxr1998 ?

knackebrot:
All right, I switched it to OkHttp and cronet-embedded and it doesn't depend on any Google Play Services now. The apk is significantly bigger now, though.

Issue #731, opened 2022-07-12
Characters in lock screen media controls are URL encoded


colonelpopcorn:
### Describe the bug

1. Go to music
2. Select a track that has a character that needs to be URL encoded (e.g. a track by Coheed & Cambria)
3. Lock your device or pull down from the top to view Jellyfin's media controls

Expected Result:
Artist name should show up without any escaped characters.

Actual result:
Artist name shows up with URL encoded characters (e.g. `Coheed &amp; Cambria`)

### Application version

2.4.4

### Where did you install the app from?

Google Play

### Device information

moto g(8) plus

### Android version

Android 10

### Jellyfin server version

10.8.1

### Which video player implementations does this bug apply to?

- [ ] Web player (default)
- [ ] Integrated player (ExoPlayer)
- [ ] External player (VLC, mpv, MX Player)

PR #775, opened 2022-08-31
Update notify method to strip escaped HTML characters.

colonelpopcorn:
So, after doing some investigation I found that it was actually the `notify` method in the `RemotePlayerService` class. Not sure why some properties show escaped HTML characters instead of their standard equivalents, and it may be worth looking into why the characters are returned that way from the server. For now, however, this should suffice and allow special characters to be displayed without issue. See the before and after picture below for proof of bug fix, as well.
jelly-fin-before-after

EDIT: This addresses issue #731, btw


app/src/main/java/org/jellyfin/mobile/utils/HtmlUtils.kt
```diff
@@ -0,0 +1,12 @@
...
+import android.text.Html
+
+@Suppress("DEPRECATION")
+fun stripHtmlChars(content: String?): String {
+    return if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.N) {
+        Html.fromHtml(content, Html.FROM_HTML_MODE_LEGACY).toString()
+    } else {
+        Html.fromHtml(content).toString()
+    }
+}
```
Maxr1998, line 12: You should be able to replace this function with [`HtmlCompat.fromHtml()`](
https://developer.android.com/reference/androidx/core/text/HtmlCompat#fromHtml(java.lang.String,int)) called directly in `RemotePlayerService`. Note that the passed content string must be non-null, thus you should wrap the `fromHtml` call with a let, e.g. `title?.let { HtmlCompat.fromHtml(it, Html.FROM_HTML_MODE_LEGACY) }`.

Maxr1998 (changes requested):
Thanks for investigating this! Your fix looks good, but I found you could replace your helper function with a compat function from AndroidX.

app/src/main/java/org/jellyfin/mobile/utils/HtmlUtils.kt
colonelpopcorn: K, I made the fix as requested. For my edification, what happens if the null check returns true? Does the `it` lambda parameter fall through to null?

Maxr1998: If `title` is null, the let won't be executed and the whole statement evaluates to null.

Maxr1998 (approved):

nielsvanvelzen:
The title is also used in some other places like the media session metadata. I'm not sure if we should fix the HTML in those places too.

Maxr1998:
Are those shown somewhere where they wouldn't be decoded by the system?

nielsvanvelzen:
Media session metadata can be read by any app and I don't think they expect HTML in the strings.

Maxr1998:
Right. Should we still merge this already and have a different PR for the other issue?

nielsvanvelzen:
Yeah that's fine

PR #777, opened 2022-09-01
Add adaptive monochrome icon.

Irishsmurf:
Android 13 supports themed app icons.
This changes adds a monochromatic icon to `ic_launcher` and `ic_launcher_round`.

---

- **Before**:
Screenshot_20220901-224322

- **After**:
Screenshot_20220901-224904


Maxr1998 (approved):
This is ok for now, but we should probably add an extra variant of the icon with better scaling later.

PR #800, opened 2022-09-21
Rewrite connection screen in Jetpack Compose

Maxr1998:
Already somewhat ready for review, but I need to look at it another time tomorrow and might want to do some more testing.


gradle/libs.versions.toml
```diff
@@ -71,6 +75,7 @@ coroutines-android = { group = "org.jetbrains.kotlinx", name = "kotlinx-coroutin
 
 # Core
 koin = { group = "io.insert-koin", name = "koin-android", version.ref = "koin" }
+koin-compose = { group = "io.insert-koin", name = "koin-androidx-compose", version = "3.2.0" }
```
nielsvanvelzen, line 78: Add a [versions] name

app/src/main/res/values/strings_donottranslate.xml
```diff
@@ -1,4 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <string name="app_name" translatable="false">Jellyfin</string>
+    <string name="app_name_short" translatable="false">@string/app_name</string>
```
nielsvanvelzen, line 4: If we add "Jellyfin" here we don't need to add the line in the debug flavor

app/src/main/java/org/jellyfin/mobile/setup/ComposeFragment.kt
```diff
@@ -0,0 +1,46 @@
...
+import androidx.fragment.app.Fragment
+import org.jellyfin.mobile.MainViewModel
+import org.jellyfin.mobile.databinding.FragmentComposeBinding
+import org.jellyfin.mobile.ui.screens.connect.ConnectScreen
+import org.jellyfin.mobile.ui.utils.AppTheme
+import org.jellyfin.mobile.utils.Constants
+import org.jellyfin.mobile.utils.applyWindowInsetsAsMargins
+import org.koin.androidx.viewmodel.ext.android.sharedViewModel
+
+class ComposeFragment : Fragment() {
```
nielsvanvelzen, line 18: Why not call it ConnectFragment like before? It still only does connecting

gradle/libs.versions.toml
```diff
@@ -90,6 +95,16 @@ androidx-constraintlayout = { group = "androidx.constraintlayout", name = "const
...
 
+# Compose
+compose-runtime = { group = "androidx.compose.runtime", name = "runtime", version.ref = "compose" }
+compose-ui = { group = "androidx.compose.ui", name = "ui", version.ref = "compose" }
+compose-foundation = { group = "androidx.compose.foundation", name = "foundation", version.ref = "compose" }
+compose-animation = { group = "androidx.compose.animation", name = "animation", version.ref = "compose" }
+compose-material = { group = "androidx.compose.material", name = "material", version.ref = "compose" }
+compose-material-icons = { group = "androidx.compose.material", name = "material-icons-core", version.ref = "compose" }
+compose-material-icons-extended = { group = "androidx.compose.material", name = "material-icons-extended", version.ref = "compose" }
+compose-runtime-livedata = { group = "androidx.compose.runtime", name = "runtime-livedata", version.ref = "compose" }
```
nielsvanvelzen, line 106: I don't think you're using livedata in any compose code right now, and if you are; it should be migrated to stateflow's.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/mobile/setup/ComposeFragment.kt
Maxr1998: For now, yes. It'll probably handle more in the future or be outright removed but I can of course rename it.

gradle/libs.versions.toml
Maxr1998: You're right, I think it can be removed.

gradle/libs.versions.toml
Maxr1998: Good catch. It should actually just use the default koin version, but the koin maintainers forgot to push this artifact for the latest version, thus I had to downgrade..

app/src/main/java/org/jellyfin/mobile/setup/ComposeFragment.kt
nielsvanvelzen: If we add more to it in the future we can rename it at that point. I'd imagine the webview will be embedded directly in a compose screen using https://google.github.io/accompanist/webview/

Maxr1998: Oh, I didn't know there was an Accompanist library for this. I'd just have used `AndroidView`. Thanks for the link!

nielsvanvelzen: I used that in one of my own apps after I used an AndroidView myself and ran into an issue :-)

gradle/libs.versions.toml
nielsvanvelzen: The compose part is in a separate repository (the normal androidx libraries are in a monorepo with core) so I think splitting into two versions is the way to go, worst case we bump two versions at the same time (although Renovate helps with that).

app/src/main/res/values/colors.xml
```diff
@@ -3,13 +3,15 @@
...
     <color name="jellyfin_primary_dark">#101010</color>
+    <!-- Secondary -->
     <color name="jellyfin_accent">#00a4dc</color>
 
     <!-- Launcher icon background -->
     <color name="ic_launcher_background">#000b25</color>
 
     <!-- App colors -->
     <color name="theme_background">#101010</color>
+    <color name="theme_surface">#363636</color>
```
nielsvanvelzen, line 14: Do we want to keep using xml files for theming stuff or hardcode them in AppTheme/constants? (I prefer the latter)

Maxr1998: I think we can hardcode them, yes.

Maxr1998: And done.

nielsvanvelzen (approved):
yolo

PR #805, opened 2022-09-21
Raise compileSdk to 33

Maxr1998:
I think this should be enough for the moment to merge #780.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -21,7 +21,7 @@ detekt {
 }
 
 android {
-    compileSdk = 32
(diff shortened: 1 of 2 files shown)


nielsvanvelzen (commented):
Do media notifications (now playing) still work in Android 13 with this PR?

Maxr1998:
Yes, the `compileSdk` doesn't affect the behavior of the app, it only allows to use new APIs without imposing any restrictions. The `targetSdk` would need to be raised to enable new behaviors.

nielsvanvelzen (approved):

PR #809, opened 2022-09-23
don't automatically change orientation for landscape video

### LINKED ISSUE #141: Enable portrait mode
Would like to enable the portrait mode .

Not to be locked on landscape mode when playing video .

Best regards.

eglia:
This change prevents the screen orientation from automatically changing on playback start in ExoPlayer.

The player now keeps the currently active screen orientation. When the device is rotated (with auto-rotate active), the player rotates accordingly. The player is also rotated when pressing the fullscreen button.

Only works in ExoPlayer, the web player still shows the old behaviour. I'm not sure where the corresponding code is for the web player.

Probably fixes #141

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/app/AppPreferences.kt
+++ b/app/src/main/java/org/jellyfin/mobile/app/AppPreferences.kt
@@ -92,6 +92,9 @@ class AppPreferences(context: Context) {
     val videoPlayerType: String
         get() = sharedPreferences.getString(Constants.PREF_VIDEO_PLAYER_TYPE, VideoPlayerType.WEB_PLAYER)!!
 
+    val exoPlayerStartLandscapeVideoInLandscape: Boolean
+        get() = sharedPreferences.getBoolean(Constants.PREF_EXOPLAYER_START_LANDSCAPE_VIDEO_IN_LANDSCAPE, false)
+
     val exoPlayerAllowSwipeGestures: Boolean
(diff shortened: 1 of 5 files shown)


Maxr1998:
The way it currently works is modelled after other video players like mpv. Jellyfin doesn't show any additional data (description, comments) like YouTube, and thus the focus can be on the video which should be shown as large as possible.

We can, of course, make this behavior configurable, since I noticed other video players (VLC, …) behave differently and this seems more like a personal preference.

eglia:
Feel free to make it configurable, but I'm afraid this exceeds my capabilities in this project.

I find the current behaviour annoying and no other players I use (VLC, YouTube) behaves that way.
I also don't see the need for it, if I want to watch in landscape I can simply turn my phone and the video turns with it.

Maxr1998:
> Feel free to make it configurable, but I'm afraid this exceeds my capabilities in this project.

Sure, I can do that.

> I find the current behaviour annoying and no other players I use (VLC, YouTube) behaves that way.

There doesn't seem to be a common behavior really, as I said, mpv does just that.

> I also don't see the need for it, if I want to watch in landscape I can simply turn my phone and the video turns with it.

Only if auto-rotate is enabled, or by manually pressing the fullscreen button (which is another step). I can see how that can be annoying as well.

Making it configurable really seems to be the best option.

Maxr1998 (approved):

PR #813, opened 2022-09-23
Decouple JavaScript bridge components from Fragments

Maxr1998:


app/src/main/java/org/jellyfin/mobile/MainActivity.kt
```diff
@@ -24,12 +25,14 @@ import org.jellyfin.mobile.utils.extensions.replaceFragment
...
 import org.jellyfin.mobile.webapp.RemotePlayerService
 import org.jellyfin.mobile.webapp.WebViewFragment
+import org.koin.android.ext.android.get
 import org.koin.android.ext.android.inject
 import org.koin.androidx.fragment.android.setupKoinFragmentFactory
 import org.koin.androidx.viewmodel.ext.android.viewModel
 
 class MainActivity : AppCompatActivity() {
-    private val mainViewModel: MainViewModel by viewModel()
+    private val activityEventHandler: ActivityEventHandler = get()
```
nielsvanvelzen, line 34: Why no inject?

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/mobile/MainActivity.kt
Maxr1998: It's not needed. We can resolve this class directly on class instantiation, and since we always need the instance after activity creation it's not optional either.

nielsvanvelzen (dismissed):

Maxr1998:
Merge conflicts fixed :+1:

nielsvanvelzen (approved):

PR #815, opened 2022-09-24
Update GitHub actions workflows

nielsvanvelzen:
**Changes**
- Use Ubuntu 22.04
- Use Java 17
- Use Java Temurin distribution
- Use official Gradle setup action
- Use consistent naming scheme
- Use organization check instead of repository check

**Issues**

### DIFF
--- a/.github/workflows/repo-merge-conflict.yaml
+++ b/.github/workflows/repo-merge-conflict.yaml
@@ -1,4 +1,4 @@
-name: Merge conflict labeler
+name: Repo / Label merge conflict
 
 on:
   push:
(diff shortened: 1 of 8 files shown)


.github/workflows/repo-stale.yaml
```diff
@@ -1,4 +1,4 @@
-name: Issue stale check
+name: Repo Stale
```
Maxr1998, line 1: Not a fan of the "Repo ..." names. Would be prefer something more descriptive like "Stale Issues Check" or "Check for stale issues".

.github/workflows/repo-merge-conflict.yaml
```diff
@@ -1,4 +1,4 @@
-name: Merge conflict labeler
+name: Repo Merge conflict
```
Maxr1998, line 1: Same here, "Label Merge Conflicts" would work imo.

.github/workflows/repo-stale.yaml
Maxr1998: Capitalisation consistency on the titles is a bit weird too. If titles should be uppercase, all words should be.
We could theoretically follow the MusicBrainz English styleguide for titles ^^

nielsvanvelzen: For consistency everything repository related changed to start with "Repo" so we now have App, Repo and Gradle

Maxr1998: For the files that's ok, but the title shown in GitHub is a downgrade imo. It's less easy to understand what's happening.

Maxr1998 (approved):

Issue #696, opened 2022-05-28
Subtitles cause media to be transcoded while device supports subtitles


YodaQ:
### Describe the bug

On my phone (Samsung Node 8) as well as another phone (Samsung A52) when using the ExoPlayer and watching media stored in an mkv container with SSA/ASS subtitles the cause the entire video to transcode even though if you disable subtitles start the video then re-enable the subtitles the subtitles will play without requiring transcoding. This doesnt happen on the web player but the web player has the subtitles being out of sync issue.

### Application version

2.4.4

### Where did you install the app from?

Google Play

### Device information

Samsung Note 8

### Android version

Android 9

### Jellyfin server version

10.7.7

### Which video player implementations does this bug apply to?

- [ ] Web player (default)
- [X] Integrated player (ExoPlayer)
- [ ] External player (VLC, mpv, MX Player)

Maxr1998:
That's actually (mostly) intended, because ExoPlayer doesn't fully support SSA/ASS styling. Thus, we disabled support for them when using ExoPlayer. See google/ExoPlayer#8435 for more details.

It works in your case because we don't refresh the device profile when switching subs within the player. For the decision to direct play/transcode, only the initial configuration when starting playback counts at the moment.

PR #826, opened 2022-09-30
Add ability to direct play ass with exoplayer

Nalsai:
**Changes**
Add a check box in the client settings for ExoPlayer to enable direct play SSA/ASS subtitles without transcoding.

ExoPlayer only partially supports ass subtitles. While support for more features of ass is being worked on it currently only displays it as basic text. Because of that the Jellyfin Android app used to always transcode those subtitles when selected before starting playback. However some people may only have files with basic text in ass subtitles, not care about all the fancy styling or need to avoid transcoding because their server isn't that powerful. Therefore this PR adds the option to direct play ass subtitles.

**Issues**

https://github.com/jellyfin/jellyfin-android/issues/696
https://github.com/jellyfin/jellyfin/issues/7809

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/player/deviceprofile/DeviceProfileBuilder.kt
+++ b/app/src/main/java/org/jellyfin/mobile/player/deviceprofile/DeviceProfileBuilder.kt
@@ -1,6 +1,7 @@
 package org.jellyfin.mobile.player.deviceprofile
 
 import android.media.MediaCodecList
+import org.jellyfin.mobile.app.AppPreferences
 import org.jellyfin.mobile.bridge.ExternalPlayer
 import org.jellyfin.mobile.utils.Constants
 import org.jellyfin.sdk.model.api.CodecProfile
@@ -13,8 +14,11 @@ import org.jellyfin.sdk.model.api.SubtitleDeliveryMethod
 import org.jellyfin.sdk.model.api.SubtitleProfile
 import org.jellyfin.sdk.model.api.TranscodeSeekInfo
 import org.jellyfin.sdk.model.api.TranscodingProfile
+import org.koin.core.component.KoinComponent
+import org.koin.core.component.inject
 
-class DeviceProfileBuilder {
+class DeviceProfileBuilder : KoinComponent {
+    private val appPreferences: AppPreferences by inject()
     private val supportedVideoCodecs: Array<Array<String>>
     private val supportedAudioCodecs: Array<Array<String>>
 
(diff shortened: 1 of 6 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -87,6 +87,8 @@
     <string name="pref_exoplayer_remember_brightness">Remember display brightness</string>
     <string name="pref_exoplayer_allow_background_audio">Background audio</string>
     <string name="pref_exoplayer_allow_background_audio_summary">Allow playing videos in the background with audio-only</string>
+    <string name="pref_exoplayer_direct_play_ass">Don\'t transcode SSA/ASS subtitles</string>
```
Maxr1998, line 90:
```suggestion
    <string name="pref_exoplayer_direct_play_ass">Allow SSA/ASS subtitles in direct play</string>
```

app/src/main/java/org/jellyfin/mobile/player/deviceprofile/DeviceProfileBuilder.kt
```diff
@@ -13,8 +14,11 @@ import org.jellyfin.sdk.model.api.SubtitleDeliveryMethod
 import org.jellyfin.sdk.model.api.SubtitleProfile
 import org.jellyfin.sdk.model.api.TranscodeSeekInfo
 import org.jellyfin.sdk.model.api.TranscodingProfile
+import org.koin.core.component.KoinComponent
+import org.koin.core.component.inject
 
-class DeviceProfileBuilder {
+class DeviceProfileBuilder : KoinComponent {
+    private val appPreferences: AppPreferences by inject()
```
Maxr1998, line 21: You should be able to use constructor injection here. On top of my suggestion, you need to add a `get()` parameter in the AppModule.

```suggestion
class DeviceProfileBuilder(
    private val appPreferences: AppPreferences,
) {
```

app/src/main/java/org/jellyfin/mobile/player/deviceprofile/DeviceProfileBuilder.kt
```diff
@@ -155,13 +159,17 @@ class DeviceProfileBuilder {
             }
         }
 
+        var subtitleProfiles = if (appPreferences.exoPlayerDirectPlayAss)
+            getSubtitleProfiles(EXO_EMBEDDED_SUBTITLES_ASS, EXO_EXTERNAL_SUBTITLES_ASS)
+        else getSubtitleProfiles(EXO_EMBEDDED_SUBTITLES, EXO_EXTERNAL_SUBTITLES);
```
Maxr1998, line 164: Can be a `val` (it's immutable) and should probably be a when statement for better readability:
```suggestion
        val subtitleProfiles = when {
            appPreferences.exoPlayerDirectPlayAss -> getSubtitleProfiles(EXO_EMBEDDED_SUBTITLES_ASS, EXO_EXTERNAL_SUBTITLES_ASS)
            else -> getSubtitleProfiles(EXO_EMBEDDED_SUBTITLES, EXO_EXTERNAL_SUBTITLES)
        }
```

*Also I truly like when statements :sweat_smile:*

app/src/main/java/org/jellyfin/mobile/player/deviceprofile/DeviceProfileBuilder.kt
```diff
@@ -299,7 +307,9 @@ class DeviceProfileBuilder {
         private val FORCED_AUDIO_CODECS = arrayOf(*PCM_CODECS, "alac", "aac", "ac3", "eac3", "dts", "mlp", "truehd")
 
         private val EXO_EMBEDDED_SUBTITLES = arrayOf("srt", "subrip", "ttml")
+        private val EXO_EMBEDDED_SUBTITLES_ASS = arrayOf("srt", "subrip", "ttml", "ssa", "ass")
```
Maxr1998, line 310: Alternatively, you could do just one constant that only contains the ssa/ass formats (`private val SUBTITLES_SSA = arrayOf("ssa", "ass")`), and append the array to the other one on callsite, like `EXO_EMBEDDED_SUBTITLES + SSA_SUBTITLES`.

app/src/main/res/values/strings.xml
```diff
@@ -87,6 +87,8 @@
...
+    <string name="pref_exoplayer_direct_play_ass_summary">Play subtitles directly with basic styling to prevent transcoding</string>
```
Maxr1998, line 91:
```suggestion
    <string name="pref_exoplayer_direct_play_ass_summary">Prevent transcoding and show subtitles with basic styling only. Advanced subtitle styling will not be available if enabled.</string>
```

Maxr1998 (changes requested):
Thanks for the PR! That's definitely a good idea.
Have some comments, but otherwise it looks good. Clean and simple solution :+1:

Maxr1998 (approved):
Thanks for your contribution!

Asgard70:
Bonjour je vois que ç'a été fusionner mais il y a toujours du transcode avec les sous titre ass

PR #842, opened 2022-10-16
Implement downloadFiles in NativeShell

nielsvanvelzen:
Jellyfin 10.9 will add a new function to the NativeShell: downloadFiles. This is the same as downloadFile but can request multiple files at once.

This PR adds a basic implementation for it, internally it splits the request. In the future we can use  a queue or something to download the files more efficiently.

**Changes**
- Implement downloadFiles in NativeShell
- Add NativeInterface.downloadFiles
- Remove NativeInterface.downloadFile

**Issues**

jellyfin-web pull request: https://github.com/jellyfin/jellyfin-web/pull/4016

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/bridge/NativeInterface.kt
+++ b/app/src/main/java/org/jellyfin/mobile/bridge/NativeInterface.kt
@@ -118,21 +118,24 @@ class NativeInterface(private val context: Context) : KoinComponent {
     }
 
     @JavascriptInterface
-    fun downloadFile(args: String): Boolean {
-        val title: String
-        val filename: String
-        val url: String
+    fun downloadFiles(args: String): Boolean {
         try {
-            val options = JSONObject(args)
-            title = options.getString("title")
-            filename = options.getString("filename")
-            url = options.getString("url")
+            val files = JSONArray(args)
+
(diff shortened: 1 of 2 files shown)


nielsvanvelzen:
It'd good to mention that even without this PR everything will keep working. jellyfin-web automatically falls back to using multiple downloadFile calls when downloadFiles is not implemented.

app/src/main/java/org/jellyfin/mobile/bridge/NativeInterface.kt
```diff
@@ -118,21 +118,27 @@ class NativeInterface(private val context: Context) : KoinComponent {
...
-        try {
-            val options = JSONObject(args)
-            title = options.getString("title")
-            filename = options.getString("filename")
-            url = options.getString("url")
-        } catch (e: JSONException) {
-            Timber.e("Download failed: %s", e.message)
-            return false
+    fun downloadFiles(args: String): Boolean {
+        val files = JSONArray(args)
```
Maxr1998, line 122: This should be in the try/catch, instead. Probably for the best to cancel all downloads if one fails anyway.

Maxr1998 (changes requested):

Maxr1998 (approved):

PR #867, opened 2022-10-26
Pin GitHub actions to specific version tag

Maxr1998:
If GitHub actions use a specific version number, Renovate is more likely to supply a changelog when updating actions since the version number is directly linked to a release.

### DIFF
--- a/.github/workflows/app-build.yaml
+++ b/.github/workflows/app-build.yaml
@@ -15,20 +15,20 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - name: Checkout repository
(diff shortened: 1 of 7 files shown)


nielsvanvelzen:
Is there some way to configure Renovate to do this for us? I'm afraid Renovate will revert this change in a PR after this is merged.

Maxr1998:
It doesn't, it should be an explicitly supported use case even. Unfortunately, I cannot find the issue where this was discussed anymore. However, I've been using this in a private work repo for a while, and it's been working as expected.

nielsvanvelzen (approved):

PR #893, opened 2022-11-19
Adds option to change the decoder used by the player

### LINKED ISSUE #760: ExoPlayer: Add SW/HW decoding toggle
**Describe the feature you'd like**

It would be nice if there's a toggle button to switch between SW and HW decoding in the player UI when using ExoPlayer.

**Additional context**

Based on my observation, looks like currently ExoPlayer in Jellyfin Android choose either automatically or HW decoding since there's a video that will lag in my device if it's played using HW decoding (it's playing smoothly if using SW decoding). Other alternative that has been tried and it's problem experienced by me:
- Web Player: didn't have the toggle. Can't direct play for HEVC files.
- MPV: have the toggle ~~but can't play eac3 audio files in my device (the player just flashing and then closed).~~ It can, the problem was another thing. MPV fulfill my requirement, but it's still nice to have in the ExoPlayer since it's more integrated.
- VLC: didn't have the toggle in the player but it's in the app setting. Video files using external subtitle didn't detected in my device.

susonthapa:
**Changes**
This PR adds an option to change between Hardware and Software decoding. When this option is changed the current player is recreated with the new decoder. We are using the `mediaCodecSelector` to control which decoder we want to select for the particular codec. The property `MediaCodecInfo.hardwareAccelerated`  should be taken with a grain of salt. 

**Issues**
Closes #760


susonthapa:
I did a quick test on both the emulator and the real device and I got mixed results. This feature works as expected but the actual device might not support decoding the track with the specified decoder but that is more of an ExoPlayer issue than ours. I was thinking of maybe forcing the decoding even if the decoder doesn't support the capabilities(resolutions, profiles) but this turns out to be the [default behavior](https://github.com/google/ExoPlayer/commit/759cb3231545f3bc4cceae43ed77fdf007b28c5f) of ExpoPlayer. 

Here is a quick demo video on the emulator.

https://user-images.githubusercontent.com/33973551/202887394-deebdefe-1c13-4727-9451-27920ac89c4e.mp4

app/src/main/java/org/jellyfin/mobile/player/PlayerViewModel.kt
```diff
@@ -68,6 +70,14 @@ import org.koin.core.qualifier.named
 import timber.log.Timber
 import java.util.concurrent.atomic.AtomicBoolean
 
+/**
+ * Represents the type of decoder
+ */
+enum class DecoderType {
+    HARDWARE,
+    SOFTWARE,
+}
```
Maxr1998, line 79: Should be moved to another file.

app/src/main/java/org/jellyfin/mobile/player/PlayerViewModel.kt
```diff
@@ -168,6 +182,30 @@ class PlayerViewModel(application: Application) : AndroidViewModel(application),
                 else -> DefaultRenderersFactory.EXTENSION_RENDERER_MODE_ON
             }
             setExtensionRendererMode(rendererMode)
+            // set the media selector
```
Maxr1998, line 185:
```suggestion
```

app/src/main/java/org/jellyfin/mobile/player/PlayerViewModel.kt
```diff
@@ -168,6 +182,30 @@ class PlayerViewModel(application: Application) : AndroidViewModel(application),
...
+            setMediaCodecSelector { mimeType, requiresSecureDecoder, requiresTunnelingDecoder ->
+                val decoderInfoList = MediaCodecSelector.DEFAULT.getDecoderInfos(mimeType, requiresSecureDecoder, requiresTunnelingDecoder)
+                // allow decoder selection only for video
+                if (!MimeTypes.isVideo(mimeType)) {
+                    return@setMediaCodecSelector decoderInfoList
+                }
```
Maxr1998, line 191: We don't need to check this, since this Player is only used for video anyway.
```suggestion
```

app/src/main/java/org/jellyfin/mobile/player/PlayerViewModel.kt
```diff
@@ -168,6 +182,30 @@ class PlayerViewModel(application: Application) : AndroidViewModel(application),
...
+                val filteredDecoderList = when (decoderType.value) {
+                    DecoderType.HARDWARE -> {
+                        // only get the hardware decoder
+                        decoderInfoList.filter { it.hardwareAccelerated }
+                    }
+                    DecoderType.SOFTWARE -> {
+                        decoderInfoList.filter { !it.hardwareAccelerated }
+                    }
```
Maxr1998, line 199:
```suggestion
                    DecoderType.HARDWARE -> decoderInfoList.filter(MediaCodecInfo::hardwareAccelerated)
                    DecoderType.SOFTWARE -> decoderInfoList.filterNot(MediaCodecInfo::hardwareAccelerated)
```

app/src/main/java/org/jellyfin/mobile/player/PlayerViewModel.kt
```diff
@@ -168,6 +182,30 @@ class PlayerViewModel(application: Application) : AndroidViewModel(application),
...
+                    else -> decoderInfoList
+                }
+                // update the decoderType based on the first decoder selected
+                if (filteredDecoderList.isNotEmpty()) {
+                    _decoderType.postValue(if (filteredDecoderList[0].hardwareAccelerated) DecoderType.HARDWARE else DecoderType.SOFTWARE)
+                }
```
Maxr1998, line 205: What will happen if this is empty (after filtering)?

app/src/main/java/org/jellyfin/mobile/player/PlayerViewModel.kt
```diff
@@ -193,6 +231,10 @@ class PlayerViewModel(application: Application) : AndroidViewModel(application),
         _player.value = null
     }
 
+    private fun getAnalyticsCollector() = DefaultAnalyticsCollector(Clock.DEFAULT).apply {
```
Maxr1998, line 234: Probably a better name to ensure it's not confused with a getter method. You can also move this below the init block/above `setupPlayer()`.
```suggestion
    private fun buildAnalyticsCollector() = DefaultAnalyticsCollector(Clock.DEFAULT).apply {
```

app/src/main/java/org/jellyfin/mobile/player/PlayerViewModel.kt
```diff
@@ -225,6 +267,26 @@ class PlayerViewModel(application: Application) : AndroidViewModel(application),
...
+        _decoderType.postValue(type)
+        analyticsCollector.release()
+        analyticsCollector = getAnalyticsCollector()
+        // get the player position so that we can resume it
+        val playedTime = playerOrNull?.currentPosition ?: 0L
+        // soft stop the player
+        playerOrNull?.run {
+            removeListener(this@PlayerViewModel)
+            release()
+        }
```
Maxr1998, line 284:
```suggestion
        val playedTime = playerOrNull?.currentPosition ?: 0L
        // Stop and release the player without ending playback
        playerOrNull?.run {
            removeListener(this@PlayerViewModel)
            release()
        }
        analyticsCollector = getAnalyticsCollector()
```

app/src/main/java/org/jellyfin/mobile/player/source/JellyfinMediaSource.kt
```diff
@@ -108,4 +109,8 @@ class JellyfinMediaSource(
             )
         } else null
     }
+
+    fun updateStartTimeTicks(startTime: Long) {
+        startTimeTicks = startTime * Constants.TICKS_PER_MILLISECOND
+    }
```
Maxr1998, line 115: The name might imply you need to pass ticks – I'd just call this `updateStartTime`, *or* add a custom setter to `startTimeMs` that delegates to `startTimeTicks`.

Maxr1998 (changes requested):
Nice work! I tweaked some small stuff myself already, but still left some other comments for you. Looking forward to merging this once these are addressed.

app/src/main/java/org/jellyfin/mobile/player/PlayerViewModel.kt
susonthapa: Yeah, but the thing is the video will have audio, and subtitle tracks as well. So this function will be called for each of them and I added that to ensure we only change the decoder of the video track.

app/src/main/java/org/jellyfin/mobile/player/PlayerViewModel.kt
susonthapa: In this case, if the player can play the audio then it will only play the audio and not play the video. That was what I got when I tested it.

app/src/main/java/org/jellyfin/mobile/player/PlayerViewModel.kt
Maxr1998: Oh, that makes sense, then I misunderstood this. Thanks for the explanation!

app/src/main/java/org/jellyfin/mobile/player/PlayerViewModel.kt
Maxr1998: As long as it doesn't crash, that should be fine.

susonthapa:
Thanks, @Maxr1998 for taking the time to review the code. I have made the requested changes. Please let me know if I need to change anything.

app/src/main/java/org/jellyfin/mobile/player/PlayerViewModel.kt
Maxr1998: Maybe also remove the comment and curly braces here, like I suggested. I don't think an extra comment is necessary in this case.

app/src/main/java/org/jellyfin/mobile/player/PlayerViewModel.kt
Maxr1998: Thanks, I think it would be better to have it in the `ui` sub-package, though. The `interaction` package is meant for connecting outside components like the web-interface to the player.

app/src/main/java/org/jellyfin/mobile/player/PlayerViewModel.kt
Maxr1998: Could you also please move this still? :sweat_smile:

Maxr1998:
Thanks, I left some more comments in the threads above. The rest is good now. One last request, I usually prefer line comments to start with an uppercase letter, so I'd appreciate if you could fix that as well. I apologize for being a bit nitpicky, I just love consistency :sweat_smile:

susonthapa:
I understand @Maxr1998, I would have done the same if I were you. Since I am very new to this repository I will be making a lot of rookie mistakes so feel free to correct me. It will take me some time to get familiar with the coding style and app architecture.

Maxr1998:
It's ok, I just have relatively specific standards. Some of these style requirements should probably be documented somewhere.

Maxr1998 (approved):
Thanks a lot for your contribution!

PR #901, opened 2022-11-23
Add back button in the integrated player

### LINKED ISSUE #896: Suggestions for the integrated player
I hope the next version of the built-in exo player fix the black background of subtitles when playing videos, add a back button when playing videos, and add a bit rate switch.

This function can effectively improve the built-in exo player's movie-watching experience, perfectly replace the built-in html player, and solve the bugs such as out-of-sync subtitles when using the built-in html player

susonthapa:
**Changes**
I have added the back button functionality to the integrated player. I have used a `Toolbar` to achieve this. I was thinking of just adding a simple button but at this point, I feel like I'm reimplementing the `Toolbar` so I just replaced the `TextView` with the `Toolbar`.

**Issues**
Closes #896 

**Before**
SCR-20221123-e4g

**After**
SCR-20221123-eaj


Maxr1998 (approved):
Thanks for the contribution!
Did some small tweaks myself, thus I can merge this now.

PR #905, opened 2022-11-26
Add link to Android subtitle settings in client settings

Maxr1998:
Related to #896.

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/settings/SettingsFragment.kt
+++ b/app/src/main/java/org/jellyfin/mobile/settings/SettingsFragment.kt
@@ -1,6 +1,8 @@
 package org.jellyfin.mobile.settings
 
+import android.content.Intent
 import android.os.Bundle
+import android.provider.Settings
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/jellyfin/mobile/settings/SettingsFragment.kt
```diff
@@ -164,6 +168,13 @@ class SettingsFragment : Fragment() {
             titleRes = R.string.external_player_app
             enabled = appPreferences.videoPlayerType == VideoPlayerType.EXTERNAL_PLAYER
         }
+        pref(Constants.PREF_SUBTITLE_STYLE) {
+            titleRes = R.string.pref_subtitle_style
+            summaryRes = R.string.pref_subtitle_style_summary
+            defaultOnClick {
+                startActivity(Intent(Settings.ACTION_CAPTIONING_SETTINGS))
```
nielsvanvelzen, line 175: Do we need to check if the system exposes these settings. IIRC in some Android flavors the subtitle settings are not accessible.

> In some cases, a matching Activity may not exist, so ensure you safeguard against this.
> - https://developer.android.com/reference/android/provider/Settings#ACTION_CAPTIONING_SETTINGS

Maxr1998: You're right, a try catch + Toast would be a good idea in this case.

nielsvanvelzen: Maybe just check if the intent is available and hide/disable the option

nielsvanvelzen (approved):

Issue #900, opened 2022-11-22
Don't try connecting to the server automaticaly on app startup


CodeCubeNeo:
Hello! I think it would be useful, that before trying to connect to the server on app startup, the user can change the server route, meaning a user has to click the connect button. That is useful in cases where the server route has changed or the server isnt there at all, because now, when you start the app, there is just a black screen and you have to turn off the wifi in order to change the server route. Also maybe add a cancel button for connecting to the server if you misspelled the server location.

susonthapa:
I don't understand your use case, can you please elaborate? In my case, if the app can't connect to the server I get this screen which is what I expect. 

<img width="385" alt="image" src="https://user-images.githubusercontent.com/33973551/203685155-d79449e4-1ada-49b3-ab81-8b4d27cfd2c7.png">

CodeCubeNeo:
Yes, but you have to wait like 5 seconds in order to change the server location

susonthapa:
Can you please send a screen recording of that? I am having a hard time figuring it out. I have Jellyfin running locally and I'm not able to replicate this.

CodeCubeNeo:
Sure I can, sorry for a but of unimaginable explanation. Hope this helps. I dont have a server running and it is trying to connect (ig the timeout is 5 seconds) so I have to wait (aka black screen) for the timeout time in order to change the server location.

https://user-images.githubusercontent.com/61159092/203852435-e0c8a47f-c85b-4455-91e8-ed7ac092efc0.mp4

susonthapa:
@CodeCubeNeo I got it. I think we need a splash screen as well. I will add the splash screen and a loading indicator. I think adding a cancel button beneath the loading indicator is possible but I don't know if it is worth it. For most people, I think they rarely change the server.

Maxr1998:
We should already have a splash screen on the current version, not sure why it doesn't show though. We could add a cancel button there I think, maybe with a smaller timeout.

susonthapa:
I looked into the code and didn't see a splash screen. We might have a different understanding of when a splash screen is. My understanding of the splash screen is from [here](https://medium.com/swlh/splash-screen-in-android-8ab250e40190). I am working on adding a loading indicator. I will see if I can find the issue with the splash screen.

CodeCubeNeo:
[...]

PR #910, opened 2022-11-27
Added loading indicator when the app is connecting to the server

susonthapa:
**Changes**
I have added a loading indicator that will show when the app is connecting to the server so that users aren't left with a black screen. I have used `ContentLoadingProgressBar` which will try to ensure that the loading process is flicker-free. 

**Issues**
Relates #900 

**Demo**
When it takes 5 seconds to connect to the server the progress indicator is visible. 

https://user-images.githubusercontent.com/33973551/204138567-2178eaf3-2d7b-42de-b9d5-baf8e7c0d80d.mp4

When the connection is quick then the progress indicator won't be visible to reduce flicker.

https://user-images.githubusercontent.com/33973551/204138592-6b350727-a136-448d-ba67-113b53234da7.mp4

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/MainActivity.kt
+++ b/app/src/main/java/org/jellyfin/mobile/MainActivity.kt
@@ -13,6 +13,7 @@ import android.widget.Toast
 import androidx.appcompat.app.AlertDialog
 import androidx.appcompat.app.AppCompatActivity
 import androidx.lifecycle.lifecycleScope
+import org.jellyfin.mobile.databinding.ActivityMainBinding
 import org.jellyfin.mobile.events.ActivityEventHandler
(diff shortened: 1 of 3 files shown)


Maxr1998:
We already have a loading indicator in the WebViewFragment, so this shouldn't be necessary. Currently, the server state is only `Pending` while the server data is loaded from the database, which should be nearly instantaneous.

susonthapa:
I agree, I didn't know about that. I saw a `TODO` in the code and implemented the loader. Since we have implemented the Splash screen I don't think this is necessary so I will close this PR.

Maxr1998:
Yeah, the TODO is quite misleading. I need to ensure if I can just remove it and then do that to prevent any confusion in the future.

PR #911, opened 2022-11-27
fixed potential crash when user reopens the app quickly

susonthapa:
**Changes**
I actually ran into this issue when I reopened the app quickly. The app will crash when it tries to release the `MediaRouter` callbacks. 
SCR-20221127-r8v


Maxr1998 (approved):
Did a few more fixes for issues that could cause similar crashes. Eventually, I'll have to rewrite this code from scratch, Java's null safety is really terrible…

PR #915, opened 2022-12-03
Adds splash screen

susonthapa:
**Changes**
This PR adds a splash screen to the app. This is useful when the app takes time to start(during a cold start or if the user's device is slow). Android 12 automatically uses the app's icon as the window background so that users are not left with a blank screen but still it is a good idea to add a splash screen. 

**Before**
Here is a simulated situation when the app takes more than 5 seconds to initialize.

API - 22

https://user-images.githubusercontent.com/33973551/205443392-ea8dfa47-4ddc-41f5-9c32-0710ca4f5ca2.mp4

**After**
API - 22

https://user-images.githubusercontent.com/33973551/205443473-ac92e67f-d75f-4310-8b88-4081f2922b4b.mp4

For Android 12 and higher, there is no difference but the splash screen background will respect the device's dark mode settings. 

API - 32
Dark Mode

https://user-images.githubusercontent.com/33973551/205443556-2a9ab459-aac0-41ae-94f1-274b776b0683.mp4

Light Mode

https://user-images.githubusercontent.com/33973551/205443571-a4b15a06-4878-4c8e-81c8-7983b860bf8f.mp4

**Issues**
Relates #900


Maxr1998:
I'm not partial to the white backgrounds in light theme when the rest of the app is always in dark mode, and the pixelated launcher icon on lower API levels isn't ideal either…
Both can be easily fixed, I just need to generate some bitmap drawables because vector icons don't work for the splash screen on older Android versions apparently…

Maxr1998 (approved):

Maxr1998:
Ended up doing it slightly differently, Android N+ will see the normal icon with gradient, while versions below will have a solid white vector logo. Thus, it's still very sharp, and we don't have to ship a ton of bitmaps for ancient Android versions.

PR #930, opened 2022-12-21
Don't ignore detekt failures anymore

Maxr1998:


nielsvanvelzen (changes requested):
This option is used for the GH workflow, without it the report is not written.

Maxr1998:
Looking at the logs, the report is still written, so it should be fine. I also split the lint and detekt jobs so that they don't influence each other.

nielsvanvelzen:
I think that's because there's only warnings right now and no errors.

nielsvanvelzen (approved):

PR #931, opened 2022-12-21
Fix and rewrite audio and subtitle track selection when transcoding

Maxr1998:


app/src/main/java/org/jellyfin/mobile/player/ui/PlayerFragment.kt
```diff
@@ -262,30 +264,34 @@ class PlayerFragment : Fragment() {
...
 
     /**
-     * @return true if the audio track was changed
+     * @param callback called if track selection was successful and UI needs to be updated
      */
-    fun onAudioTrackSelected(index: Int): Boolean {
-        return viewModel.selectAudioTrack(index)
+    fun onAudioTrackSelected(index: Int, callback: TrackSelectionCallback): Job = lifecycleScope.launch {
+        if (viewModel.selectAudioTrack(index)) {
+            callback.onTrackSelected(true)
```
nielsvanvelzen, line 271: Why not make this suspended instead of using a callback?

nielsvanvelzen: Same question for the other methods

app/src/main/java/org/jellyfin/mobile/player/source/MediaSourceResolver.kt
```diff
@@ -33,6 +33,10 @@ class MediaSourceResolver(private val apiClient: ApiClient) {
                 itemId = itemId,
                 data = PlaybackInfoDto(
                     userId = apiClient.userId,
+                    // We need to remove the dashes so that the server can find the correct media source.
+                    // And if we didn't pass the mediaSourceId, our stream indices would silently get ignored.
+                    // https://github.com/jellyfin/jellyfin/blob/9a35fd673203cfaf0098138b2768750f4818b3ab/Jellyfin.Api/Helpers/MediaInfoHelper.cs#L196-L201
+                    mediaSourceId = itemId.toString().replace("-", ""),
```
nielsvanvelzen, line 39: This conversion feels hacky, do we not have a reference to the mediasource so we can use the id from that?

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/mobile/player/source/MediaSourceResolver.kt
Maxr1998: It is hacky but sadly not. The id shouldn't be a string anyways but the api model would require changes for this.

app/src/main/java/org/jellyfin/mobile/player/ui/PlayerFragment.kt
Maxr1998: I didn't want to add coroutines to the UI part

app/src/main/java/org/jellyfin/mobile/player/source/MediaSourceResolver.kt
nielsvanvelzen: @crobibero can you help us out here, would it be possible to 1) change the parameter type to a UUID, or 2) allow a UUID with dashes to be auto-converted serverside?

app/src/main/java/org/jellyfin/mobile/player/ui/PlayerFragment.kt
nielsvanvelzen: We can use the viewmodel lifecycle so I don't really see the issue.

Maxr1998: Need to check if PlayerMenus has that. But it's an option, true.

app/src/main/java/org/jellyfin/mobile/player/source/MediaSourceResolver.kt
crobibero: Which parameter do you want changed? `MediaSourceInfo`?

Maxr1998: The `mediaSourceId`, but I don't know whether it supports more than UUIDs.

nielsvanvelzen (approved):

PR #933, opened 2022-12-24
Add button to cancel loading before timeout

### LINKED ISSUE #900: Don't try connecting to the server automaticaly on app startup

Maxr1998:
Closes #900.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -21,7 +21,7 @@
     <string name="connect_button_text">Connect</string>
     <string name="choose_server_button_text">Choose server</string>
     <string name="available_servers_title">Available servers</string>
-    <string name="button_cancel_loading">Cancel loading</string>
+    <string name="button_use_different_server">Use a different server</string>
 
     <string name="battery_optimizations_message">Please disable battery optimizations for media playback while the screen is off.</string>
     <string name="network_title">Allowed Network Types</string>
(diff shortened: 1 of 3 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -21,6 +21,7 @@
     <string name="connect_button_text">Connect</string>
     <string name="choose_server_button_text">Choose server</string>
     <string name="available_servers_title">Available servers</string>
+    <string name="button_cancel_loading">Cancel loading</string>
```
nielsvanvelzen, line 24: Maybe something like "Use a different server" or "Connect to another server" would work better, and at that point maybe even add some more info like the name/address of the server we're currently connecting to.

Maxr1998: I think showing more details would make this a bit bloated. It also only shows after a second of loading already, so I don't think it should be too verbose.

Maxr1998: Fixed in 36a7cafae5c02315bc3e047c50c61a7a9b900a5e.

nielsvanvelzen (approved):

PR #936, opened 2022-12-28
Rework fullscreen/insets handling and fix deprecations

Maxr1998:


app/src/main/java/org/jellyfin/mobile/webapp/WebViewFragment.kt
```diff
@@ -14,6 +14,7 @@ import android.widget.Toast
 import androidx.activity.addCallback
 import androidx.appcompat.app.AlertDialog
 import androidx.core.view.ViewCompat
+import androidx.core.view.WindowCompat
```
nielsvanvelzen, line 17: Is this import used?

app/src/main/java/org/jellyfin/mobile/player/ui/PlayerFullscreenHelper.kt
```diff
@@ -0,0 +1,21 @@
+package org.jellyfin.mobile.player.ui
+
+import android.view.Window
+import androidx.core.view.WindowCompat
+import androidx.core.view.WindowInsetsCompat
+import androidx.core.view.WindowInsetsControllerCompat
+
+class PlayerFullscreenHelper(private val window: Window) {
```
nielsvanvelzen, line 8: Would it make sense to use extension methods on `Window` instead?

Maxr1998: I want to cache the reference to the controller in the helper.

app/src/main/java/org/jellyfin/mobile/webapp/WebViewFragment.kt
Maxr1998: No, no idea how it went there..

nielsvanvelzen (approved):

nielsvanvelzen (approved):

Issue #970, opened 2023-02-09
Discouraging external player from playing some videos


Cnotech:
### Describe the bug

When using external players(MX Player Pro for me), some videos can't be played by clicking the "play" icon in Jellyfin app.It toasts "Media cannot be played by this device".

![Snipaste_2023-02-09_23-02-01](https://user-images.githubusercontent.com/53827731/217849685-add42ec7-b512-4a49-990e-41faaff993b9.png)

But if you "Copy stream URL" and use MX Player to play, MX Player actually could play that normally.

Even if using the "Web player" or "Integrated player" that works fine!

After checking the source code I found it was caused by the `DeviceProfileBuilder.getExternalPlayerProfile()` created a `DeviceProfile` declares far few play profiles decoders than MX Player (or VLC) has.

![Snipaste_2023-02-09_22-23-15](https://user-images.githubusercontent.com/53827731/217849987-99086fa6-cbc9-492e-a4a8-44e2e4b66357.png)

![Snipaste_2023-02-09_22-19-33](https://user-images.githubusercontent.com/53827731/217849991-011668a2-35cb-4fc2-90dc-307a75128378.png)

, so when request `mediaInfoApi.getPostedPlaybackInfo()` with that play profile, server returns 

```
    ...
    "SupportsTranscoding": false,
    "SupportsDirectStream": false,
    "SupportsDirectPlay": false,
    ...
```

,then `JellyfinMediaSource` failed to find a valid play method and throw `"No play method found for $name ($itemId)"`

The temporary solution is to change the generator for `externalPlayerProfile` in class `ExternalPlayer`, use `deviceProfileBuilder.getDeviceProfile()` since external players could play videos supported by the device,I would put a new pull request later.But a better solution is to let `JellyfinMediaSource` know whether `appPreferences.videoPlayerType == VideoPlayerType.EXTERNAL_PLAYER` and let it go when "No play method found",since external players are generally strong,like this :
![Snipaste_2023-02-09_23-11-14](https://user-images.githubusercontent.com/53827731/217852570-d06e1424-3ec1-42ff-844f-a8ec89611fa5.png)

I'm not familiar with Kotlin so I didn't do so.Looking forward to your suggestion!

### Logs

```shell
E/MediaSourceResolver: Cannot create JellyfinMediaSource
    java.lang.IllegalArgumentException: No play method found for VIDEO_TITLE_HERE (ITEM_ID_HERE)
[...]

PR #971, opened 2023-02-09
Fix discouraging external player from playing some videos

Cnotech:
**Changes**
Change the generator for `externalPlayerProfile` in class `ExternalPlayer`, use `deviceProfileBuilder.getDeviceProfile()` instead

**Issues**
#970


nielsvanvelzen (changes requested):
Issues with the external player profile should be fixed in the external player profile. Swapping it with the internal profile is not a fix.

Issue #994, opened 2023-03-20
Allow exoplayer swipe gesture exclusion zone to be configurable


lucacataldo:
**Describe the feature you'd like**
Allow the user to configure the existing exoplayer gesture exclusion zone. This would solve issues with swiping home to enable PIP.

**Additional context**
Many Android versions now rely on swipe gestures for home / recent apps - the former of which often triggers picture-in-picture. The default vertical exclusion value of 64 can cause issues depending on the Android version and/or skin, but changing the hardcoded default could cause unwanted effects for users who aren't affected.

lucacataldo:
Decided to give this a shot myself in #995 😄

lucacataldo:
Hey @nielsvanvelzen sorry I sprung the unsolicited PR on you, but is this something that the Jellyfin team would be interested in adding to the project?

I believe it's a helpful quality-of-life upgrade with little to no downside. I'm open to discussion as well 😃

Maxr1998:
Hi @lucacataldo, sorry for the lack of response. The recent weeks have been very busy for me.

Preferably the gesture exclusion zones would be defined through the insets provided by the OS, if that's not sufficient making them configurable would certainly be an option. I'll try to look at it soon!

lucacataldo:
@Maxr1998 Not a problem whatsoever! That would be an awesome solution as well.

I'd be happy to help with testing on a Pixel 7 if you're able to get that implemented since that's the device I initially had the issue with. Feel free to reach out if there's any way I can help!

PR #995, opened 2023-03-20
Add swipe gesture exclusion size preference

lucacataldo:
**Changes**
Adds preference to modify the existing `SWIPE_GESTURE_EXCLUSION_SIZE_VERTICAL` constant, allowing the user to configure the size of the vertical exclusion zone when using ExoPlayer. Also implemented the necessary conditionals to enable / disable this setting based on other relevant settings. Retains the current default value of `64`, allows for a range between 0-128.

_NOTE:_ `SWIPE_GESTURE_EXCLUSION_SIZE_VERTICAL` was changed to `PREF_EXOPLAYER_SWIPE_GESTURE_EXCLUSION_SIZE_VERTICAL` to better conform with existing conventions - let me know if a different name is desired.

**Issues**

Implements #994

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/app/AppPreferences.kt
+++ b/app/src/main/java/org/jellyfin/mobile/app/AppPreferences.kt
@@ -91,6 +91,9 @@ class AppPreferences(context: Context) {
     val exoPlayerAllowSwipeGestures: Boolean
         get() = sharedPreferences.getBoolean(Constants.PREF_EXOPLAYER_ALLOW_SWIPE_GESTURES, true)
 
+    val exoPlayerSwipeExclusionSizeVertical: Int
+        get() = sharedPreferences.getInt(Constants.PREF_EXOPLAYER_SWIPE_GESTURE_EXCLUSION_SIZE_VERTICAL, 64)
+
     val exoPlayerRememberBrightness: Boolean
         get() = sharedPreferences.getBoolean(Constants.PREF_EXOPLAYER_REMEMBER_BRIGHTNESS, false)
 
--- a/app/src/main/java/org/jellyfin/mobile/settings/SettingsFragment.kt
+++ b/app/src/main/java/org/jellyfin/mobile/settings/SettingsFragment.kt
@@ -18,6 +18,7 @@ import de.Maxr1998.modernpreferences.helpers.defaultOnSelectionChange
 import de.Maxr1998.modernpreferences.helpers.pref
 import de.Maxr1998.modernpreferences.helpers.screen
 import de.Maxr1998.modernpreferences.helpers.singleChoice
+import de.Maxr1998.modernpreferences.helpers.seekBar
 import de.Maxr1998.modernpreferences.preferences.CheckBoxPreference
 import de.Maxr1998.modernpreferences.preferences.choice.SelectionItem
 import org.jellyfin.mobile.R
@@ -37,6 +38,7 @@ class SettingsFragment : Fragment() {
     private val settingsAdapter: PreferencesAdapter by lazy { PreferencesAdapter(buildSettingsScreen()) }
     private lateinit var startLandscapeVideoInLandscapePreference: CheckBoxPreference
     private lateinit var swipeGesturesPreference: CheckBoxPreference
+    private lateinit var swipeExclusionZoneVertical: Preference
     private lateinit var rememberBrightnessPreference: Preference
     private lateinit var backgroundAudioPreference: Preference
     private lateinit var directPlayAssPreference: Preference
@@ -97,6 +99,7 @@ class SettingsFragment : Fragment() {
             defaultOnSelectionChange { selection ->
                 startLandscapeVideoInLandscapePreference.enabled = selection == VideoPlayerType.EXO_PLAYER
                 swipeGesturesPreference.enabled = selection == VideoPlayerType.EXO_PLAYER
+                swipeExclusionZoneVertical.enabled = selection == VideoPlayerType.EXO_PLAYER
                 rememberBrightnessPreference.enabled = selection == VideoPlayerType.EXO_PLAYER && swipeGesturesPreference.checked
                 backgroundAudioPreference.enabled = selection == VideoPlayerType.EXO_PLAYER
                 directPlayAssPreference.enabled = selection == VideoPlayerType.EXO_PLAYER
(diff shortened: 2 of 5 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -90,6 +90,7 @@
     <string name="pref_exoplayer_start_landscape_video_in_landscape">Start landscape mode videos in landscape orientation</string>
     <string name="pref_exoplayer_allow_brightness_volume_gesture">Brightness and volume gestures</string>
     <string name="pref_exoplayer_remember_brightness">Remember display brightness</string>
+    <string name="pref_exoplayer_swipe_exclusion_size_vertical">Gesture dead zone</string>
```
lucacataldo, line 93: Not sure what the desired label would be here

app/src/main/java/org/jellyfin/mobile/app/AppPreferences.kt
```diff
@@ -91,6 +91,9 @@ class AppPreferences(context: Context) {
     val exoPlayerAllowSwipeGestures: Boolean
         get() = sharedPreferences.getBoolean(Constants.PREF_EXOPLAYER_ALLOW_SWIPE_GESTURES, true)
 
+    val exoPlayerSwipeExclusionSizeVertical: Int
+        get() = sharedPreferences.getInt(Constants.PREF_EXOPLAYER_SWIPE_GESTURE_EXCLUSION_SIZE_VERTICAL, 64)
```
Maxr1998, line 95:
```suggestion
        get() = sharedPreferences.getInt(Constants.PREF_EXOPLAYER_SWIPE_GESTURE_EXCLUSION_SIZE_VERTICAL, 64)
```

I think you could still use the previous constant here.

app/src/main/java/org/jellyfin/mobile/utils/Constants.kt
```diff
@@ -35,6 +35,7 @@ object Constants {
     const val PREF_VIDEO_PLAYER_TYPE = "pref_video_player_type"
     const val PREF_EXOPLAYER_START_LANDSCAPE_VIDEO_IN_LANDSCAPE = "pref_exoplayer_start_landscape_video_in_landscape"
     const val PREF_EXOPLAYER_ALLOW_SWIPE_GESTURES = "pref_exoplayer_allow_swipe_gestures"
+    const val PREF_EXOPLAYER_SWIPE_GESTURE_EXCLUSION_SIZE_VERTICAL = "pref_exoplayer_swipe_exclusion_size_vertical"
```
Maxr1998, line 38: Idk why I originally named it like that, but this is better:
```suggestion
    const val PREF_EXOPLAYER_SWIPE_GESTURE_EXCLUSION_ZONE_HEIGHT = "pref_exoplayer_swipe_exclusion_zone_height"
```

Please rename the constant/preference/string name in other places as well.

app/src/main/java/org/jellyfin/mobile/utils/Constants.kt
```diff
@@ -101,7 +102,6 @@ object Constants {
     const val TICKS_PER_MILLISECOND = 10000
     const val PLAYER_TIME_UPDATE_RATE = 10000L
     const val DEFAULT_CONTROLS_TIMEOUT_MS = 2500
-    const val SWIPE_GESTURE_EXCLUSION_SIZE_VERTICAL = 64
```
Maxr1998, line 104: Should be kept and used as a default for the preference, see above.

app/src/main/res/values/strings.xml
Maxr1998: Seems fine as is.

Maxr1998 (commented):
I realized that my suggestion with insets doesn't really work (the gesture insets are way smaller than what we have now, even), so I prefer your solution. A few small things to fix, then it's good to go.

app/src/main/java/org/jellyfin/mobile/settings/SettingsFragment.kt
```diff
@@ -113,8 +116,18 @@ class SettingsFragment : Fragment() {
...
                 rememberBrightnessPreference.enabled = checked
+                swipeExclusionZoneVertical.enabled = checked
             }
         }
+        swipeExclusionZoneVertical = seekBar(Constants.PREF_EXOPLAYER_SWIPE_GESTURE_EXCLUSION_SIZE_VERTICAL){
+            titleRes = R.string.pref_exoplayer_swipe_exclusion_size_vertical
+            min = 0
+            default = 64
+            max = 128
+            step = 8
```
Maxr1998, line 127: Preferably use constants here as well.

app/src/main/java/org/jellyfin/mobile/settings/SettingsFragment.kt
```diff
@@ -113,8 +116,18 @@ class SettingsFragment : Fragment() {
...
+            enabled = appPreferences.videoPlayerType == VideoPlayerType.EXO_PLAYER && appPreferences.exoPlayerAllowSwipeGestures
+        }
+
```
Maxr1998, line 130:
```suggestion
```

app/src/main/java/org/jellyfin/mobile/settings/SettingsFragment.kt
```diff
@@ -97,6 +99,7 @@ class SettingsFragment : Fragment() {
             defaultOnSelectionChange { selection ->
                 startLandscapeVideoInLandscapePreference.enabled = selection == VideoPlayerType.EXO_PLAYER
                 swipeGesturesPreference.enabled = selection == VideoPlayerType.EXO_PLAYER
+                swipeExclusionZoneVertical.enabled = selection == VideoPlayerType.EXO_PLAYER
```
Maxr1998, line 102:
```suggestion
                swipeExclusionZoneVertical.enabled = selection == VideoPlayerType.EXO_PLAYER && swipeGesturesPreference.checked
```

lucacataldo:
> @lucacataldo are you still interested in finishing this PR?

@nielsvanvelzen My apologies, I ended up losing track of this PR and unfortunately no longer have the time availability to get it up-to-snuff. Anecdotally, I haven't faced the issue that prompted this change in quite some time.

Let me know if you'd like me to close it or feel free to do so yourself.

nielsvanvelzen:
I'll close it now, still thanks for your contribution though! If anyone else thinks this PR still makes sense for them and wants to pick it up, please do so by opening a new PR (but do address the review comments 😉)

PR #1011, opened 2023-04-03
Add default changelog to fastlane

nielsvanvelzen:
**Changes**
- Add default changelog to fastlane
  It's copied from the ATV repo (https://github.com/jellyfin/jellyfin-androidtv/blob/master/fastlane/metadata/android/en-US/changelogs/default.txt) with a small change for the name.
  The changelog is shown by F-Droid like here https://f-droid.org/en/packages/org.jellyfin.androidtv/

**Issues**

### DIFF
--- a/fastlane/metadata/android/en-US/changelogs/default.txt
+++ b/fastlane/metadata/android/en-US/changelogs/default.txt
@@ -1,0 +1,3 @@
+Thank you for using Jellyfin for Android! Using the latest released Jellyfin Server is always recommended.
+
+For more information, please see our blog at jellyfin.org, or read the full changelog on GitHub.


nielsvanvelzen:
Sneaking it into the upcoming release :-)

Maxr1998 (approved):

fastlane/metadata/android/en-US/changelogs/default.txt
```diff
@@ -0,0 +1,3 @@
+Thank you for using Jellyfin mobile! Using the latest released Jellyfin Server is always recommended.
```
Maxr1998, line 1: Maybe "Jellyfin (on/for/-) Android" would be better here?

nielsvanvelzen: yeah it was either "Jellyfin mobile" or "Jellyfor on Android" because "Jellyfin on Android TV" can be interpreted as the name or as Jellyfin on [platform]. I'll change it.

Maxr1998: Thinking about it, Jellyfin on Android is ambiguous as well. What about "Thanks for using the Jellyfin client for Android"?

nielsvanvelzen: Or maybe just "Thank you for using the Jellyfin app"? No need to indicate it's an Android app on an Android store.

Maxr1998: Imo "on" doesn't really work because it would imply there's one Jellyfin (client) that works on multiple operating systems, but each have their own client actually.

nielsvanvelzen: What about "Thank you for using Jellyfin for Android!", since that's our "official" name on the jellyfin.org website.

nielsvanvelzen: Also just realized that was an option in your initial comment... oops

Maxr1998: Yeah lol, to be fair, the parentheses weren't the most clear.

Maxr1998 (approved):

PR #1012, opened 2023-04-04
Rework track handling and selection

Maxr1998:
Add proper support for external audio streams, rework audio and subtitle selection, and support extracted subtitles during transcoding playback.

Also extracts some code to a new TrackSelectionHelper class.


nielsvanvelzen (approved):
Looking forward to the bugfix PR's 😄

PR #1013, opened 2023-04-04
Disable Jetifier

Maxr1998:
We don't have any dependencies requiring it anymore, so it can be disabled.

If we want to be on the safe side, we could also leave this open until 2.5.0 is released, just doing a PR so that I don't forget.


nielsvanvelzen (approved):

nielsvanvelzen:
Should be fine

PR #1043, opened 2023-05-13
Add pgssub to ExoPlayer subtitle codecs

Maxr1998:
ExoPlayer supports decoding embedded pgssub subtitles, so they should be enabled.

Tested locally and confirmed working.
I also removed the likely invalid `pgs` specification from the external player profile, since ffprobe lists pgs subtitles as `pgssub`.

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/player/deviceprofile/DeviceProfileBuilder.kt
+++ b/app/src/main/java/org/jellyfin/mobile/player/deviceprofile/DeviceProfileBuilder.kt
@@ -319,11 +319,11 @@ class DeviceProfileBuilder(
          */
         private val FORCED_AUDIO_CODECS = arrayOf(*PCM_CODECS, "alac", "aac", "ac3", "eac3", "dts", "mlp", "truehd")
 
-        private val EXO_EMBEDDED_SUBTITLES = arrayOf("srt", "subrip", "ttml")
+        private val EXO_EMBEDDED_SUBTITLES = arrayOf("pgssub", "srt", "subrip", "ttml")
         private val EXO_EXTERNAL_SUBTITLES = arrayOf("srt", "subrip", "ttml", "vtt", "webvtt")
         private val SUBTITLES_SSA = arrayOf("ssa", "ass")
         private val EXTERNAL_PLAYER_SUBTITLES = arrayOf(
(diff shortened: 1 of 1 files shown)


nielsvanvelzen (approved):
Funnily enough I added this to the TV app 3 weeks ago :p

https://github.com/jellyfin/jellyfin-androidtv/pull/2690

Maxr1998:
Nice :smile:
I actually thought it was either already supported in the app or unsupported by ExoPlayer, but I tried to watch a movie recently and only then realized we simply didn't enable it yet.

silentTeee:
So I guess I will place my warning here: I don't think ExoPlayer's implementation is all that solid yet. I've been meaning to capture a video of it happening compared to the web player, but if more than one track is supposed to show up on the screen at the same time, I had issues on the TV app with them not showing up correctly after we switched to native support from transcoding. In about 17 hours I should have free time to properly characterize the TV app, and I can try abd characterize the mobile app as well if we have a test image.

silentTeee:
So I guess I will place my warning here: I don't think ExoPlayer's implementation is all that solid yet. I've been meaning to capture a video of it happening compared to the web player, but if more than one track is supposed to show up on the screen at the same time, I had issues on the TV app with them not showing up correctly after we switched to native support from transcoding. In about 17 hours I should have free time to properly characterize the TV app, and if I still see issues I'll submit a ticket there. I can characterize the mobile app as well if we have a test image.

Maxr1998:
Good point, support for multiple subtitles at the same time doesn't really exist yet with the current track selection code. Many movies have two sets of subtitles to provide subtitles for foreign languages spoken in the movie (also known as forced subtitles) in addition to the closed captions that transcribe all the conversations.
However, that's unrelated to the PGS subtitle format, the same would happen if you had two text-based subtitle tracks (like SRT).

I believe it would make sense to create a separate ticket for this issue.

silentTeee:
I think we _might_ be referring to two different things (my fault, I worded it poorly). I think it's not so much that there are multiple _tracks_ playing at the same time, rather it seems more like there are multiple, parallel captions of text on the screen, at different positions, but part of the same track...at least, in an MKV container it's only labeled as one track. 

Sadly that free time I thought I had is going to be delayed, but I will submit a separate ticket to the TV App repo and this one (assuming the same issue happens) once I have recordings of what it looks like across several clients. 

Otherwise, congrats on getting this and 2.5.0 out the door!

PR #1060, opened 2023-05-16
Use helper class for Android version checks

Maxr1998:


app/src/main/java/org/jellyfin/mobile/utils/AndroidVersion.kt
```diff
@@ -0,0 +1,32 @@
+package org.jellyfin.mobile.utils
+
+import android.os.Build
+
+object AndroidVersion {
```
nielsvanvelzen, line 5: We should probably add a kdoc to each of the vals to specify both the name & API level.

PR #1076, opened 2023-05-22
Disable ExoPlayer and External player while in SyncPlay

Maxr1998:
Because neither player currently supports SyncPlay, detecting a running session and falling back to the web player makes a lot of sense.

### DIFF
--- a/app/src/main/assets/native/ExoPlayerPlugin.js
+++ b/app/src/main/assets/native/ExoPlayerPlugin.js
@@ -48,7 +48,9 @@ export class ExoPlayerPlugin {
     }
 
     canPlayItem(item, playOptions) {
-        return this._nativePlayer.isEnabled() && playOptions.fullscreen;
+        return this._nativePlayer.isEnabled() &&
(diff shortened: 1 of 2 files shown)


nielsvanvelzen (approved):

nielsvanvelzen:
SyncPlay with external players doesn't work anyway. Yes it will start the external player but after that it will never sync it's state to the others in the SyncPlay room or execute the commands it receives. This is the entire reason why this change was made and we will not revert that.

swimminsparrow:
Syncplay with external players was useful infact to "start" a video simultaneously. Now the only solution is to ask each user in the group to search the video in the library and start it manually, making something like a fingers countdown.
The real solution to the problem should be to improve the web player to avoid transcoding on each video, but obviously this is not in the roadmap.
I really think a workaround could be an option in the syncplay preferences. I don't think many users use the web player. The only reason to have jellyfin for android installed was the syncplay working even in this "bad" way and so was preferrable to a native solution as findroid is. 
This is only an opinion. Closed.

PR #1103, opened 2023-06-03
Fix VLC external player crashing on Android 13

### LINKED ISSUE #912: VLC Player Crash
### Describe the bug

I'm on a Galaxy Tab S8 Plus (tablet) running android 13, I've got the jellyfin android app installed with the video player set to external.
The external video player selected is VLC Player. 
When I hit play on any video the app crashes to desktop (home screen) with no error code.
When I relaunch the app it says it crashed and offers to clear cache but doing this does not help. 
I've tried the Web player, Exo player and MPV (external) and all three work. Only VLC player causes this crash.
Have used VLC as external on other devices with no issues, only encountered this after trying it on my new tablet. 

### Logs

_No response_

### Application version

2.4.4

Google Play

### Device information

Galaxy Tab S8 Plus

### Android version

Android 13

### Jellyfin server version

10.8.7

- [ ] Web player (default)

Maxr1998:
**Changes**
Due to security-related changes in Android 13, the entrypoint must be the default start activity.

This requires https://code.videolan.org/videolan/vlc-android/-/merge_requests/1738 to be merged and released for playback tracking to work again.

**Issues**
Fixes #912

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/bridge/ExternalPlayer.kt
+++ b/app/src/main/java/org/jellyfin/mobile/bridge/ExternalPlayer.kt
@@ -334,7 +334,7 @@ class ExternalPlayer(
                 ComponentName(packageName, "$packageName.ActivityScreen")
             }
             ExternalPlayerPackage.VLC_PLAYER -> {
-                ComponentName(packageName, "$packageName.gui.video.VideoPlayerActivity")
+                ComponentName(packageName, "$packageName.StartActivity")
             }
             else -> null
         }


app/src/main/java/org/jellyfin/mobile/bridge/ExternalPlayer.kt
```diff
@@ -334,7 +334,7 @@ class ExternalPlayer(
                 ComponentName(packageName, "$packageName.ActivityScreen")
             }
             ExternalPlayerPackage.VLC_PLAYER -> {
-                ComponentName(packageName, "$packageName.gui.video.VideoPlayerActivity")
+                ComponentName(packageName, "$packageName.StartActivity")
```
nielsvanvelzen, line 337: Do we want to add a version check here so users with an older version of the VLC app don't need to update on Android 12<=?

Maxr1998: I don't think so, it'll still work, just not have progress tracking.
Also I'm not even sure the recent VLC updates didn't break it anyway on all versions.

nielsvanvelzen (approved):

PR #1120, opened 2023-06-27
A convenient solution for implementing exoplayer rate control. Step 2/2

shaino:
Add rate limiting variables to exoplayer.
Step 2/2

Provide a temporary solution before the developer provides exoplayer rate control.

**Changes**
Obtain and initialize the exoplayer bitrate.

**Issues**


nielsvanvelzen:
We're not interested in this kind of code. Feel free to open a pull request with a proper fix (e.g. doesn't use the file system / adheres to our coding standards).

Closing as not interested.

shaino:
Okay, I will try again according to the requirements.

Maxr1998:
I'd advocate for directly taking the data from client settings instead of a stopgap solution that only persists to local config, so that the change is useful across all used devices, however.

PR #1134, opened 2023-07-16
Pre-fill last hostname and add history to suggestions in connect screen

Maxr1998:


app/src/main/java/org/jellyfin/mobile/ui/screens/connect/ServerSelection.kt
```diff
@@ -53,31 +53,60 @@ import androidx.compose.ui.text.input.KeyboardType
...
         connectionHelper.discoverServersAsFlow().collect { serverInfo ->
-            discoveredServers.add(serverInfo)
+            serverSuggestions.add(
+                index = 0,
+                ServerSuggestion(
+                    name = serverInfo.name,
+                    address = serverInfo.address,
+                    lastUsedTimestamp = System.currentTimeMillis(),
+                ),
+            )
```
nielsvanvelzen, line 109: We should check if the suggestion is already the current server or a previously used server to avoid duplicates. I'd also argue we should not set the lastUsedTimestamp to now but to null/0 instead.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/mobile/ui/screens/connect/ServerSelection.kt
Maxr1998: But that implies that every time you click an entry you'd have to remove it from the suggestions and add back the currently filled server to the suggestions if applicable. That's a lot of complexity for something we could just ignore.
Having the server in both the selection and suggestions panel isn't a huge issue anyway. Duplicates within the suggestions should be eliminated, though, agreed.

I set the lastUsedTimestamp to the current time on purpose, since that's what the servers should be sorted by. It's true that it's not necessarily a "usage time" then, but 0/null only makes things worse.

nielsvanvelzen: > But that implies that every time you click an entry you'd have to remove it from the suggestions and add back the currently filled server to the suggestions if applicable. That's a lot of complexity for something we could just ignore.
Having the server in both the selection and suggestions panel isn't a huge issue anyway. Duplicates within the suggestions should be eliminated, though, agreed.

Fair enough

> I set the lastUsedTimestamp to the current time on purpose, since that's what the servers should be sorted by. It's true that it's not necessarily a "usage time" then, but 0/null only makes things worse.

The rename to just "timestamp" works for that 👍

app/src/main/java/org/jellyfin/mobile/ui/screens/connect/ServerSelection.kt
```diff
@@ -53,31 +53,64 @@ import androidx.compose.ui.text.input.KeyboardType
...
+                timestamp = server.lastUsedTimestamp,
+            )
+        }
+    }
+
+    // Server discovery
     LaunchedEffect(Unit) {
         connectionHelper.discoverServersAsFlow().collect { serverInfo ->
-            discoveredServers.add(serverInfo)
+            serverSuggestions.removeIf { existing -> existing.address == serverInfo.address }
```
nielsvanvelzen, line 104: This _might_ not always work due to race conditions if the first discovery is made before loadPreviouslyUsedServers returns. I guess it can be easily fixed by putting both in the same LaunchedEffect so discovery runs only after previous used servers are loaded.

app/src/main/java/org/jellyfin/mobile/ui/screens/connect/ServerSuggestion.kt
```diff
@@ -0,0 +1,18 @@
...
+data class ServerSuggestion(
+    val type: Type,
+    val name: String,
+    val address: String,
+    /**
+     * A timestamp for this suggestion, used for sorting.
+     * For discovered servers, this should be the discovery time,
+     * for saved servers, this should be the last used time.
+     */
+    val timestamp: Long,
```
nielsvanvelzen, line 12: Didn't catch this earlier, I would use an Instant here so it's more typesafe and less prone to issues with formatting/timezones/etc. Although that would likely need changes to the db scheme and for its current usage a long works fine.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/mobile/ui/screens/connect/ServerSelection.kt
Maxr1998: Yeah, I was ware about that race condition, but couldn't think of a good solution. Just using the same LaunchedEffect just seems so.. easy. I should've thought of that 😅

app/src/main/java/org/jellyfin/mobile/ui/screens/connect/ServerSuggestion.kt
Maxr1998: Right, an instant would definitely be better, but at the moment we only use it for sorting and as long as we don't change the database changing the rest doesn't make a lot of sense.

nielsvanvelzen (approved):

PR #1160, opened 2023-08-19
Update compile and target SDK to 34

Maxr1998:


nielsvanvelzen (changes requested):
Changes look fine but since it targets SDK 34 I'd like to wait for that to be stable and then merge.

We could also bump to SDK 33 now and 34 in [a week or two](https://developer.android.com/about/versions/14/overview).

nielsvanvelzen (approved):

PR #1161, opened 2023-08-19
Fix deprecation and extract method

Maxr1998:


app/src/main/java/org/jellyfin/mobile/MainActivity.kt
```diff
@@ -84,25 +87,10 @@ class MainActivity : AppCompatActivity() {
...
-                                    Bundle().apply {
-                                        putParcelable(Constants.FRAGMENT_WEB_VIEW_EXTRA_SERVER, state.server)
-                                    },
-                                )
-                            }
-                        }
-                    }
+        lifecycleScope.launch {
+            repeatOnLifecycle(Lifecycle.State.STARTED) {
+                mainViewModel.serverState.collect { state ->
```
nielsvanvelzen, line 92: This could be simplified with
```
mainViewModel.serverState
	.flowWithLifecycle(lifecycle, Lifecycle.State.STARTED)
	.onEach { state -> /* magic */ }
	.launchIn(lifecycleScope)
```

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/mobile/MainActivity.kt
Maxr1998: This creates a `callbackFlow` under the hood which adds an unnecessary buffer, is thus slightly less efficient and imo also harder to read and understand.

nielsvanvelzen: I find it more readable personally as it decreases indenting, no strong preference for either approach though.

Maxr1998: I don't personally mind the indentation too much, since the code itself was extracted to that new function. I'm mostly not a fan of the `launchIn` at the end seeming kind of sneaky, and if we wrap it inside a `launch` block and use `collect` instead, it's not that much of a difference anymore.

nielsvanvelzen (approved):

PR #1172, opened 2023-08-24
Introduce BackPressInterceptor

timm0e:
This removes usages of the deprecated onBackPress method and instead relies on the new onBackPressedDispatcher while providing compatibility with the way we stack Fragments on top of each other (see BackPressInterceptor.kt for a more detailed explanation).


Maxr1998 (approved):

Maxr1998 (approved):
Thanks for your contribution! Those oddities with the OnBackPressDispatcher API are surely interesting. Thanks for documenting your changes with so much detail!

PR #1173, opened 2023-08-24
Allow to exit from fullscreen with the back button in ExoPlayer

timm0e:
When using ExoPlayer, this allows users to exit the full-screen mode by pressing the back button, similar to e.g. the YouTube app.

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/player/ui/PlayerFragment.kt
+++ b/app/src/main/java/org/jellyfin/mobile/player/ui/PlayerFragment.kt
@@ -36,6 +36,7 @@ import org.jellyfin.mobile.player.PlayerException
 import org.jellyfin.mobile.player.PlayerViewModel
 import org.jellyfin.mobile.player.interaction.PlayOptions
 import org.jellyfin.mobile.utils.AndroidVersion
+import org.jellyfin.mobile.utils.BackPressInterceptor
 import org.jellyfin.mobile.utils.Constants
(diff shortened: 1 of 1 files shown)


Maxr1998 (approved):
Sounds reasonable. Thanks for the contribution!

elnoelsi:
Im not sure if this was the change but now when i use back gesture it goes to portrait mode from landscape mode instead of actually leaving fullscreen @Maxr1998

elnoelsi:
i dont mean to be offensive, but this is pretty useless behavior forcing you to take 2 actions if you want to close the player, no matter what. its actually horrible cause after leaving landscape you have to move your finger to a different edge to be able to back gesture. 

i dont see how it could ever be useful, if i want to close the player i want to do just that. if i want extra options/info i will tap the screen. its pretty standard behavior on every app and player that ive used. 

at least give an option to disable it or have it disabled when "start landscape mode" is enabled. 

i dont think i can emphasize enough how much of a pain it is to have to gesture back twice every time i need to close the player, it doesnt even pause it. if i wanted to hair on portrait mode i would press "exit full screen" icon that is there for a reason.

mdinger:
Is there an option to disable this?

elnoelsi:
There is no option thats the most annoying part of all, i genuinely think you have to be masochist to enjoy this feature in any way

nielsvanvelzen:
Please open an issue so we can discuss this change. If there is a high demand for a revert/preference we may consider it.

mdinger:
@nielsvanvelzen as suggested, please see #1256

PR #1181, opened 2023-08-31
Ensure ExternalPlayer always direct plays

### LINKED ISSUE #970: Discouraging external player from playing some videos

Maxr1998:
Instead of using a device profile like for ExoPlayer, we always build the direct stream URL, and at the same time tweak the media source request so that it never suggests transcoding. We do that by supplying an effectively unlimited maximum bit rate, which was previously hard-coded to 120 Mbps through the device profile.

Fixes #970.

### DIFF
--- a/app/src/main/assets/native/ExternalPlayerPlugin.js
+++ b/app/src/main/assets/native/ExternalPlayerPlugin.js
@@ -141,12 +141,11 @@ export class ExternalPlayerPlugin {
     async getDeviceProfile() {
         return {
             Name: 'Android External Player Stub',
-            MaxStreamingBitrate: 100000000,
-            MaxStaticBitrate: 100000000,
-            MusicStreamingTranscodingBitrate: 320000,
+            MaxStreamingBitrate: 1000000000,
+            MaxStaticBitrate: 1000000000,
             DirectPlayProfiles: [{Type: 'Video'}, {Type: 'Audio'}],
             CodecProfiles: [],
-            SubtitleProfiles: JSON.parse(this._externalPlayer.getSubtitleProfiles()),
+            SubtitleProfiles: [{Method: 'Embed'}, {Method: 'Drop'}],
             TranscodingProfiles: []
         };
     }
--- a/app/src/main/assets/native/ExternalPlayerPlugin.js
+++ b/app/src/main/assets/native/ExternalPlayerPlugin.js
@@ -141,8 +141,8 @@ export class ExternalPlayerPlugin {
     async getDeviceProfile() {
         return {
             Name: 'Android External Player Stub',
-            MaxStreamingBitrate: 1000000000,
-            MaxStaticBitrate: 1000000000,
+            MaxStreamingBitrate: 1_000_000_000,
+            MaxStaticBitrate: 1_000_000_000,
             DirectPlayProfiles: [{Type: 'Video'}, {Type: 'Audio'}],
             CodecProfiles: [],
             SubtitleProfiles: [{Method: 'Embed'}, {Method: 'Drop'}],
--- a/app/src/main/java/org/jellyfin/mobile/bridge/ExternalPlayer.kt
+++ b/app/src/main/java/org/jellyfin/mobile/bridge/ExternalPlayer.kt
@@ -102,7 +102,7 @@ class ExternalPlayer(
                 startTimeTicks = playOptions.startPositionTicks,
                 audioStreamIndex = playOptions.audioStreamIndex,
                 subtitleStreamIndex = playOptions.subtitleStreamIndex,
-                maxStreamingBitrate = playOptions.maxBitrate,
+                maxStreamingBitrate = null,
             ).onSuccess { jellyfinMediaSource ->
                 playMediaSource(playOptions, jellyfinMediaSource)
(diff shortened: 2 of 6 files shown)


app/src/main/assets/native/ExternalPlayerPlugin.js
```diff
@@ -141,12 +141,11 @@ export class ExternalPlayerPlugin {
     async getDeviceProfile() {
         return {
             Name: 'Android External Player Stub',
-            MaxStreamingBitrate: 100000000,
-            MaxStaticBitrate: 100000000,
-            MusicStreamingTranscodingBitrate: 320000,
+            MaxStreamingBitrate: 1000000000,
+            MaxStaticBitrate: 1000000000,
```
nielsvanvelzen, line 145:
```suggestion
            MaxStreamingBitrate: 1_000_000_000,
            MaxStaticBitrate: 1_000_000_000,
```

app/src/main/java/org/jellyfin/mobile/bridge/ExternalPlayer.kt
```diff
@@ -205,7 +167,8 @@ class ExternalPlayer(
     }
 
     private fun notifyEvent(event: String, parameters: String = "") {
-        if (event in arrayOf(Constants.EVENT_CANCELED, Constants.EVENT_ENDED, Constants.EVENT_TIME_UPDATE) && parameters == parameters.filter { it.isDigit() }) {
+        val allowedEvents = arrayOf(Constants.EVENT_CANCELED, Constants.EVENT_ENDED, Constants.EVENT_TIME_UPDATE)
```
nielsvanvelzen, line 170: Should this be a companion object constant?

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/mobile/bridge/ExternalPlayer.kt
Maxr1998: I suppose it would be better, yes.

nielsvanvelzen (approved):

PR #1198, opened 2023-09-08
Improve Android TV compatibility

PhieF:
**Changes**
- Setting default layout to TV when running on Android TV 
- Adding an entry to the android TV launcher

the web UI is better that the native app, which encounters too many problems on my end. Still a bit of work to be done on the web UI but still, it's more usable on my Nexus Player from 2014

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -59,6 +59,7 @@
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
                 <category android:name="android.intent.category.LAUNCHER" />
+                <category android:name="android.intent.category.LEANBACK_LAUNCHER" />
             </intent-filter>
             <!-- declare legacy support for voice actions -->
             <intent-filter>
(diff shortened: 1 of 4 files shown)


Maxr1998:
I believe it would be better to report the issues with the Android TV app and have those fixed than to increase the complexity of the mobile app by making it support platforms it isn't made for..

PhieF:
> I believe it would be better to report the issues with the Android TV app and have those fixed than to increase the complexity of the mobile app by making it support platforms it isn't made for..

The android TV app is native, not integrating the web UI like this one does

nielsvanvelzen:
We have two Android apps, one for touch devices (phones/tablets) and one for big screens (televisions/projectors). We will not support this app on big screens or the TV app on touch devices.

If you encounter issues with our Android TV app you should report them (or help fixing them) to the [respective repository](https://github.com/jellyfin/jellyfin-androidtv).

Closing this PR as we will not merge it.

PhieF:
Too bad... just 10 lines to get the web UI to android TV, but as  you wish, I will keep the webUI on my TV though since it works better on my device. Some bugs are already reported like these ones

https://github.com/jellyfin/jellyfin-androidtv/issues/2752

https://github.com/jellyfin/jellyfin-androidtv/issues/145


As well as multiple issue, for some files I need to select libVLC, some other exoplayer, my mother doesn't know how to use it properly so she is always stuck since the auto select doesn't select the good one (sometimes only sound is not working) 

Thanks for your great work and have a nice evening

Issue #1029, opened 2023-04-23
Chromecast unstable dosen't work


LedyBacer:
### Describe the bug

Chromecast unstable dosen't work from android app. From Google Chrome from pc or android works fine. 
https://github.com/jellyfin/jellyfin-chromecast/issues/401

### Logs

```shell
04-22 13:55:03.908 V/WebView (27595): chromecast.initialize, http://192.168.31.82:8096/web/chromecastPlayer-plugin.2dedfdab139185b94249.chunk.js (1)
04-22 13:55:03.909 V/WebView (27595): execCast initialize, https://www.gstatic.com/cv/js/sender/v1/cast_sender.js (1408)
04-22 13:55:03.918 V/WebView (27595): chromecast init success, http://192.168.31.82:8096/web/chromecastPlayer-plugin.2dedfdab139185b94249.chunk.js (1)
04-22 13:55:03.919 V/WebView (27595): chromecast receiver list empty, http://192.168.31.82:8096/web/chromecastPlayer-plugin.2dedfdab139185b94249.chunk.js (1)
04-22 13:55:03.919 V/WebView (27595): chromecast receiver found, http://192.168.31.82:8096/web/chromecastPlayer-plugin.2dedfdab139185b94249.chunk.js (1)
04-22 13:55:09.869 I/WebView (27595): Requesting http://192.168.31.82:8096/System/Endpoint, http://192.168.31.82:8096/web/main.jellyfin.bundle.js?ba0eba4f8e459eb7c494?deferred=true&ts=1682160901956 (2)
04-22 13:55:14.751 I/WebView (27595): Requesting http://192.168.31.82:8096/Sessions?ControllableByUserId=ff54816536a14e36b0e0df3828b4a882, http://192.168.31.82:8096/web/main.jellyfin.bundle.js?ba0eba4f8e459eb7c494?deferred=true&ts=1682160901956 (2)
04-22 13:55:15.923 V/WebView (27595): chromecast launching app..., http://192.168.31.82:8096/web/chromecastPlayer-plugin.2dedfdab139185b94249.chunk.js (1)
04-22 13:55:15.924 V/WebView (27595): execCast requestSession, https://www.gstatic.com/cv/js/sender/v1/cast_sender.js (1408)
04-22 13:55:17.192 V/WebView (27595): chromecast receiver found, http://192.168.31.82:8096/web/chromecastPlayer-plugin.2dedfdab139185b94249.chunk.js (1)
[...]

PR #1296, opened 2024-02-14
Chromecast: call success callback when result was successful

dumpfheimer:
**Changes**

Call the success callback  when result was successful.

To be honest this is so obvious that I am not sure if I am not missing something. I am still having issues with casting but this should be a step in the right direction.

**Issues**

Possibly related: https://github.com/jellyfin/jellyfin-android/issues/1029


Maxr1998 (approved):
This seems to have been in the code since before the rewrite, and I have no idea how I missed that. Thanks for spotting it!

PR #1297, opened 2024-02-14
Add support for the predictive back gesture

dumpfheimer:
**Changes**

Add support for the predictive back gesture. This improves gesture control on modern Android devices.

**Issues**


dumpfheimer:
I have converted this to a draft for further testing. It seems like I have gotten a system update which could have caused the improvements in gesture control, too.

Maxr1998:
I think without additional modifications to the app, this property change doesn't have any effect anyway. Since we currently use a web app, we can't really support predictive back in most parts of the app. Therefore, it doesn't make sense to enable it already. Thus, I'll close this pull request for now.

PR #1298, opened 2024-02-15
Improving Chromecast Integration

dumpfheimer:
**Changes**
This PR is targeting Chrome Cast support in the Jellyfin Android app. Chromecast seems to have multiple issues in my environment and I am aiming to make it a smoother experience.

I am by far NOT an expert in Android development but I was involved in Android development, so I am not absolutely incompetent to make improvements. However my PRs should be taken with a grain of salt and I will probably need a fair amount of help.

Changed RemotePlayerService wake lock and foreground binding to ensure it does not get killed by OS
TODOs:
- [x] Ensure RemotePlayerService acquires foreground lock (locked to notification)
- [x] Do not immediately release wake lock in RemotePlayerService and do not remove notification in onStopped immediately, but wait 15s if another track starts
- [x] Request wake lock for longer than current playing title
- [ ] Check why no notification is created initially when starting app and connecting to cast target (restarting app then recreates session and creates the notification, starting the service)
- [ ] Fix volume control
- [ ] Find out what causes "mediaItem" not found in JSONObject which kills the app

**Issues**


castillofrancodamian:
Also fix the bug where swiping on the timeline does not work when casting to TV using Google Cast. Playback stops when you want to go to another time.

Maxr1998:
Thanks for looking into this! So far, the changes look pretty good. Please tell me when you have specific questions about fixing the remaining TODOs.

Maxr1998:
What's the state of this? Would you prefer to continue working on it first, or do you think we can already merge the existing changes?

dumpfheimer:
Hey, sorry for the absurdly late response.

I have been using the debug app for quite some time now and it was much more reliable than the normal version in regards to chrome-casting.

If someone could look over the changes I have done who has more experience in android development and says that the changes make sense I believe this PR would make things better.

It is, however, not perfect. Things get out of sync. It seems like in the browser part events are filtered out and do not get passed down to the app sometimes. Unfortunately my knowledge is not deep enough to get the grater picture but I think to make things considerably better there would need to be a shared effort between android and web repository.

nielsvanvelzen:
Hey @dumpfheimer are you still working on finishing these changes?

nielsvanvelzen:
Closing as stale.

PR #1331, opened 2024-04-04
Allow changing download network preference

### LINKED ISSUE #1326: Option to change Download Preference from WiFi to Mobile Data not available after initial selection
### Describe the bug

I encountered a limitation within the Jellyfin Android app regarding the download settings. Initially, when prompted to select a download preference, I chose ‘WiFi only’. However, I now wish to enable downloads over mobile data as well, but the app does not offer an option to change this setting post the initial choice. The only workaround seems to be clearing the app’s data and starting the configuration process anew.

Steps to Reproduce:

1. Go to a movie or episode page
2. Go into options (3 dots) and select Download
3. When prompted to select a download preference, choose ‘WiFi only’.
4. Attempt to change the download preference to ‘WiFi + Mobile Data’ at a later stage.
5. Notice that there is no option available in the app settings to modify this preference.

The only solution appears to be clearing the app’s data and reconfiguring from scratch.
This issue has been discussed in the Jellyfin forum ([see the thread](https://forum.jellyfin.org/t-trying-to-download-shows-over-mobile-data-on-android-cant-waiting-for-wifi)), where it was suggested to clear app preferences as a solution. However, this is not a practical long-term solution, as it requires reconfiguration of the app settings. An option to change the download preference within the app settings would greatly enhance user flexibility and experience.

### Logs

_No response_

### Application version

2.6.0

Google Play

### Device information

Samsung S24

### Android version

Android 14

### Jellyfin server version

latest

- [ ] Web player (default)

taitsmith:
**Changes**
Adds a section in the settings screen which gives users the option of changing their preferred download network. Previously, users were unable to allow or restrict mobile data downloads after making the initial choice. Additionally updates the message displayed when the initial dialog is displayed to inform users their choice can be updated.

ss_dialog_update
ss_allowed_dialog

**Issues**
Fixes #1326

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/settings/SettingsFragment.kt
+++ b/app/src/main/java/org/jellyfin/mobile/settings/SettingsFragment.kt
@@ -25,6 +25,7 @@ import org.jellyfin.mobile.app.AppPreferences
 import org.jellyfin.mobile.databinding.FragmentSettingsBinding
 import org.jellyfin.mobile.utils.BackPressInterceptor
 import org.jellyfin.mobile.utils.Constants
+import org.jellyfin.mobile.utils.DownloadMethod
 import org.jellyfin.mobile.utils.applyWindowInsetsAsMargins
 import org.jellyfin.mobile.utils.extensions.requireMainActivity
 import org.jellyfin.mobile.utils.getDownloadsPaths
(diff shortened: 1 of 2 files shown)


taitsmith:
The issue I ran into is that `singleChoice()` expects a string key, and DownloadMethod is stored as an Int. As it is, the PR creates items with string keys and separately handles the updating of the download preference in `defaultOnSelectionChange`, which obviously is not ideal given that some amount of unnecessary work is being done and unneeded preferences are being created. Potential other solutions were updating the library to create something like `singleChoiceInt()` which seems like quite a bit of work for this one issue, or providing multiple check boxes like the player options currently provide.

Maxr1998 (approved):

Maxr1998:
> The issue I ran into is that `singleChoice()` expects a string key, and DownloadMethod is stored as an Int.

True, that's definitely a shortcoming of the library. Not a huge problem though and we can change this later when the library gets updated to directly support ints, I believe.. the only thing that is a bit of a pain is that this new preference will store the setting twice (as int in `pref_download_method` and as string in `download_method`, the latter being unused), and currently doesn't restore the current preference value from the int preference. Could you look into fixing the second point? Should be relatively simple.

Otherwise this looks pretty good.

taitsmith:
awesome thanks, was actually going to sit down today and look into updating the library so i'll get this wrapped up

Maxr1998:
Rebased your PR onto the latest master branch and fixed the conflicts. Thanks for your contribution!

Maxr1998 (approved):

PR #1364, opened 2024-04-26
Add new Jellyfin logo to connect screen

nielsvanvelzen:
XML copied from ATV repository.

**Changes**

Add the new Jellyfin logo to the connect screen.

image

But in fancy Jellypurple and Jellyblue for release flavor.


app/src/main/java/org/jellyfin/mobile/ui/screens/connect/ConnectScreen.kt
```diff
@@ -47,24 +41,14 @@ fun ConnectScreen(
 @Stable
 @Composable
 fun LogoHeader() {
-    CenterRow {
+    CenterRow(
+        modifier = Modifier.padding(vertical = 25.dp),
```
Maxr1998, line 45: Why specifically the 25 dp?

app/src/main/java/org/jellyfin/mobile/ui/screens/connect/ConnectScreen.kt
```diff
@@ -47,24 +41,14 @@ fun ConnectScreen(
...
-            contentScale = @Suppress("MagicNumber") FixedScale(1.2f),
+                .height(72.dp),
             contentDescription = null,
         )
-        Text(
-            text = stringResource(R.string.app_name_short),
-            modifier = Modifier
-                .padding(vertical = 56.dp)
-                .padding(start = 12.dp, end = 24.dp),
-            fontFamily = FontFamily(Font(R.font.quicksand)),
```
Maxr1998, line 65: Is that font file still used somewhere else? If not, could we remove it?

app/src/main/java/org/jellyfin/mobile/ui/utils/CenterRow.kt
```diff
@@ -10,9 +10,10 @@ import androidx.compose.ui.Modifier
 
 @Composable
 inline fun CenterRow(
+    modifier: Modifier = Modifier,
     content: @Composable RowScope.() -> Unit,
 ) = Row(
-    modifier = Modifier.fillMaxWidth(),
+    modifier = modifier.fillMaxWidth(),
```
Maxr1998, line 16: I'm not 100% sure about the modifier application order, but normally, you would do size first, then padding. But it's possibly that this doesn't matter if you use `fillMaxWidth()`.

app/src/main/java/org/jellyfin/mobile/ui/screens/connect/ConnectScreen.kt
nielsvanvelzen: Graphic design is just inputting random numbers until it feels right

app/src/main/java/org/jellyfin/mobile/ui/utils/CenterRow.kt
nielsvanvelzen: Changed to
```
 modifier = Modifier
        .fillMaxWidth()
        .then(modifier),
```

Maxr1998: Perfect!

app/src/main/java/org/jellyfin/mobile/ui/screens/connect/ConnectScreen.kt
Maxr1998: The odd number irks me, but oh well.

Maxr1998 (approved):

PR #1365, opened 2024-04-27
Remove baseline workaround

nielsvanvelzen:
We temporarily added this to fix reproducible builds. We've since upgrades to a newer version of the Android Gradle Plugin that fixes the bug we worked around. As such we can remove it.

**Changes**
- Remove baseline workaround

**Issues**


Maxr1998 (approved):

Maxr1998:
Totally forgot about this one..

PR #1391, opened 2024-05-19
Only DirectPlay hardware-accelerated video codecs for ExoPlayer

levogevo:
**Changes**
- When using ExoPlayer as the playback client option, only direct play videos that have hardware-supported decoders. As I was making this PR, I noticed the older one #1135 and I relate to exactly the problem they encountered (and proposed fix).

**Issues**
AV1 playback can range from great to unplayable depending on the bitrate/encode settings of the video and device playing it back. Android Jellyfin users already have options to attempt directplay, by either using the web client/external players or alternative clients like Findroid. We do not have an option for playing back videos that need transcoding help. The trick of lowering the internet bandwidth doesn't work if the av1 encode is already pretty low bitrate (10-15) but due to things like film-grain synthesis and low(er) preset encoding, are still very resource intensive to decode, causing many many dropped frames.

I am not an android developer at all, nor believe this is the best option. I am hoping if it is (at least) not a passable option, it can still highlight a need for change. Ideally, a dialog/dropdown to allow the client to select the encoder on server-side but that was more complicated than to get this at least working as intended. Thank you for reading.

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/player/deviceprofile/DeviceProfileBuilder.kt
+++ b/app/src/main/java/org/jellyfin/mobile/player/deviceprofile/DeviceProfileBuilder.kt
@@ -1,6 +1,8 @@
 package org.jellyfin.mobile.player.deviceprofile
 
+import android.media.MediaCodecInfo
 import android.media.MediaCodecList
+import android.os.Build
 import org.jellyfin.mobile.app.AppPreferences
 import org.jellyfin.mobile.utils.Constants
 import org.jellyfin.sdk.model.api.CodecProfile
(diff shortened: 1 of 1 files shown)


satmandu:
Would this still be a good thing?

Obviously direct play by default with exoplayer doesn't make sense if the client hardware can't support hardware decoding?

nielsvanvelzen:
I don't think this is a good idea. Whenever the device reports it can play a certain codec we should be able to trust that. There's a lot of video codecs that decode perfectly fine on CPU.

nielsvanvelzen:
Closing this as unwanted. As mentioned before, a lot of devices can efficiently playback with software decoders. Additionally, hardcoding various codec names is _not_ maintainable in the long-term.

PR #1395, opened 2024-05-22
Fix broken external links

### LINKED ISSUE #984: Links in episode "Overview" (<plot>) don't work as expected
### Describe the bug

An episode "Overview" (`<plot>`) can contain html to describe the content, including links (`<a>`).
This works well in the browser version of jellyfin, however the android app will treat them as app-internal-links.

## current behaviour
Links in `<plot>` open as part of the app.
If you click on a link, it will open that link inside the jellyfin app as if it was part of the app.

when a link is clicked:
- there is no way to go back (other than closing the app and restarting)
- no browser bar is shown
- no way to open a link externally in another app (browser)

## expected behaviour
links in `<plot>` should be opened externally (open in the default android app that handles that type of link. For example open normal website links with the browser, open youtube links with the youtube app, ...)

## My usecase
I am downloading videos and podcasts from websites and create `.nfo` files for them. In the `<plot>` I add the original link. Sometimes I want to open them to view comments or lookup other details.

## How to reproduce
1. edit the overview of an episode to contain an html link (this can be done via UI or `<plot>` in `.nfo`**)
2. view the episode in the android app. it will show the plot and show the link as a blue clickable link
3. click on the link

** if you edit the .nfo file, you have to escape the html. so for example: 
```<plot>&lt;a href="https://github.com/jellyfin/jellyfin-android"&gt;my test link&lt;a&gt;</plot>```

### Logs

_No response_

### Application version

2.4.4

F-Droid

### Device information

Sony Experia XA2 Ultra

### Android version

LineageOS 11

### Jellyfin server version

10.8.9

AllegraCodes:
**Changes**
Links in metadata would open in the app and lock it up. I override the function shouldOverrideUrlLoading to launch external links with the system. External links are defined as those that aren't intercepted by shouldInterceptRequest and don't contain the server hostname.

**Issues**
Fixes #984

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/webapp/JellyfinWebViewClient.kt
+++ b/app/src/main/java/org/jellyfin/mobile/webapp/JellyfinWebViewClient.kt
@@ -41,12 +41,12 @@ abstract class JellyfinWebViewClient(
 
     override fun shouldOverrideUrlLoading(view: WebView, request: WebResourceRequest): Boolean {
         Timber.d("shouldOverrideUrlLoading: %s", request.url.toString())
-        val url = request.url
         return when {
-            url.toString().startsWith(server.hostname) -> false
+            !request.hasGesture() -> false
+            request.url.toString().startsWith(server.hostname) -> false
             else -> {
-                Timber.d("shouldOverrideUrlLoading: external link, handle with system")
-                val intent = Intent(Intent.ACTION_VIEW, url)
+                Timber.d("shouldOverrideUrlLoading: external link with gesture, handle with system")
+                val intent = Intent(Intent.ACTION_VIEW, request.url)
                 if (intent.resolveActivity(view.context.packageManager) != null) {
                     startActivity(view.context, intent, null)
                 }
--- a/app/src/main/java/org/jellyfin/mobile/webapp/JellyfinWebViewClient.kt
+++ b/app/src/main/java/org/jellyfin/mobile/webapp/JellyfinWebViewClient.kt
@@ -1,11 +1,13 @@
 package org.jellyfin.mobile.webapp
 
+import android.content.Intent
 import android.net.Uri
 import android.net.http.SslError
 import android.webkit.SslErrorHandler
 import android.webkit.WebResourceRequest
 import android.webkit.WebResourceResponse
 import android.webkit.WebView
+import androidx.core.content.ContextCompat.startActivity
 import androidx.webkit.WebResourceErrorCompat
 import androidx.webkit.WebViewAssetLoader.AssetsPathHandler
 import androidx.webkit.WebViewClientCompat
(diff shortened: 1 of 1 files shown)


Maxr1998:
Hi, thanks for the PR. Do you have a specific example where this happens? For me, external links open in the browser just fine. So this seems to be an issue in the web client for some links instead.

AllegraCodes:
Sure, I replicated the issue by editing the metadata of one of my items to have `<a href="https://developer.android.com/guide/components/intents-filters#extras">my test link</a>` in the Overview. Then I navigated to that item's page in the app and tapped the link, and it opened in the app, unresponsive to back gestures.
Screenshot_20240522-162040

Maxr1998:
Fascinating. What type of library, item and which metadata field was it?

AllegraCodes:
I always used the Overview field and the behavior is the same for a movie, show, playlist, or collection

Maxr1998:
I see, I can reproduce it now, thanks for providing all these details.

I'm worried that the current matchers aren't enough and would cause pages to open externally which shouldn't. Authentication components/plugins, for example. I feel like web should preferably handle this, but I'm not sure how.

AllegraCodes:
that's a good point, i'll keep looking into this to see how android handles these calls and try to make it more robust.

Maxr1998:
Thanks! Appreciate your work, hopefully you can find a good solution to this issue.

AllegraCodes:
ok, so the [docs](https://developer.android.com/reference/android/webkit/WebViewClient#shouldOverrideUrlLoading(android.webkit.WebView,%20android.webkit.WebResourceRequest)) offer some notes on when this method actually gets called, and it's fairly limited. It's called for navigations initiated by the web page, the user, or an HTTP redirect.

Notably, it's not called for any navigations initiated by the app, but that doesn't mean it's impossible to trigger it from the app's code. The WebViewFragment triggers it when it loads <serverUrl>/web on initialization (but not when it loads a javascript url from the lifecycle scope). This is allowed to load internally because it's on the same host (including port) as the server, and it's the only instance I found of the app itself triggering this method in my limited testing so far.

One way I could lock it down a bit further is to require the request be associated with a gesture (such as a click, but we only get a boolean, not the type) in order to load it externally. I think I'll add that now so any reviewers can see it.

With all that said, I'm an android and web app novice, so I could easily be missing something with how android works here.

MakD:
This also happens in the trailer button for the remote trailers in the android app. It the user has YouTube installed the YouTube intent should kick in and open the trailer in the app rather than jellyfin itself

nielsvanvelzen:
I'm not sure if this is an issue that must be fixed in the app. The web interface (jellyfin-web) should properly set the `target="_blank"` attribute on links so the app can open them externally. If it doesn't do that the app is asked to open it in-app by the webui.

nielsvanvelzen:
Closing as stale.

PR #1404, opened 2024-06-03
Add in app download functionality

7ritn:
This PR internalizes the download process.

**Changes**
Downloads are stored inside the app directory including thumbnail and external subtitle files. MediaSource data is downloaded as well and stored inside the Room Database.
A new Downloads fragment is added to show available downloads. Downloads are played back using the Exoplayer.

**What's left**
This is not a complete list

- [x] better download overview
- [x] add functionality to remove downloads
- [x] add button in main menu to get to downloads (currently only available from server selection)
- [x] overall polish

**Notes**
There is still lots of work to do, but it is already working. I new to Android development, so I am really sorry if something is not correct. Furthermore I implement this functionality as part of my Master's degree, so even if for whatever reason this PR is rejected, I will still complete the project to my best ability. I sorry that I changed some ExoPlayer code even though, it is currently reworked, but due to time constraints, I could not wait.

Thanks for any feedback :)


7ritn:
Updated download page so it looks somewhat nice. Also added functionality to delete downloads.

7ritn:
So I would call the PR in a beta state. Feature wise, I think it is already at a good point. Adding the view downloads button requires modifying the web app. If something is missing I can still add it, but otherwise next step would be the polish.

rcv11x:
Thank you very much for this PR, it looks very good, I have a few questions to ask you:

1 - This is to download media on the device and watch them locally without internet?
2 - When you download the media are they compressed so that they take up less space on the device or are they downloaded with the actual size they occupy? ❤️

satmandu:
Does the app use the downloaded file automatically now if the file has been downloaded?

Also, I'm not sure that further compression of the downloaded file makes sense, as compression algorithms like zstd or whatever are not likely to reduce file size much for video files.

7ritn:
While thinking about implementing the ability to download transcoded media (streams), I realized it would be better if ExoPlayer would handle downloading, since it natively supports HLS streams, making future efforts easier. So I rewrote the download logic to utilize ExoPlayers in-buildt download capabilities.

satmandu:
I would love to test this if you want to post some builds...

I have a flight coming up. 😅

app/src/main/java/org/jellyfin/mobile/player/source/JellyfinMediaSource.kt
```diff
@@ -10,13 +11,15 @@ import org.jellyfin.sdk.model.api.PlayMethod
...
 
+@Serializable(with = JellyfinMediaSourceSerializer::class)
 class JellyfinMediaSource(
     val itemId: UUID,
     val item: BaseItemDto?,
     val sourceInfo: MediaSourceInfo,
     val playSessionId: String,
     val liveStreamId: String?,
     val maxStreamingBitrate: Int?,
+    var isDownload: Boolean = false,
```
neBM, line 22: > `The constructor(itemId: UUID, item: BaseItemDto?, sourceInfo: MediaSourceInfo, playSessionId: String, liveStreamId: String?, maxStreamingBitrate: Int?, isDownload: Boolean, startTimeTicks: Long?, audioStreamIndex: Int?, subtitleStreamIndex: Int?) has too many parameters. The current threshold is set to 10. [LongParameterList]`

A potential solution for this might be to serialise the params in a DTO to group some of our client side media source state together? Something a bit like this:

```kotlin
data class MediaSourceDetails(
    val itemId: UUID,
    val item: BaseItemDto?,
    val sourceInfo: MediaSourceInfo,
    val playSessionId: String,
    val liveStreamId: String?
)

class JellyfinMediaSource(
    val details: MediaSourceDetails,
    val maxStreamingBitrate: Int?,
    var isDownload: Boolean = false,
    private var startTimeTicks: Long? = null,
    audioStreamIndex: Int? = null,
    subtitleStreamIndex: Int? = null
) 
```

neBM (commented):
Much of the detekt issues blocking the PR seems like quick fixes except for one :/
Looks like we're the unlucky PR that needs that final param and hit the limit! 😆 

I've thought of a potential solution but it's by no means perfect.

app/src/main/java/org/jellyfin/mobile/player/source/JellyfinMediaSource.kt
neBM: That would work! :) 
My only thoughts on that is _what if the users network connection is unstable_... I'm thinking a perfect use case for this PR would be for travelling with a potentially intermittent connection &mdash; an API request might succeed but the user would (should 😁 ) be playing from the local file.

I have yet to do a deep dive in the codebase but what's your thoughts on always defaulting to the local file if it exists in the `DownloadDao` as a viable alternative? We could simply poll the Dao in the `isDownload` predicate?

satmandu: Agree on having the app be useable when the network is down being an important use case.

The most obvious example is being on a plane and downloading a video before boarding, after which one may not have network access.

neBM: Absolutely! I believe @7ritn's solution of polling the server when the user plays the content to see if there's a connection would work for when there is absolutely no connection (i.e., on a plane).
My concern is _what if the connection is slow or intermittent?_ For example, if you're on a train, the connection might work (if you're lucky!) but, the moment you go in a tunnel, the player will still be following the "I'm not playing local content" behaviour (as the connection was available at the time of polling).

We could also consider an "offline mode" toggle switch that could force use of the `DownloadDao` but implementations in other apps have always been clunky (and still seems excessive for a code quality rule :/).

7ritn: Due to the caching / downloading being handled by exoplayer, if there is a download matching the requested content id available, exoplayer will automatically playback the download. Thus with intermittent or even good connection the download will be preferred. This currently only works if the original quality is used, I'm not sure what the behaviour is if the user chooses a different quality i.e. transcoded stream.

Maxr1998: If no elegant solution exists, it's also ok to simply suppress that warning, btw. Don't feel pressured by warnings like these. They are a good practice, but not always mandatory.

7ritn: > If no elegant solution exists, it's also ok to simply suppress that warning, btw. Don't feel pressured by warnings like these. They are a good practice, but not always mandatory.

Well we are now working on a solution using inheritance to have RemoteJellyfinMediaSource and a LocalJellyfinMediaSource which also additionally contains download url and oter downloads specs, simplifying the database

7ritn: Done

app/src/main/res/layout/fragment_downloads.xml
```diff
@@ -0,0 +1,22 @@
...
+    <androidx.appcompat.widget.Toolbar
+        android:id="@+id/toolbar"
+        android:layout_width="match_parent"
+        android:layout_height="56dp"
+        android:background="?colorPrimary"
+        app:layout_constraintTop_toTopOf="parent" />
+
+    <androidx.recyclerview.widget.RecyclerView
+        android:id="@+id/recycler_view"
+        android:layout_width="match_parent"
```
SubhrajyotiSen, line 17: It's usually recommended to use 0dp instead of `match_parent` when using a ConstraintLayout. 
Using `match_parent` sometimes leads to weird sculling issues when a RecyclerView is inside a ConstraintLayout

app/build.gradle.kts
```diff
@@ -81,6 +82,7 @@
         buildConfig = true
         viewBinding = true
         compose = true
+        dataBinding = true
```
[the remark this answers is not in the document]
SubhrajyotiSen: If you don't have a specific reason to use data binding here, I can help refactor the PR to avoid data binding altogether since the project doesn't already use it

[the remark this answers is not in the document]
7ritn: I can switch to view binding. I think going back to findViewById is not desireable since it has no compile time verification.

app/src/main/res/layout/fragment_downloads.xml
7ritn: sure

app/build.gradle.kts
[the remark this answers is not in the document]
SubhrajyotiSen: I believe viewBinding would be perfect here since it's already enabled in the project. 
I also agree that `findViewById` is definitely not desirable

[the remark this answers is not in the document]
nielsvanvelzen: Using Jetpack Compose is our current preferred way for UI

app/src/main/res/layout/fragment_downloads.xml
7ritn: When I set width to 0dp, I get a MissingConstraints warning. In the settings fragment it is also handled the same.

app/build.gradle.kts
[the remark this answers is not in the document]
7ritn: Okay just added the commit to move to view binding

app/src/main/res/layout/fragment_downloads.xml
SubhrajyotiSen: Are you adding the following as well?

```xml
app:layout_constraintStart_toStartOf="parent"
app:layout_constraintEnd_toEndOf="parent"
```

satmandu:
Any chance we could get the workflows approved?

satmandu:
I'm not able to build this on Ubuntu 24.04 with the android sdk installed using `sudo apt install -y android-sdk`.

I get
```
Deprecated Gradle features were used in this build, making it incompatible with Gradle 9.0.

You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.

For more on this, please refer to https://docs.gradle.org/8.8/userguide/command_line_interface.html#sec:command_line_warnings in the Gradle documentation.

BUILD FAILED in 1m 59s
4 actionable tasks: 4 executed
```


I did this:
```
git clone https://github.com/7ritn/jellyfin-android.git -b download
cd jellyfin-android
./gradlew assembleRelease
```

satmandu:
It would also be nice to be able to select a player app for the files that get listed under downloads.

For instance, downloaded HDR videos may not play through the built-in player, but may play ok with VLC...

SubhrajyotiSen:
This patch should help resolve the detekt failures:
```
---
 .../java/org/jellyfin/mobile/downloads/DownloadsAdapter.kt | 7 ++++++-
 1 file changed, 6 insertions(+), 1 deletion(-)

diff --git a/app/src/main/java/org/jellyfin/mobile/downloads/DownloadsAdapter.kt b/app/src/main/java/org/jellyfin/mobile/downloads/DownloadsAdapter.kt
index 41e4c1ae..f258b9e5 100644
--- a/app/src/main/java/org/jellyfin/mobile/downloads/DownloadsAdapter.kt
+++ b/app/src/main/java/org/jellyfin/mobile/downloads/DownloadsAdapter.kt
@@ -37,7 +37,12 @@ class DownloadsAdapter(
             val mediaItem: BaseItemDto? = downloadEntity.mediaSource.item
             binding.textViewName.text = downloadEntity.mediaSource.name
             binding.textViewDescription.text = when {
-                mediaItem?.seriesName != null -> context.getString(R.string.tv_show_desc, mediaItem.seriesName, mediaItem.parentIndexNumber, mediaItem.indexNumber)
+                mediaItem?.seriesName != null -> context.getString(
+                    R.string.tv_show_desc,
+                    mediaItem.seriesName,
+                    mediaItem.parentIndexNumber,
+                    mediaItem.indexNumber,
+                )
                 mediaItem?.productionYear != null -> mediaItem.productionYear.toString()
                 else -> downloadEntity.mediaSource.id
             }
--
```

7ritn:
Detekt now runs sucessfully

7ritn:
> It would also be nice to be able to select a player app for the files that get listed under downloads.
> 
> For instance, downloaded HDR videos may not play through the built-in player, but may play ok with VLC...

This is not possible since we do not have access to the files when using ExoPlayer

satmandu:
If a video is downloaded, playing the video when connected to the network tries to stream the file from the server, instead of preferring the locally downloaded file.

Would it be useful to be able to change that?

7ritn:
For me the cached version is used when playing the video by clicking on the item in the web view. Requirement for this to happen is to use the ExoPlayer in the settings. Can you check if you have selected the ExoPlayer and not the Web Player?

satmandu:
> For me the cached version is used when playing the video by clicking on the item in the web view. Requirement for this to happen is to use the ExoPlayer in the settings. Can you check if you have selected the ExoPlayer and not the Web Player?

Ah yes! It was reset to web player when I updated to my locally built version!

satmandu:
Maybe some sort of `if cached && player_not_set_to_exoplayer then notify user` logic wold be warranted? (Or offer to use exoplayer to play the file?)

7ritn:
The PR is already rather large and I'd prefer keeping further improvements to new PRs after this one gets merged

-------- Original Message --------

neBM:
I've been daily driving the latest build. I've noticed the Download All functionality for an entire TV show appears to successfully download one or two eps only to not download any of the remaining eps.

If someone can trial this behaviour to test this isn't just my env, that would be great. Otherwise, this might be worth patching. 

I have yet to test this against other media groups such as download all for a season.

satmandu:
Also spent the last week using a build of this.

I would note that when viewing a cached video offline from the downloads folder, that the watched state/progress doesn't get uploaded back to the jellyfin server once a connection is re-established.

Should I make that a separate feature request?

7ritn:
> I've been daily driving the latest build. I've noticed the Download All functionality for an entire TV show appears to successfully download one or two eps only to not download any of the remaining eps.
> 
> If someone can trial this behaviour to test this isn't just my env, that would be great. Otherwise, this might be worth patching.
> 
> I have yet to test this against other media groups such as download all for a season.

I'll look into this

> Also spent the last week using a build of this.
> 
> I would note that when viewing a cached video offline from the downloads folder, that the watched state/progress doesn't get uploaded back to the jellyfin server once a connection is re-established.
> 
> Should I make that a separate feature request?

Yeah. I am aware of this. This can be added later if you want. Would need to remember the relevant server instance and also verify if play sessions are remembered that long so progress reports can be still made

neBM:
> Also spent the last week using a build of this.
> 
> I would note that when viewing a cached video offline from the downloads folder, that the watched state/progress doesn't get uploaded back to the jellyfin server once a connection is re-established.
> 
> Should I make that a separate feature request?

Good catch, @satmandu!

Personally, I feel that the purpose of this PR: "To add download functionality", has been met. In addition to this, we already have a fairly large PR and I fear for keeping this branch up to date if/when potentially other large PRs get merged upstream. 

+1 for doing this in a separate PR 👍

satmandu:
Is there any chance of getting this merged to master so that new PRs to expand this functionality don't have to target this PR?

I, for one, am fine with testing builds built from master until v2.7.0 is tagged...

kimonmeier:
An eta when this gets merged?

7ritn:
The commit I just pushed should fix the problems of dropping downloads when multiple downloads are enqueued. The problem was the small size of the event buffer and the policy to drop old events. Remaining problem is the a bit freaky download notification, when downloading files at the same time.

7ritn:
Would need to update the getForegroundNotification to consider all pending downloads when building the notification ugh

kimonmeier:
I tested the whole pr for a bit and i found a strange bug. 

Somehow the replacement of the `DefaultDataSource` with the `CacheDataSource` in the `AppModule`, breaks the change in bitrate during playback. If you change the bitrate, the player reports "source error".
```
playerFailed [eventTime=22.98, mediaPos=395.34, window=0, period=0, errorCode=ERROR_CODE_PARSING_MANIFEST_MALFORMED
  com.google.android.exoplayer2.ExoPlaybackException: Source error
      at com.google.android.exoplayer2.ExoPlayerImplInternal.handleIoException(ExoPlayerImplInternal.java:684)
      at com.google.android.exoplayer2.ExoPlayerImplInternal.handleMessage(ExoPlayerImplInternal.java:654)
      at android.os.Handler.dispatchMessage(Handler.java:102)
      at android.os.Looper.loopOnce(Looper.java:224)
      at android.os.Looper.loop(Looper.java:318)
      at android.os.HandlerThread.run(HandlerThread.java:67)
  Caused by: com.google.android.exoplayer2.ParserException: Input does not start with the #EXTM3U header.{contentIsMalformed=true, dataType=4}
      at com.google.android.exoplayer2.source.hls.playlist.HlsPlaylistParser.parse(HlsPlaylistParser.java:268)
      at com.google.android.exoplayer2.source.hls.playlist.HlsPlaylistParser.parse(HlsPlaylistParser.java:75)
      at com.google.android.exoplayer2.upstream.ParsingLoadable.load(ParsingLoadable.java:181)
      at com.google.android.exoplayer2.upstream.Loader$LoadTask.run(Loader.java:420)
      at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
      at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:644)
      at java.lang.Thread.run(Thread.java:1012)
]
```

If you debug into the exoplayer source code you see that in `com.google.android.exoplayer2.source.hls.playlist.HlsPlaylistParser` in the method `checkPlaylistHeader`. That the header isn't correct anymore. The char codes are completly different and if you read the whole line you get a malformed header. 
Output of Download via DefaultDataSource:
`#EXTM3U
... and more`
Output of Download via CacheDataSource:
`Eߣ�B��matroskaB��B��S�g	W�t�M�t�M��S��I�fS��M��S��T�kS���M��S��S�kS��W�1�M��S��C�pS��`

Did i configure something wrong or is it indeed a bug?

7ritn:
I just tried to replicate this bug. It only seems to appear on some bitrates i.e. 360p. But on higher bitrates, I do not get the exception.

7ritn:
Commit #7ce600c fixes the HLS content source error. Problem was that the exoplayer cache cached videos when playing them back, this was not intended and let to complication. The commit makes the cache read-only.

7ritn:
Commit #6db98b7 improves download notification when multiple downloads are in progress. This should fix all current outstanding problems.

felixschndr:
Would it be possible for someone to take a look at this PR and review it please? That'd be great ❤️

Maxr1998:
Sorry for taking so long to look into this. I'd like to get these changes in next. Would you mind rebasing this onto the latest master so that I can properly test and review it?

7ritn:
I currently can't really test changes rn. So I can't that any guarantees if it still works.
Is this conflict-resolving commit enough or do you want a proper rebase? I am not really used to having to rebase onto so many new commits lol

satmandu:
@Maxr1998 Any chance you have time for comments or a review of this rebase?

7ritn:
I have to be real, this feature is the most requested feature of the official Jellyfin project (https://features.jellyfin.org/?view=most-wanted). It is understandable that people lead busy lives, but if the maintainer is not able to look into this for almost a whole year, then maybe a solution needs to be found. It makes sense that there are many other Jellyfin Android apps with more features and better maintenance.

satmandu:
@nielsvanvelzen would you be ok with an edge release with some of these long waiting PRs pulled in?

A quarterly or sooner unstable release cadence with lower expected polish, separate from the default _ship it when it is ready_ schedule would help keep outside developers who want to collaborate interested...

nielsvanvelzen:
@satmandu I agree with you that the current situation is not ideal. If someone is willing to help out with one or all of these tasks we can proceed with merging:

- Fixing the merge conflicts (ideally the author of course)
- Reviewing the code
- Testing the new functionality

Polishing and fixing smaller issues can be done in separate pull requests.

7ritn:
Sure I can look into this *again*

7ritn:
Removed merge conflict, will now make sure everything works.

Maxr1998:
> I have to be real, this feature is the most requested feature of the official Jellyfin project (https://features.jellyfin.org/?view=most-wanted). It is understandable that people lead busy lives, but if the maintainer is not able to look into this for almost a whole year, then maybe a solution needs to be found. It makes sense that there are many other Jellyfin Android apps with more features and better maintenance.

I get you, the situation totally sucks. I probably shouldn't even have left the project for so long, because getting back into it only got harder because of that. The reality is that I simply have no time to actively maintain this app right now, and it pains me since I also had a lot of plans (including unfinished work like the player UI rewrite). The little time I can spend on it every few weeks to months usually goes mostly into housekeeping and basic maintenance aka dependency updates and such, but that may be the wrong approach. I'm honestly fine with pulling in this PR mostly unreviewed at this point, considering multiple community members already worked on it. 

> `@nielsvanvelzen` would you be ok with an edge release with some of these long waiting PRs pulled in?
> 
> A quarterly or sooner unstable release cadence with lower expected polish, separate from the default _ship it when it is ready_ schedule would help keep outside developers who want to collaborate interested...

That was already my plan when I last talked to Niels, but that was probably weeks if not months ago, again. Having a download feature in the app would be amazing, and merging this and other bigger PRs would then allow us to iterate on the existing state without risking to break other open PRs again and again.

Unfortunately, I really can't and don't want to promise anything right now. But I definitely hear and understand the frustration, and won't get in the way while trying to find a solution.

7ritn:
Fixed cache not being properly used when downloading.

Tested:
	- detekt builds successfully (when runnig gradlew detekt)
	- watching an episode not downloaded using ExoPlayer while online with external and embedded subtitles
	- watching an episode dowloaded using Exoplayer while online
		- uses download cache in DirectPlay
		- fast seeking
	- downloading multiple episodes
		- progress notification for total download progress
		- individual success notification for each download
	- disrupting download
		- when going offline, download is resumed when connected again
		- when Jellyfin is closed, the download service continues
	- watching a downloaded episode while offline with external and embedded subtitles
		- seeking in the episode is instant
		- subtitle switching works seemless
	- deleting episode from Downloads
		- episodes is removed from Available Downloads
		- cache is cleared

nielsvanvelzen (approved):
Approving (& merging) for the sake of progression. I have not reviewed or tested these changes, I just looked over it briefly.

As mentioned in my previous comment, any issues with these changes can be resolved in separate pull requests. If you're interested in helping out please do so! We'd appreciate any testers/reviewers & developers. For questions please visit our Android chat via Matrix or Discord (listed at jellyfin.org/contact)

7ritn:
Just one thing I noticed that when downloading at lot of episodes at the same time, the total download progress tends to reset, when an episode finishes downloading. Could look into it the next few days.

app/src/main/java/org/jellyfin/mobile/data/DatabaseModule.kt
```diff
@@ -9,9 +9,11 @@ val databaseModule = module {
         Room.databaseBuilder(androidApplication(), JellyfinDatabase::class.java, "jellyfin")
             .addMigrations()
             .fallbackToDestructiveMigrationFrom(1)
+            .fallbackToDestructiveMigrationFrom(2)
```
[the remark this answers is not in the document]
Maxr1998: It's a bit unfortunate that I never finished and submitted my original review, but I just noticed it again when building the latest version for myself: right now, the app clears the whole database, including all currently saved servers, when updating the app. This definitely shouldn't happen and should be fixed by adding a proper Room migration.
That should preferably be done in a new file, and referenced from this builder in the `DatabaseModule`.

PR #1412, opened 2024-06-23
Fix background playback stopping on Android 14

### LINKED ISSUE #1378: Music playback stops when app is in the background
### Describe the bug

I am using a Google Pixel 6 and have app battery usage set to 'unrestricted' - see screenshot - but as soon as I switch away from having the app in the foreground music playback stops. This makes it unusable for audio playback. None of my other audio apps have this problem, e.g. Spotify; VLC.

As per [previous bug report](https://github.com/jellyfin/jellyfin-android/issues/1208) I see a trail of reports of this same issue here on GitHub https://github.com/jellyfin/jellyfin-android/issues/948 and elsewhere, e.g. [Reddit](https://www.reddit.com/r/jellyfin/comments/jfkuij/jellyfin_stops_playing_next_song_if_jellyfin_is/) but no solutions. In each case we see the issue was closed by a bot without effective resolution. 

In my case I am not using a VPN

303915035-db97c12d-27aa-41c7-aa46-19e1196d2847

### Logs

_No response_

### Application version

2.6.1

Google Play

### Device information

Google Pixel 6

### Android version

Android 14

### Jellyfin server version

10.8.13

- [ ] Web player (default)

### LINKED ISSUE #1401: Jellyfin pauses music playback after 20 - 30 seconds of the app being in the background
### Describe the bug

I play a song on the app, exit out and it continues playing, then after 30 seconds or so it stops and will only play again once I click on the app. I use graphene os if that has anything to do with it? I disabled all the graphene os memory stuff for the app but it still doesnt work. Battery optimisation is on unoptimised. Thanks

### Logs

_No response_

### Application version

2.6.1

F-Droid

### Device information

Google Pixel 7

### Android version

Graphene OS

### Jellyfin server version

10.8.13

- [X] Web player (default)

michd:
`RemotePlayerService` service was not being properly started as a foreground service for media playback, causing playback stoppage after every track (or sooner for some folks), on android 14.

Different vendors may be more or less aggressive with killing of non-foreground service etc, so people may have gotten different results from the same root issue.

**Changes**
- Changes the manifest entry for `RemotePlayerService` to include the `foregroundServiceType="mediaPlayback"` attribute (see ["Foreground service types are required"](https://developer.android.com/about/versions/14/behavior-changes-14#fgs-types))
- Declares the foreground service behavior as part of the notification (see same req)
- Uses `startForeground` instead of `notificationManager.notify`
- Uses `ContextCompat.startForegroundService` to send intents to the service

Some of these changes can also be found in @dumpfheimer's PR [Improving Chromecast Integration #1298](https://github.com/jellyfin/jellyfin-android/pull/1298)

**Issues**

- Fixes #1378
- Fixes #1401

---

(For a bit of my background: my day job is working on a music streaming android app and we've previously hit similar issues)

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/webapp/RemotePlayerService.kt
+++ b/app/src/main/java/org/jellyfin/mobile/webapp/RemotePlayerService.kt
@@ -265,10 +265,6 @@ class RemotePlayerService : Service(), CoroutineScope {
                     setPriority(Notification.PRIORITY_LOW)
                 }
 
-                if (AndroidVersion.isAtLeastS) {
-                    setForegroundServiceBehavior(Notification.FOREGROUND_SERVICE_IMMEDIATE)
-                }
-
                 setContentTitle(title?.let { HtmlCompat.fromHtml(it, HtmlCompat.FROM_HTML_MODE_LEGACY) })
                 setContentText(artist?.let { HtmlCompat.fromHtml(it, HtmlCompat.FROM_HTML_MODE_LEGACY) })
                 setSubText(album)
@@ -350,12 +346,12 @@ class RemotePlayerService : Service(), CoroutineScope {
                 startForeground(
                     MEDIA_PLAYER_NOTIFICATION_ID,
                     notification,
-                    ServiceInfo.FOREGROUND_SERVICE_TYPE_MANIFEST
+                    ServiceInfo.FOREGROUND_SERVICE_TYPE_MANIFEST,
                 )
             } else {
                 startForeground(
                     MEDIA_PLAYER_NOTIFICATION_ID,
-                    notification
+                    notification,
                 )
             }
 
--- a/app/src/main/java/org/jellyfin/mobile/webapp/RemotePlayerService.kt
+++ b/app/src/main/java/org/jellyfin/mobile/webapp/RemotePlayerService.kt
@@ -11,6 +11,7 @@ import android.content.BroadcastReceiver
 import android.content.Context
 import android.content.Intent
 import android.content.IntentFilter
+import android.content.pm.ServiceInfo
 import android.graphics.Bitmap
 import android.media.AudioAttributes
 import android.media.AudioManager
@@ -263,6 +264,11 @@ class RemotePlayerService : Service(), CoroutineScope {
                 } else {
                     setPriority(Notification.PRIORITY_LOW)
                 }
+
+                if (AndroidVersion.isAtLeastS) {
+                    setForegroundServiceBehavior(Notification.FOREGROUND_SERVICE_IMMEDIATE)
+                }
+
                 setContentTitle(title?.let { HtmlCompat.fromHtml(it, HtmlCompat.FROM_HTML_MODE_LEGACY) })
                 setContentText(artist?.let { HtmlCompat.fromHtml(it, HtmlCompat.FROM_HTML_MODE_LEGACY) })
                 setSubText(album)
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/jellyfin/mobile/webapp/RemotePlayerService.kt
```diff
@@ -263,6 +264,11 @@ class RemotePlayerService : Service(), CoroutineScope {
                 } else {
                     setPriority(Notification.PRIORITY_LOW)
                 }
+
+                if (AndroidVersion.isAtLeastS) {
+                    setForegroundServiceBehavior(Notification.FOREGROUND_SERVICE_IMMEDIATE)
+                }
```
Maxr1998, line 270: Might not be necessary [since the `foregroundServiceType` in the manifest is `mediaPlayback`](https://developer.android.com/develop/background-work/services/foreground-services#notification-immediate).

app/src/main/java/org/jellyfin/mobile/webapp/RemotePlayerService.kt
```diff
@@ -340,7 +346,18 @@ class RemotePlayerService : Service(), CoroutineScope {
...
+                    MEDIA_PLAYER_NOTIFICATION_ID,
+                    notification,
+                    ServiceInfo.FOREGROUND_SERVICE_TYPE_MANIFEST
+                )
+            } else {
+                startForeground(
+                    MEDIA_PLAYER_NOTIFICATION_ID,
+                    notification
+                )
+            }
```
Maxr1998, line 360: Could use [`ServiceCompat.startForeground()`](https://developer.android.com/reference/androidx/core/app/ServiceCompat#startForeground(android.app.Service,int,android.app.Notification,int)) instead.

Maxr1998 (commented):
Looks good, have a few comments nonetheless. Please also fix the lint failure (should be the missing trailing comma, I think).

app/src/main/java/org/jellyfin/mobile/webapp/RemotePlayerService.kt
michd: True, however the [`ServiceInfo.FOREGROUND_SERVICE_TYPE_MANIFEST`](https://developer.android.com/reference/android/content/pm/ServiceInfo#FOREGROUND_SERVICE_TYPE_MANIFEST) constant was only added in API level 29, so it either requires a version check, or has to be duplicated locally. I think I prefer the current way over that. What's your preference?

app/src/main/java/org/jellyfin/mobile/webapp/RemotePlayerService.kt
michd: Oh yeah thank you, that's indeed redundant. Removed that and fixed the lint issues.

app/src/main/java/org/jellyfin/mobile/webapp/RemotePlayerService.kt
Maxr1998: Huh, I'd have expected a compat version of that to exist. Since it doesn't, I agree and will merge it as is.

Maxr1998 (approved):

Maxr1998:
Thank you!

simson:
I can't tell for android 14, but using android 15 i have the same issue : background stop after 30sec. I disabled all battery optimisation : allowing background usage "Unrestricted" on my pixel 6a. Is there another thing  i can do or test ?

Issue #184, opened 2020-10-10
Reconnect Chromecast on app resume


idittansikte:
**Describe the feature you'd like**
When watching a movie and the Android OS thinks its time to shutdown/suspended Jellyfin app it is annoying being needed to reconnect manually to the Chromecast when opening the App again. 

Netflix, HBO etc, automatically detects that it the current Chromecast session is owned by the app and automatically reconnects on these occasions.

Can we make this a thing in Jellyfin as well? Would increase user-experience a lot for me.

neopc10:
It would require the "intent to join" part of the Android sender. The new cast client isn't available on 2.0 yet.

idittansikte:
Related thing I've noticed: The Jellyfin app think it's connected to chromecast when I have actually started casting from another app and thus taken over the casting from there. I don't know if this is always the case though.

Maxr1998:
I don't think we'll still support this in the current web client, unfortunately. The current Chromecast code is a huge mess, and especially the integration to the web client is just not something I want to closely take a look into.
Long term, we're planning to do a fully native client (there's some initial progress already in the `compose` branch), and for that I'll rewrite the Chromecast code from scratch.

Issue #459, opened 2021-07-04
Chromecast controls disappear immediately 


apollace:
**Describe the bug**
When the app is connected to a Chromecast deice and I start to cast a video the controls appears for a short time at the bottom of the screen just to disappear in less than a second. After that they are gone and the video is not controllable anymore from the app. 

**To Reproduce**
1. Connect to a Chromecast
2. Press play to any random video on your jellyfin home screen
3. The controls appear 
4. The controls disappear. 

**Expected behavior**
The controls should remain visible for the whole reproduction time of the video. 

**Logs**
Collecting the logs

**System:**
 - App version: 2.2.3
 - Android version: Android 11
 - Device manufacturer and model: OnePlus 8 Pro
 - Jellyfin server version: 10.7.6
 - The type of video player implementation used: Chromecast

mholt:
See #464 for slightly different repro instructions and a screenshot of the remote control.

bradbeattie:
Workaround I found. When it disappears, hit play a second time. Controls persist after that.

gr3ger:
Just here to confirm that this is still happening on the latest stable version.
I will compile from master when I find the time and see if the issue is still present.

Flaxee:
This is still an issue for me with server version 10.7.7 and app version 2.4.4.
Are there any work being done to improve this?

bestrocker221:
Same issue here, when I connect to the cast and then play the movie, controls disappear and it is impossible to stop/play/forward/etc the movie. Any update?

MrReSc:
I have the problem too.

Mike55555:
> Workaround I found. When it disappears, hit play a second time. Controls persist after that.

Doesn't work for me. Only way I can get the controls back is to disconnect the cast connection, force kill Jellyfin, then open Jellyfin, reconnect the cast, and then maybe maybe maybe the controls come back. But usually no, just the same missing controls behaviour.

GottZ:
any specific logs I should look into to give proper support / debug this myself? I even have this issue in microsoft edge on windows desktop. unrelated.. I know.. but if it happens there too, it might be related to this issue here as well.
(the only workaround I found on android is constantly closing and opening jellyfin on android until the remote shows up properly)

![20230619_195209](https://github.com/jellyfin/jellyfin-android/assets/559564/e7f0e0cd-3008-46cb-8902-ab08fd17c266)

Snowmansir:
Still happening!

[...]

PR #1469, opened 2024-08-25
Prevent destruction of session, so that an app restart can reconnect (chromecast).

Tukajo:
This is my first commit but I'd like to keep things small-ish/have targeted improvement.

There is an issue where the controls have repeatedly disappeared for various users:
https://github.com/jellyfin/jellyfin-android/issues/459

This change set at least allows the user to forcibly close the app and re-open it, allowing for a reconnection to the session.

The downside is that the session will continue until the user intentionally stops it. Which means that it could potentially continue playing in the background indefinitely unless the server is somehow configured with an "Are you still here?" style of message & session booting based on no user input.

I still think this is better for the time being, because the alternative is that every time I lose control of the chromecast controls, I have to restart the app, which actually stops my TV from chromecasting, and I lose all my progress. This is exceedingly annoying as is.

By comparison, when YouTube is closed while chromecasting, it continues to operate (eventually displaying that "are you still here?" message).

Further notes:
1. I do not think this should be the final state. I am having trouble trying to figure out how to get the app (on opening) to automatically reconnect to any still-alive session, and re-display the controls automatically. Today the user has to manually navigate the controls again to reconnect, and hit "remote control" again. At least it's better than the entire casting session stopping.
 

**Changes**
- Do not destroy the session when app is closed.
- Use of implicit this
- Some build debug changes (can remove these if we want, it just sped up some of my builds).

**Issues**
https://github.com/jellyfin/jellyfin-android/issues/459


app/src/main/java/org/jellyfin/mobile/MainActivity.kt
```diff
@@ -190,6 +190,7 @@ class MainActivity : AppCompatActivity() {
                 fragment.onUserLeaveHint()
             }
         }
+        super.onUserLeaveHint()
```
Tukajo, line 193: This was added due to the linter/editor complaining it should not have been left out.

app/src/proprietary/java/org/jellyfin/mobile/player/cast/Chromecast.java
```diff
@@ -33,7 +33,7 @@ public final class Chromecast implements IChromecast {
     /**
      * Object to control the media.
      */
-    private ChromecastSession media;
+    private ChromecastSession chromecastSession;
```
Tukajo, line 36: changed the variable name to be more "descriptive" as to what this variable actually was. For the longest time I got a little confused and thought this was the actual media, not the session.

gradle.properties
```diff
@@ -9,7 +9,7 @@
 #
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
-org.gradle.jvmargs=-Xmx2G -XX:MaxMetaspaceSize=512m
+org.gradle.jvmargs=-Xmx8G -XX:MaxMetaspaceSize=1G
```
Tukajo, line 12: I have a beefier debugging machine, but we can reset this down if folks want. This helped speed up my build times quite a bit.

Maxr1998: [This](https://discuss.gradle.org/t/precedence-of-gradle-properties/41366) might be interesting for you. I'd prefer to keep the lower defaults in this repo.

3flex:
Would this also fix #184? Or is #184 describing this limitation:

> I am having trouble trying to figure out how to get the app (on opening) to automatically reconnect to any still-alive session, and re-display the controls automatically. Today the user has to manually navigate the controls again to reconnect, and hit "remote control" again

Tukajo:
> Would this also fix #184? Or is #184 describing this limitation:
> 
> > I am having trouble trying to figure out how to get the app (on opening) to automatically reconnect to any still-alive session, and re-display the controls automatically. Today the user has to manually navigate the controls again to reconnect, and hit "remote control" again

It will allow you to reconnect. (You can build this yourself to see).

But it's not automatic yet. I was having trouble getting the automatic reconnect to work.

However you can manually reconnect by reperforming the normal steps you would do.

gradle.properties
Tukajo: Definitely I'll remove these changes when I can get back on my computer.

nielsvanvelzen (approved):

PR #1472, opened 2024-09-01
Switch to binary-only Gradle distribution

3flex:
**Changes**
Alter build to use binary-only distribution of the Gradle wrapper.

This saves nearly 100MB on download (138MB instead of 230MB) and more than that in disk space. The complete distribution of the wrapper offers no value to the majority of contributors.

**Issues**

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -17,10 +17,6 @@ allprojects {
     }
 }
 
-tasks.wrapper {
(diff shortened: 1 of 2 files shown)


Maxr1998 (approved):
You're right, we don't need this anymore. It was useful when I was writing some advanced build logic back then, but isn't required anymore now that the build system is pretty stable.

PR #1475, opened 2024-09-01
Configure Room with Gradle plugin

3flex:
**Changes**
Migrate to the Room Gradle plugin introduced in Room 2.6.0.

> The plugin configures the project such that generated schemas (which are
> an output of the compile tasks and are consumed for auto-migrations) are
> correctly configured to have reproducible and cacheable builds.

https://developer.android.com/jetpack/androidx/releases/room#gradle-plugin

Running Gradle tasks such as `assembleDebug` twice in a row now leads to tasks in the second run being up to date, allowing incremental compilation to work correctly.

**Issues**

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -67,6 +67,7 @@ kotlin-parcelize = { id = "org.jetbrains.kotlin.plugin.parcelize", version.ref =
 kotlin-ksp = { id = "com.google.devtools.ksp", version.ref = "kotlin-ksp" }
 compose-compiler = { id = "org.jetbrains.kotlin.plugin.compose", version.ref = "kotlin" }
 android-junit5 = { id = "de.mannodermaus.android-junit5", version.ref = "android-junit5" }
+android-room = { id = "androidx.room", version.ref = "androidx-room" }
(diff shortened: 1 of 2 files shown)


gradle/libs.versions.toml
```diff
@@ -67,6 +67,7 @@ kotlin-parcelize = { id = "org.jetbrains.kotlin.plugin.parcelize", version.ref =
 kotlin-ksp = { id = "com.google.devtools.ksp", version.ref = "kotlin-ksp" }
 compose-compiler = { id = "org.jetbrains.kotlin.plugin.compose", version.ref = "kotlin" }
 android-junit5 = { id = "de.mannodermaus.android-junit5", version.ref = "android-junit5" }
+android-room = { id = "androidx.room", version.ref = "androidx-room" }
```
3flex, line 70: The alias should be `androidx-room`, so `libs.plugins.androidx.room`

Maxr1998: Right, that would be preferable. I'll quickly do that before I merge :+1:

Maxr1998 (approved):

Issue #1420, opened 2024-07-06
"Play next episode automatically" is ignored when using the ExoPlayer


domeniko-gentner:
### Describe the bug

1. Disable play next episode automatically
2. Use the ExoPlayer
3. Play an Episode
4. Next Episode will play automatically

As described above, disabling "play next episode automatically" is ignored by the app when using the ExoPlayer. When using the internal player, this issue is not present. I haven't tried the external player setting.

### Application version

2.6.1

### Where did you install the app from?

Google Play

### Device information

Google Pixel 8 Pro

### Android version

Android 14  of June 5 2004

### Jellyfin server version

 10.9.7

### Which video player implementations does this bug apply to?

- [ ] Web player (default)
- [X] Integrated player (ExoPlayer)
- [ ] External player (VLC, mpv, MX Player)

ForestQ:
It's impossible to live with this error. I also fall asleep with this client.

frashure:
Also experiencing this issue. 
Jellyfin Android 2.6.2 from F-Droid
Pixel 7 Pro
Android 15 (CalyxOS 6.2.1)

EmeraldCoco:
Same issue android.

zsolt-p:
+1

dahamsta:
I've checked this setting so many times! Please fix!

boblarouche:
Same here. I've been using external player for years to mitigate this issue and it sucks because then the "resume" and "watched" features don't work. Either remove the option or fix it please.

Razator73:
This is an issue for me too. It works fine with the Internal Player, but when I switched to the ExoPlayer it is broken and continues on. I much prefer the ExoPlayer so hoping this is fixed soon. Some more details

Jellyfin Server version - 10.10.6
Jellyfin App version - 2.6.2
Android version - 15
Phone - Google Pixel 7

JayceeB1:
Got this issue too... Please fix 😥

raider-com:
I can also replicate this with Android

merschformann:
Same issue for me. Web Player works, but integrated player ignores "Play next episode automatically" setting. This is a bit annoying due to the limited features of the web player.

Jellyfin Server version: 10.10.7
Jellyfin App version: 2.6.3
Android version: 15
Phone: Google Pixel 8 Pro

Thanks for all the work btw. Great project(s)! 🤗

VSofter:
Same here too. 
[...]

Issue #1384, opened 2024-05-14
Play next episode automatically bugged since 10.9.0


kraM1tt:
### Describe the bug

Despite having the option unticked, it continues to play the next episodes in the series infinitely.

![Screenshot_20240514_133003_Jellyfin](https://github.com/jellyfin/jellyfin-android/assets/89113987/2e33e6b9-078e-4381-83bc-a57f51b10932)

It worked perfectly before the server update. I usually fall asleep, the episode finishes and my phone auto locks itself, but now I wake up to a dead phone, the JF payback reporting log confirms it plays episodes all night.

Thanks

### Application version

2.6.1

### Where did you install the app from?

Google Play

### Device information

Samsung S21

### Android version

Android 14, one ui 6.0

### Jellyfin server version

10.9.0

### Which video player implementations does this bug apply to?

- [ ] Web player (default)
- [X] Integrated player (ExoPlayer)
- [ ] External player (VLC, mpv, MX Player)

kraM1tt:
Not sure if it helps but I just tested Findroid and it obeys the setting. Also Jellyfin for Tizen OS obeys the setting, so it does seem to be an Jellyfin Android issue

Crash1602:
I have the same problem with a Google Pixel 8 and Google Pixel 8 Pro. It plays one episode after the other without stopping. Jellyfin Server 10.9.3 on Windows 11

ConnorS1110:
Does this still happen for anyone in this thread? I had fixed an issue that sounds ~very~ similar to this in [#5398](https://github.com/jellyfin/jellyfin-web/pull/5398) in the web repo

Crash1602:
Hi,
I tested the problem a few days ago with Jellyfin Server 10.9.6 and the AndroidApp 2.6.1 on both Pixel Smartphones. The Problem still exists with the integrated Player, but there are no Problems with the WebPlayer and it works as expected.

kraM1tt:
Yes still broken for me, have to use Findroid to get around it, Web player has issues direct playing a lot of things

dahamsta:
Also experiencing this, v. annoying.

Crash1602:
I hope there will be a fix soon :-) because the WebPlayer is not as convenient as the integrated Player.

kraM1tt:
Bumping before it gets auto closed

BotBlake:
After testing on my end this issue no longer persists.
[...]

PR #1479, opened 2024-09-05
Fix episode autoplay preference having no effect

lfranke42:
**Changes**
The playback setting "Play next episode automatically" is being ignored, next episodes are always played regardless of the setting state.

This PR adds a request to the user endpoint to check if the autoplay setting has been enabled.
If internal exoplayer is used, the autoplay preference is now being taken into account when an episode of a series ends.

**Issues**
Fixes:
#1470
#1420
#1384


Maxr1998 (approved):
Looks good. Ideally, the user config would be handled more centrally, but that's a topic for later. For now, loading it in the player is preferable.

PR #1493, opened 2024-10-11
Fix the issue of blurry media images in notifications.

peerless2012:
**Changes**
Use MediaMetadata art to override large icon in notification, fix art not clear issue.

Before:
微信截图_20241011232259

After:
微信截图_20241011232400


peerless2012:
The notification art is land, maybe use backdrop art resource?

Maxr1998 (approved):
Thanks for your PR and bringing this issue to my attention. I'd argue that setting the metadata in the notification helper isn't the best location, but it is good enough for now.

I realized that the artwork loading can generally be improved by implementing a content provider, but this is out of scope for this PR.

Thus, I'll merge it as is. Thanks again!

pavichokche:
I updated to the 2.6.3-beta1 but I'm still seeing blurry images in the notification. I cleared the app cache and it's still the same. Android 15 Samsung S25

PR #1506, opened 2024-11-04
Update jellyfin-sdk-kotlin to v1.6.1

jakobkukla:
**Changes**

Updated SDK version and fixed build errors.

Also added the API authorization header to exoplayer since the access token is no longer passed via URL for Android Auto (see https://github.com/jellyfin/jellyfin-sdk-kotlin/pull/871).

I have tested the Android Auto functionality and it seems to work fine. The only thing I cannot test and suspect that may not work is casting. Not sure how I would add the header there since CastPlayer cannot use a DataSource.Factory. In https://github.com/google/ExoPlayer/issues/4760 they claim that the receiver app is responsible for authentication. Not sure if this is implemented in [jellyfin/jellyfin-chromecast](https://github.com/jellyfin/jellyfin-chromecast) though.

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/app/AppModule.kt
+++ b/app/src/main/java/org/jellyfin/mobile/app/AppModule.kt
@@ -1,6 +1,7 @@
 package org.jellyfin.mobile.app
 
 import android.content.Context
+import androidx.core.net.toUri
 import coil.ImageLoader
 import com.google.android.exoplayer2.ext.cronet.CronetDataSource
 import com.google.android.exoplayer2.extractor.DefaultExtractorsFactory
@@ -110,17 +111,23 @@ val applicationModule = module {
         val dataSourceFactory = DefaultDataSource.Factory(context, baseDataSourceFactory)
 
         // Add authorization header. This is needed as we don't pass the
-        // access token in the url for Android Auto.
+        // access token in the URL for Android Auto.
         ResolvingDataSource.Factory(dataSourceFactory) { dataSpec: DataSpec ->
-            val authorizationHeaderString = AuthorizationHeaderBuilder.buildHeader(
-                clientName = apiClient.clientInfo.name,
-                clientVersion = apiClient.clientInfo.version,
-                deviceId = apiClient.deviceInfo.id,
-                deviceName = apiClient.deviceInfo.name,
-                accessToken = apiClient.accessToken,
-            )
+            // Only send authorization header if URI matches the jellyfin server
+            val baseUrlAuthority = apiClient.baseUrl?.toUri()?.authority
 
-            dataSpec.withRequestHeaders(hashMapOf("Authorization" to authorizationHeaderString))
+            if (dataSpec.uri.authority == baseUrlAuthority) {
+                val authorizationHeaderString = AuthorizationHeaderBuilder.buildHeader(
+                    clientName = apiClient.clientInfo.name,
+                    clientVersion = apiClient.clientInfo.version,
+                    deviceId = apiClient.deviceInfo.id,
+                    deviceName = apiClient.deviceInfo.name,
+                    accessToken = apiClient.accessToken,
+                )
+
+                dataSpec.withRequestHeaders(hashMapOf("Authorization" to authorizationHeaderString))
+            } else
+                dataSpec
         }
     }
     single<MediaSource.Factory> {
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/jellyfin/mobile/app/AppModule.kt
```diff
@@ -102,7 +106,20 @@ val applicationModule = module {
...
+
+        // Add authorization header. This is needed as we don't pass the
+        // access token in the url for Android Auto.
+        ResolvingDataSource.Factory(dataSourceFactory) { dataSpec: DataSpec ->
+            val authorizationHeaderString =
+                "MediaBrowser Token=\"${apiClient.accessToken}\", " +
+                    "Client=\"${apiClient.clientInfo.name}\", " +
+                    "Device=\"${apiClient.deviceInfo.name}\", " +
+                    "DeviceId=\"${apiClient.deviceInfo.id}\", " +
+                    "Version=\"${apiClient.clientInfo.version}\""
```
nielsvanvelzen, line 119: The SDK has an AuthorizationHeaderBuilder utility for this that has proper escaping

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/mobile/app/AppModule.kt
```diff
@@ -102,7 +107,21 @@ val applicationModule = module {
...
+            val authorizationHeaderString = AuthorizationHeaderBuilder.buildHeader(
+                clientName = apiClient.clientInfo.name,
+                clientVersion = apiClient.clientInfo.version,
+                deviceId = apiClient.deviceInfo.id,
+                deviceName = apiClient.deviceInfo.name,
+                accessToken = apiClient.accessToken,
+            )
+
+            dataSpec.withRequestHeaders(hashMapOf("Authorization" to authorizationHeaderString))
+        }
```
nielsvanvelzen, line 124: Some sources, like Live TV and strm files do not load video through the Jellyfin server but from other origins directly. We should **never** sent the Jellyfin access token to those origins.

app/src/main/java/org/jellyfin/mobile/player/source/MediaSourceResolver.kt
```diff
@@ -60,8 +59,8 @@ class MediaSourceResolver(private val apiClient: ApiClient) {
 
         // Load additional item info if possible
         val item = try {
-            val response by itemsApi.getItemsByUserId(ids = listOf(itemId))
-            response.items?.firstOrNull()
+            val response by itemsApi.getItems(ids = listOf(itemId))
+            response.items.firstOrNull()
```
nielsvanvelzen, line 63: The UserLibraryApi has a getItem operation that would be more efficient

nielsvanvelzen (changes requested):

satmandu:
Any chance of getting this reviewed again @nielsvanvelzen ? It would be nice to start generating builds from the updated sdk...

app/src/main/java/org/jellyfin/mobile/app/AppModule.kt
```diff
@@ -102,7 +108,27 @@ val applicationModule = module {
...
+            if (dataSpec.uri.authority == baseUrlAuthority) {
+                val authorizationHeaderString = AuthorizationHeaderBuilder.buildHeader(
+                    clientName = apiClient.clientInfo.name,
+                    clientVersion = apiClient.clientInfo.version,
+                    deviceId = apiClient.deviceInfo.id,
+                    deviceName = apiClient.deviceInfo.name,
+                    accessToken = apiClient.accessToken,
+                )
+
+                dataSpec.withRequestHeaders(hashMapOf("Authorization" to authorizationHeaderString))
```
nielsvanvelzen, line 128: This is unsafe as in some cases (like strm files and livestreams) the stream is not loaded from Jellyfin but from an external URL. We should **never** send the Jellyfin authorization header to those sources.

nielsvanvelzen (dismissed):
I did not test these changes and only reviewed the usages of the SDK. Those all look fine to me.

satmandu:
Would someone be so kind as to trigger the build and test workflows on this PR?

nielsvanvelzen:
GH does not let me trigger the builds so the pull request probably needs to be rebased or have master merged into it.

jakobkukla:
rebased on master. I think you should be able to trigger the build now @nielsvanvelzen

jakobkukla:
I think this is ready to merge @nielsvanvelzen :)

Maxr1998 (approved):

Maxr1998:
Thanks for the migration!

PR #1507, opened 2024-11-10
ExoPlayer: Implement media segment support

jakobkukla:
Initial support for media segments.

Most of this is shamelessly copied from the Android TV implementation in https://github.com/jellyfin/jellyfin-androidtv/pull/4052. @nielsvanvelzen, I tried to reflect that in the commit message. Let me know if that's a problem. 

**Changes**

The playback related implementation is pretty much copied 1:1 from ATV. AskToSkip is not implemented yet but I plan to do so
in a following PR.

I reused the settings from the web client by implementing a JS plugin and subscribing to settings changes. I usually try to avoid JavaScript in my daily life and have never worked with the nativeshell plugin system before. I'd appreciate if you'd take a closer look at these changes.

On another note: I'm not quite sure when to use koin's `inject()` vs `get()`. I tried to look at how the objects were injected in other places and followed that. 

**Issues**

Depends on #1506


felixschndr:
Looking forward to this being merged, thank you for your work @jakobkukla!

jakobkukla:
@Maxr1998 I think this is ready to merge :)

nielsvanvelzen:
Hi @jakobkukla can you rebase this to fix the merge conflicts? I'll have a look at the changes afterwards!

jakobkukla:
@nielsvanvelzen done :)

app/src/main/assets/native/MediaSegmentsPlugin.js
```diff
@@ -0,0 +1,32 @@
+export class MediaSegmentsPlugin {
+    SETTING_PREFIX = 'segmentTypeAction';
+
+    constructor({ events, appSettings, dashboard }) {
+        this.appSettings = appSettings;
+        this.dashboard = dashboard;
+
+        events.on(appSettings, 'change', (_, name) => this.onSettingsChanged(name));
```
nielsvanvelzen, line 8: I think it would be better if we also read all settings when the webview launches, so probably here in the plugin constructor.
Otherwise, if you configure media segments now (while the app doesn't support it) they won't work like configured when the app is updated with these changes.

app/src/main/java/org/jellyfin/mobile/player/PlayerViewModel.kt
```diff
@@ -408,6 +417,46 @@ class PlayerViewModel(application: Application) : AndroidViewModel(application),
...
+
+    private fun addSkipAction(mediaSegment: MediaSegmentDto) {
+        val player = playerOrNull ?: return
+
+        player.createMessage { _, _ ->
+            // We can't seek directly on the ExoPlayer instance as not all media is seekable
+            // the seek function in the PlaybackController checks this and optionally starts a transcode
+            // at the requested position
+            // TODO: The above is probably true for jellyfin-android as well.
+            // But I believe there is no such logic here.
```
nielsvanvelzen, line 449: Correct, the seekTo you invoke now is directly on ExoPlayer. I think that's fine for now though as the mobile app doesn't support "transcode on unsupported seek" functionality yet.

app/src/main/java/org/jellyfin/mobile/bridge/MediaSegments.kt
```diff
@@ -0,0 +1,35 @@
...
+            "Commercial" -> MediaSegmentType.COMMERCIAL
+            else -> return
+        }
+
+        val action: MediaSegmentAction = when (actionString) {
+            "None" -> MediaSegmentAction.NOTHING
+            "Skip" -> MediaSegmentAction.SKIP
+            "AskToSkip" -> MediaSegmentAction.ASK_TO_SKIP
+            else -> return
+        }
```
Maxr1998, line 31: Don't valueOf/fromName work here?

app/src/main/java/org/jellyfin/mobile/player/PlayerViewModel.kt
```diff
@@ -408,6 +417,46 @@ class PlayerViewModel(application: Application) : AndroidViewModel(application),
         }
     }
 
+    private fun applyMediaSegments(jellyfinMediaSource: JellyfinMediaSource) {
+        viewModelScope.launch {
+            if (jellyfinMediaSource.item != null) {
+                val mediaSegments = runCatching {
+                    mediaSegmentRepository.getSegmentsForItem(jellyfinMediaSource.item)
+                }.getOrNull().orEmpty()
```
Maxr1998, line 425: This shouldn't throw, since the function already uses runCatching. Can thus be simplified.

app/src/main/java/org/jellyfin/mobile/app/AppPreferences.kt
```diff
@@ -100,6 +103,19 @@ class AppPreferences(context: Context) {
...
+     * The actions to take for each media segment type. Managed by the MediaSegmentRepository.
+     */
+    var mediaSegmentActions: String
+        get() = sharedPreferences.getString(
+            Constants.PREF_MEDIA_SEGMENT_ACTIONS,
+            mapOf(
+                MediaSegmentType.INTRO to MediaSegmentAction.ASK_TO_SKIP,
+                MediaSegmentType.OUTRO to MediaSegmentAction.ASK_TO_SKIP,
+            ).toMediaSegmentActionsString(),
+        )!!
```
Maxr1998, line 116: Is the `!!` needed here?

Maxr1998 (dismissed):

app/src/main/java/org/jellyfin/mobile/bridge/MediaSegments.kt
```diff
@@ -0,0 +1,35 @@
...
+import android.webkit.JavascriptInterface
+import org.jellyfin.mobile.player.mediasegments.MediaSegmentAction
+import org.jellyfin.mobile.player.mediasegments.MediaSegmentRepository
+import org.jellyfin.sdk.model.api.MediaSegmentType
+import org.koin.core.component.KoinComponent
+import org.koin.core.component.get
+
+@Suppress("unused")
+class MediaSegments(private val context: Context) : KoinComponent {
+    private val mediaSegmentRepository: MediaSegmentRepository = get()
```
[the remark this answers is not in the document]
jakobkukla: Like this?

nielsvanvelzen:
@coderabbitai review

nielsvanvelzen (changes requested):
Aside from the repository injection in the view model this looks good to merge.

nielsvanvelzen (approved):

Issue #155, opened 2020-09-30
Add next chapter button


JulioNahum:
**Describe the feature you'd like**
Implement the next chapter button in the Exoplayer player. I see that they are improving the new player a lot and it would be a good idea to add the button to go to the next chapter or add an option that sends a notification, being able to modify the number of seconds it would take to display, thus being able to go to the next chapter.

PR #1533, opened 2024-12-30
Add chapter skipping support to exoplayer client

CeruleanRed:
**Changes**
Added buttons to support chapter skipping in the exoplayer client. Buttons are only present if chapters actually exist on the file being played. Button behaviour was modeled on how the web client works (e.g. the 10s interval in which going back a chapter results in going to the previous chapter, instead of starting the current one at 0)

**Images**

(Ignore the video quality, it's trying to software decode AV1 on the emulator)
image

**Extra comments**
- Currently when hiding the skip chapter buttons, there is a slight lag until the buttons disappear due to having to check if chapters exist, which can only happen when the media is loaded. Any alternative that improves this would be welcome.
- The "watched" status of the chapter markings only updates after the user has let go of the thumb when scrubbing

**Fixes**
#155


CeruleanRed:
There seems to be no way to add query parameters since #1506, so this will have to be fixed when that is supported again

CeruleanRed:
Hi, I can definitely fix my PR, probably sometime within the next week, if this is needed faster for the release, I'll try to make time for it earlier.

nielsvanvelzen:
No hurry! It would be nice to have this in 2.7 but there's no timeframe for that yet. It will probably be a few months :)

nielsvanvelzen:
No changes made to that, you can still configure it in the client settings like you always could.

CeruleanRed:
That is super weird because my emulator doesn't have that setting anymore. I checked it against my phone, and my phone has the setting, but the emulator does not. I think the mobile app just sees itself as the browser wrapper, since the server is also reporting it as chrome. I will look into the issue some more tomorrow, and see if I can get it to run, but I've never had that happen before

nielsvanvelzen:
Ok found the culprit, will fixed with #1643 :)

CeruleanRed:
Rebased onto master, not entirely sure why GitHub is reporting so many file changes, checked with a PR in my forked repo and there it just showed the 12 I touched.

Is there anyway to programmatically get the size attributes from a drawable? I just checked my changes from January and I still have some magic numbers floating around like the chapter marking width and height in [here](https://github.com/jellyfin/jellyfin-android/blob/0b098c2b290589852d6cae5ab71392a6f2c043b8/app/src/main/java/org/jellyfin/mobile/player/ui/ChapterMarking.kt) file that I'd like to be more explainable

CeruleanRed:
Looking at the git history, I messed this branch up quite hard doing the rebase, might be easiest to just switch to a new one and cherry-pick my changes, instead of trying to fix it here, but let me know how you want me to proceed

felixschndr:
> Rebased onto master, not entirely sure why GitHub is reporting so many file changes, checked with a PR in my forked repo and there it just showed the 12 I touched.

I would assume that something went wrong during the rebase. Maybe rebased on an old state of `main`. Maybe you want to check the [git branch graph](https://stackoverflow.com/a/1838881). I am on my phone right now, so can't check.

CeruleanRed:
I decided not to fight it and just made a new branch here #1646, sorry for the inconvenience, was the fastest way I could get it to not be the current abomination

PR #1625, opened 2025-05-07
Migrate to compose BOM

Maxr1998:
https://developer.android.com/develop/ui/compose/bom


nielsvanvelzen:
We don't use BOM's in our Gradle projects because we use Renovate to keep the versions up-to-date. Also for Compose the mix of versions we use in ATV is different from the BOM so using it in the mobile app will cause conflicts when the two repositories are merged.

Maxr1998:
> We don't use BOM's in our Gradle projects because we use Renovate to keep the versions up-to-date.

I don't think they contradict each other. Renovate can update the BOM just as well, and ensuring the individual versions are compatible and not having to specify versions for each individual dependency is imo preferable.

> Also for Compose the mix of versions we use in ATV is different from the BOM so using it in the mobile app will cause conflicts when the two repositories are merged.

I currently doubt that'll ever happen, since the use case of the two apps are just too different. Instead, I'd suggest code sharing via libraries. But even then, this change would be easily reverted, so I don't think this argument would block this PR in any way :thinking:

Maxr1998:
We decided we'll continue to explicitly specify all library versions instead of switching to a BOM.

PR #1627, opened 2025-05-07
Implement DownloadManagement app feature in nativeshell

nielsvanvelzen:
Related: jellyfin/jellyfin-web#6833

**Changes**

- Implement support for the new "downloadmanagement" app feature in the nativeshell. This will add a "downloads" link to the app menu so the user can access downloads when connected to a server.
- Sort features array in our JS file.

**Issues**


Maxr1998 (approved):

nielsvanvelzen:
Note: merging **after** jellyfin-web change is merged

PR #1646, opened 2025-05-18
Add chapter skipping support to exoplayer client

CeruleanRed:
**Changes**
Added buttons to support chapter skipping in the exoplayer client. Buttons are only present if chapters actually exist on the file being played. Button behaviour was modeled on how the web client works (e.g. the 10s interval in which going back a chapter results in going to the previous chapter, instead of starting the current one at 0)

**Images**

(Ignore the video quality, it's trying to software decode AV1 on the emulator)
image

**Extra comments**
- Currently when hiding the skip chapter buttons, there is a slight lag until the buttons disappear due to having to check if chapters exist, which can only happen when the media is loaded. Any alternative that improves this would be welcome.
- The "watched" status of the chapter markings only updates after the user has let go of the thumb when scrubbing
- This is the new branch since I messed up a rebase, old branch and discussion can be found at #1533 

**Fixes**
#155

## Summary by CodeRabbit

- **New Features**
  - Added visual chapter markers to the player progress bar for easier navigation.
  - Introduced previous and next chapter buttons in the player controls.
  - Enabled direct navigation to previous or next chapters during playback.

- **Improvements**
  - Chapter markers update dynamically as playback progresses.
  - Enhanced player UI layout to accommodate new chapter navigation and markers.

- **Bug Fixes**
  - Improved consistency in time unit conversions throughout the player.

- **Chores**
  - Added new icons, drawable, color, and dimension resources to support chapter navigation and visualization.


felixschndr:
FYI: This is a followup of https://github.com/jellyfin/jellyfin-android/pull/1533

CeruleanRed:
Sorry, hadn't noticed the linting didn't work until now. Any idea how to test this locally so I can fix the problem? Running a lint in android studio gives me completely different results than in the pipeline, but I'm not really an android dev so maybe I'm doing something very wrong

CeruleanRed:
I've added a "TooManyFunctions" suppression to show that the linting would technically work, but I'd like to see if there's a solution that just passes the markings to the PlayerViewModel more elegantly

nielsvanvelzen:
Sorry about the merge conflicts, I merged the (long overdue) media3 migration earlier today. You'll probably need to update some imports as well when you merge with master.

Regarding the linter issues, some of them can really be taken as a hint and you can ignore them if it makes sense to you.

nielsvanvelzen:
@coderabbitai review

CeruleanRed:
https://github.com/coderabbitai review

CeruleanRed:
Bot seems satisfied. Let me know if I can still improve this in any way, I'm quite happy with it atm

app/src/main/java/org/jellyfin/mobile/utils/TickUtils.kt
```diff
@@ -0,0 +1,6 @@
+package org.jellyfin.mobile.utils
+
+object TickUtils {
+    fun ticksToMs(ticks: Long): Long = ticks / Constants.TICKS_PER_MS
+    fun msToTicks(ms: Long): Long = ms * Constants.TICKS_PER_MS
```
nielsvanvelzen, line 5: We already have utilities to work with ticks via our SDK & Kotlin's Duration type.
Instead of doing `TickUtils.ticksToMs(someValue)` you can do `someValue.ticks.inWholeMilliseconds`.
for `msToTicks` that would be `.milliseconds.inWholeTicks`

app/src/main/java/org/jellyfin/mobile/player/PlayerViewModel.kt
```diff
@@ -112,6 +116,10 @@ class PlayerViewModel(application: Application) : AndroidViewModel(application),
     val playerState: LiveData<Int> get() = _playerState
     val decoderType: LiveData<DecoderType> get() = _decoderType
 
+    // Chapter Markings
+    private var chapterMarkings: List<ChapterMarking> = listOf()
```
nielsvanvelzen, line 120:
```suggestion
    private var chapterMarkings: List<ChapterMarking> = emptyList()
```

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/mobile/player/PlayerViewModel.kt
CeruleanRed: Also ended up making the same change for PlayerMenus when switching to a map over a forEach

nielsvanvelzen (approved):

nielsvanvelzen:
Thanks for the contribution! If you're willing to open another PR to use the .ticks extensions in more places please do :)

PR #1648, opened 2025-05-20
Fix integrated player control under camera cutout

### LINKED ISSUE #1647: Player control(s) under camera cutout

### Describe the bug

On my OnePlus 9, using the Integrated player, the lock button is under the camera cutout.

Image

I see that in the latest update ([2.6.3](https://github.com/jellyfin/jellyfin-android/releases/tag/v2.6.3)) there were some inset changes by @Maxr1998 which pointed me in the right direction to test things out with my phone. I don't have much of Android app coding experience so could someone check out whether the following is an appropriate fix for it and maybe could be added in the next version?

Changing [this](https://github.com/jellyfin/jellyfin-android/pull/1593/files#diff-7655213a204c22650ed9fb014b5f881b382232de8953ca31299a70cfa780b209R161-R162) to the following:
```
playerControlsView.updatePadding(
     left = insets.displayCutout?.safeInsetLeft ?: systemInsets.left,
```

Results in this (_the video itself is still playing appropriately in full screen_):
Image

_As a side note, the Web player has the same issue but I have no idea how that could be fixed. I'm not using that though, so it's not really an issue for me._

### Logs

```shell

```

### Application version

2.6.3

Google Play Store

### Device information

OnePlus 9 Pro

### Android version

Android 14

### Jellyfin server version

10.10.7

### Other sources

_No response_

miikaforma:
**Changes**
Added a fix for some player controls going under the camera cutout in a full screen view when using the integrated player.

**Issues**

Fixes #1647


app/src/main/java/org/jellyfin/mobile/player/ui/PlayerFragment.kt
```diff
@@ -159,7 +159,7 @@ class PlayerFragment : Fragment(), BackPressInterceptor {
             if (playerFullscreenHelper.isFullscreen) {
                 playerView.setPadding(0)
                 playerControlsView.updatePadding(
-                    left = systemInsets.left,
+                    left = insets.displayCutout?.safeInsetLeft ?: systemInsets.left,
                     top = systemInsets.top,
                     right = systemInsets.right,
```
nielsvanvelzen, line 164: Would it make sense to do the same for the right side? Because if you turn your phone upside down the notch will be behind the right side.

Maxr1998: I also think making it as robust as possible on all devices would be the best, so we should check the cutout in all directions.

Though, I currently worry that the change could be a bit too cautious on some devices, and add insets that aren't really necessary and make the UI look off. So, I wonder if we could instead match the cutout with the actual controls and only apply the inset when any button overlaps. But this is obviously more complex and might not be worth the effort.

Maxr1998: Thanks for the additional changes! Instead of the coerceAtLeast with a static pixel value (which wouldn't scale across devices), could we just take the maximum of the two inset types? That should hopefully provide the best experience.

miikaforma: That should work and is probably the best solution. There's however a reason why I was experimenting with the static pixel value. For some reason, on OnePlus 9, the `systemInsets.top` returns unusually high value like `172` or something (I can't remember exactly what it was atm). I can't figure out why it's so big value. The `systemInsets.bottom` for example was just something like `72` and that would be fine for the top too.

Hard to show it in images since it doesn't show the borders and camera cutout but this is from the latest 2.6.3 version with the systemInset values (camera cutout over the bottom left lockscreen button).

image

And this is with the above `coerceAtLeast(32)` solution.

image

The toolbar going so far down was bugging me a bit so that's why I liked the `coerceAtLeast(32)` solution but I understand that `insets.displayCutout?.safeInsetTop?.coerceAtLeast(systemInsets.top) ?: systemInsets.top` would be a safer solution when considering all different devices. However, if we do it like that, then there's a question of whether to do it with the paddingBlock or should the top and bottom be separate? If it's paddingBlock, then the bottom will get the unusually big padding too on OnePlus 9.

miikaforma: Roger, so in conclusion the changes should be like this?

```
if (playerFullscreenHelper.isFullscreen) {
    playerView.setPadding(0)
    val left = max(insets.displayCutout?.safeInsetLeft ?: 0, systemInsets.left)
    val right = max(insets.displayCutout?.safeInsetRight ?: 0, systemInsets.right)
    val paddingInline = max(left, right)

    playerControlsView.updatePadding(
        left = paddingInline,
        top = max(insets.displayCutout?.safeInsetTop ?: 0, systemInsets.top),
        right = paddingInline,
        bottom = max(insets.displayCutout?.safeInsetBottom ?: 0, systemInsets.bottom),
    )
}
```

Or do you think left/right should be separate as well? Doesn't look bad whether it's separate or not in my opinion. Both work and it's just the matter of taste.

Maxr1998:
Thanks for the PR!

PR #1669, opened 2025-06-05
Update coil dependency from 2.7.0 to 3.2.0

cameocoder:
Update coil dependency from 2.7.0 to 3.2.0

**Changes**
Use coil to display the download thumbnail with error handling.  This fixes issue with bitmap decoding being done in the RecyclerView adapter.

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/downloads/DownloadsAdapter.kt
+++ b/app/src/main/java/org/jellyfin/mobile/downloads/DownloadsAdapter.kt
@@ -62,8 +62,8 @@ class DownloadsAdapter(
 
             val thumbnailUrl = getThumbnailUrl(context, downloadEntity.mediaSource.itemId)
             binding.imageViewThumbnail.load(thumbnailUrl) {
-                fallback(R.drawable.ic_local_movies_white_128)
-                error(R.drawable.ic_local_movies_white_128)
+                fallback(R.drawable.ic_local_movies_white_64)
+                error(R.drawable.ic_local_movies_white_64)
             }
         }
     }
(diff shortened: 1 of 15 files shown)


app/src/main/java/org/jellyfin/mobile/downloads/DownloadsAdapter.kt
```diff
@@ -56,7 +54,10 @@ class DownloadsAdapter(
                 true
             }
 
-            binding.imageViewThumbnail.setImageBitmap(downloadEntity.thumbnail)
+            binding.imageViewThumbnail.load(downloadEntity.thumbnailPath) {
+                fallback(R.drawable.ic_local_movies_white_128)
+                error(R.drawable.ic_local_movies_white_128)
```
Maxr1998, line 59: How does this fallback look in practice? I can imagine with a huge icon like that, it might feel a little odd..

cameocoder: Screenshot_20250605-220803
Screenshot_20250605-221206~2

cameocoder: The icons were pretty big in the first place.  The image view is just "wrap_content" and not bounded by size in any way except the size parameter on the download url.

Maxr1998: Ok, so this generally needs some rework. I guess that can be done in a later PR. For now, it looks ok.

Maxr1998 (dismissed):

app/src/main/java/org/jellyfin/mobile/downloads/DownloadsAdapter.kt
cameocoder: Screenshot_20250606-114000~2

Here is an updated screenshot based on the latest changes.

Maxr1998 (approved):
Thanks for the additional adjustments!
Re-reviewed and tested locally, works great and looks much better than before.

PR #1685, opened 2025-06-22
Add button for ask-to-skip media segments

CeruleanRed:
Adds a button to allow skipping for ask-to-skip media segments similar to the web player

_Ignore the quality, it's trying to software decode AV1 on the emulator_

**Changes**
Added the button in the PlayerFragment rather than the ExoPlayerControls so that it can be shown independently. It should work for multiple segments in the same file, but I haven't been able to test that yet. Works by using an update job every second to check weather to display the button or not, similarly to the progress and chapter marking jobs.

I also ended up refactoring how PlayerMenu data was being passed up to the PlayerViewModela as it felt a bit unnecessary to keep creating functions to pass stuff every time another piece was needed. I made a PlayerMenuHelper that is in essence just a wrapper for any sub-components that need to get passed up the view model, such as the skip button or chapter markings. This way we can pass the one object while still having decent separation of concerns.

**Issues**
Complements #1507 

**To Do**

- [x] Add translations
- [x] Test that it actually works for multiple segments (such as both an intro and an outro in the same video)
- [x] Style button in a better way (depends on how maintainers would like this styled)
- [x] Do not start update jobs if no media segments are present

### DIFF
--- a/app/src/main/res/drawable/skip_media_segment_background.xml
+++ b/app/src/main/res/drawable/skip_media_segment_background.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape xmlns:android="http://schemas.android.com/apk/res/android"
+    android:shape="rectangle">
+    <corners android:radius="6dp" />
+    <solid android:color="@color/skip_media_segment_background" />
+</shape>
--- a/app/src/main/res/layout/fragment_player.xml
+++ b/app/src/main/res/layout/fragment_player.xml
@@ -97,6 +97,9 @@
             android:layout_marginBottom="96dp"
             android:layout_gravity="end|bottom"
             android:visibility="gone"
+            android:background="@drawable/skip_media_segment_background"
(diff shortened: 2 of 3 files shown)


CeruleanRed:
What should the button look like? I just made the most basic thing I could think of, since I'm not sure what would generally be wanted or if there are some design guidelines to stick to

nielsvanvelzen:
> What should the button look like? I just made the most basic thing I could think of, since I'm not sure what would generally be wanted or if there are some design guidelines to stick to

We don't really have an established design in the mobile app yet. For now I'd say do something material design or what the Android TV app has.

CeruleanRed:
I didn't realize the android-tv app had a skip button. My implementation probably differs from that then. Do you think it would make more sense to try to copy what android-tv has instead of the current approach?

nielsvanvelzen (approved):

nielsvanvelzen:
@coderabbitai you may review this pull request

PR #1740, opened 2025-07-31
Remove direct dependencies on OkHttp/Okio

nielsvanvelzen:
We don't use either of them directly in the app and they're already kept up-to-date-ish via our SDK and other dependencies using them.

**Changes**
- Remove direct dependencies on OkHttp/Okio

**Issues**


Maxr1998 (approved):
I digged a bit through the code and it seems that they were used once when we were still manually fetching the webapp index and patching the HTML… which is clearly a thing of the past.

And I don't we'll think we need an HTTP client for other things anytime soon, so let's remove it.

PR #1824, opened 2025-10-27
Port video title format from web client

rodis120:
**Changes**
Reimplemented video title format used in [Jellyfin web client](https://github.com/jellyfin/jellyfin-web).
When using ExoPlayer or external player, Jellyfin shows only the episode/movie name on the title bar,
meanwhile web client shows series name, episode name,  season number, episode number and release year.
This PR adds title format used by web client to ExoPlayer and external player.

Before:
obraz

After:
obraz

After but with VLC as external player:
obraz

Web Client for reference:
obraz

Reference code from web client:
- https://github.com/jellyfin/jellyfin-web/blob/master/src/controllers/playback/video/index.js#L236-L270
- https://github.com/jellyfin/jellyfin-web/blob/master/src/components/itemHelper.js#L14-L64

**Issues**

I haven't found any related issues.

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/player/ui/PlayerFragment.kt
+++ b/app/src/main/java/org/jellyfin/mobile/player/ui/PlayerFragment.kt
@@ -122,7 +122,7 @@ class PlayerFragment : Fragment(), BackPressInterceptor {
             }
 
             // Update title and player menus
-            toolbar.title = mediaSource.name
+            toolbar.title = mediaSource.getName(requireContext())
             playerMenus?.onQueueItemChanged(mediaSource, viewModel.queueManager.hasNext())
         }
 
--- a/app/src/main/java/org/jellyfin/mobile/bridge/ExternalPlayer.kt
+++ b/app/src/main/java/org/jellyfin/mobile/bridge/ExternalPlayer.kt
@@ -138,7 +138,7 @@ class ExternalPlayer(
                 component = getComponent(appPreferences.externalPlayerApp)
             }
             setDataAndType(url.toUri(), "video/*")
-            putExtra("title", source.name)
+            putExtra("title", source.getName(context))
             putExtra("position", source.startTime.inWholeMilliseconds.toInt())
             putExtra("return_result", true)
             putExtra("secure_uri", true)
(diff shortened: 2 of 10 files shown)


app/src/main/java/org/jellyfin/mobile/JellyfinApplication.kt
```diff
@@ -14,8 +14,18 @@ import timber.log.Timber
...
+
+        @Volatile
+        private var instance: JellyfinApplication? = null
+
+        fun getInstance() = instance
+    }
+
     override fun onCreate() {
         super.onCreate()
+        instance = this
```
nielsvanvelzen: What is the purpose of this massive memory leak?

nielsvanvelzen: This does not solve the problem

rodis120: Ok, I removed that singleton.

nielsvanvelzen:
Going over the codepaths I think it makes the most sense to change the name property in JellyfinMediaSource to a `getName(context: Context)` function instead.

rodis120:
OK. I replaced `name` property with `getName(context: Context)`. However there are two places where the `Context` is inaccessible:

- https://github.com/jellyfin/jellyfin-android/pull/1824/files#diff-d0481b457ade47bd7ae9c103588fd3c04b31e55a6990d749900f1cefe3c16de5R31-L33
- https://github.com/jellyfin/jellyfin-android/pull/1824/files#diff-fe87cd78b85a00a960e948d14497a1c87edc650dc7d8a7e1d28e15077723432cR20-L22

Let me know If you are ok with the way I handled these cases.

app/src/main/java/org/jellyfin/mobile/player/ui/PlayerFragment.kt
```diff
@@ -122,7 +122,7 @@ class PlayerFragment : Fragment(), BackPressInterceptor {
             }
 
             // Update title and player menus
-            toolbar.title = mediaSource.name
+            toolbar.title = context?.let { mediaSource.getName(it) } ?: ""
```
nielsvanvelzen: use requireContext instead

nielsvanvelzen (changes requested):

rodis120:
@nielsvanvelzen I added the changes that you requested.

nielsvanvelzen (approved):

PR #1892, opened 2026-01-14
Fix Live TV stream end calls not releasing tuner on HDHR devices when used with Jellyfin Server 10.11.x

VTRunner:
**Changes**
In PlayerViewModel update reportPlaybackStart and reportPlaybackState to include liveStreamId in the call to the server.

Starting with Jellyfin Server v10.11.0 the server builds a list of liveStreamIds and tracks consumercount for each liveStreamId.  This list is built by the server based on an expectation that liveStreamId is being provided in the playback start and progress calls.  This was not a requirement prior to server v10.11.x, and if the liveStreamId is not in the server consumercount list when the LiveTV stream end call is made then the server does not release the HDHR tuner.

This issue is specific to when using the Integrated Player/Exoplayer.

**Issues**
[Jellyfin Server Issue 15769](https://github.com/jellyfin/jellyfin/issues/15769)


nielsvanvelzen (approved):

nielsvanvelzen:
Thanks!

Esslingers:
The suggestion to use the web player in Android isn't a full work around.  I tuned to a channel and saw Jellyfin grab 2 tuners.  When I exited the stream one of the tuners released but not the other

Issue #529, opened 2021-09-10
When switching servers, the language setting cannot follow the system


yyxida:
### Describe the bug

The system language is Chinese
1. Log in to server a (at this time, the app part is normally in Chinese, and some setting interfaces are in English, such as selecting the player interface)
2. Select switch server on the right
3. The URL input interface has changed to English
4. After logging in to server B, all UIs are in English
5. In the display setting, change the default language Auto to Chinese to display normally

### Application version

2.3.3

### Where did you install the app from?

Google Play

### Device information

sony xz1 , oneplus7tpro

### Android version

Android10(sony)/11(Oxygenos)

### Jellyfin server version

10.7.6

### Which video player implementations does this bug apply to?

- [ ] Web player (default)
- [ ] Integrated player (ExoPlayer)
- [ ] External player (VLC, mpv, MX Player)

BotBlake:
@yyxida Is this issue still happening for you?
It should be resolved due to changes in jf-web.

Kind regards

PR #1896, opened 2026-01-15
Fix the BUG in localized language settings

SJJ-dot:
**Changes**
Fix the BUG in localized language settings

**Issues**
#529


nielsvanvelzen (changes requested):
If we just use kotlinx.serialization we don't need the additional checks. Something like `Json.decodeFromString<String?>(result)` should work fine for parsing both `null` and actual strings.

nielsvanvelzen:
Superseded by #2028

PR #1906, opened 2026-02-07
Fix SSO/OIDC login loop by persisting _deviceId2

nalet:
When using plugin-sso OIDC, the `/sso/OID/redirect/...` page expects `_deviceId2` to exist in `localStorage`.

In the Android native shell flow, the device id comes from native code but was not persisted under this key, which caused an infinite "Logging in..." loop.

This change ensures `_deviceId2` is set from the native `deviceId` during `NativeShell.AppHost.init()` when missing.

Related: 9p4/jellyfin-plugin-sso#189


nielsvanvelzen:
We do not support third party plugins in our official apps. The plugin should fix this on their side.

PR #1914, opened 2026-02-23
Edited headers layout to become full screen and adjust positions

AndyXheli:
This PR updates the header layout to achieve a full-screen background effect while ensuring interactive elements are correctly positioned.

### **Key Changes:**

- Edge-to-Edge Background: Updated WebViewFragment to allow the WebView to extend to the very top of the screen (behind the status bar).
- Content Positioning: Injected custom CSS via injectionScript.js to add top padding specifically to the interactive icons (hamburger menu, search, cast, etc.), preventing overlap with system status bar icons.
- New Utility: Added applyWindowInsetsAsPadding to UIExtensions.kt for better inset handling.

### **Testing:**

- Verified on device that the background covers the entire top area.
- Confirmed icons are pushed down and clearly visible below the status bar.

image


nielsvanvelzen (changes requested):
The app is not supposed to inject custom CSS into the web client as this is very prone to breakage across different jellyfin-web versions. Those changes will need to be upstreamed instead, in a way that works for all devices.

nielsvanvelzen:
Closing as stale

Issue #903, opened 2022-11-23
Android App Links


PhilippMundhenk:
It would be great if it were possible to link directly to the Jellyfin app (or parts thereof, like a single collection or item). 

I use Heimdall as a portal and would like to provide a direct link to the Jellyfin app, potentially even separate links to separate collections for my users. 

Android App Links should enable this, though some coordination with the Jellyfin website team owning the URL might be required to register domain ownership with Google. 

Example intent-filter for Main activity:
```
<intent-filter android:autoVerify="true">
    <action android:name="android.intent.action.VIEW" />
    <category android:name="android.intent.category.DEFAULT" />
    <category android:name="android.intent.category.BROWSABLE" />

    <data android:scheme="http" />
    <data android:scheme="https" />

    <data android:host="android.app.jellyfin.org" />
</intent-filter>

```

Potentially relates to #694

nielsvanvelzen:
I was thinking we could do something similar to https://my.home-assistant.io for Jellyfin. The apps can open these links directly and for the web client it can store a server in localstorage to redirect. We shouldn't create app links that only work for one specific client so it needs to be a organization wide project with proper documentation.

Another related issue is jellyfin/jellyfin-meta#28

Farbklex:
App links are supposed to work for one verified URL. Jellyfin instances are hosted on many URLs. So it won't be possible to allow the jellyfin android app to open HTTP(S) jellyfin links for self hosted instances.

So we either need to use an URI with `jellyfin://` as described in https://github.com/jellyfin/jellyfin-meta/issues/28 or we need a service like https://my.home-assistant.io/ as described by niels.

Something like `https://my.jeyylfin.org/farbklex/web/index.html#!/details?id=81d46b8908ba9f52813b9a875ddb6afa&serverId=8003032799c742b8bb05793cae8a103c` would then link to farbklex' instance which can be opened via the app or browser.

iampegram:
I don't see why the first solution can't work since all we want is the Jellyfin app to open. What the app does with the URL is arbitrary; the devs could ignore it and present the login screen for whatever servers are already configured in the app, or the app could be hard-coded to replace the generic domain with the domain of the current logged in server.

Anzic23:
Hey, I've been running into this exact issue — needed to launch Jellyfin from a custom dashboard via URL.

[...]

PR #1933, opened 2026-03-18
feat: Add custom URI scheme for deep linking support

Anzic23:
## Summary
Adds a custom URI scheme `jellyfin://` to MainActivity, enabling deep linking  from third-party launchers, dashboards, and browsers.

## Motivation
Currently there is no way to launch the Jellyfin Android app via a URL from  an external app or browser. Other media apps (Spotify, YouTube, Telegram, etc.)  support URI schemes which allow dashboards and launchers to open them directly.

This is also related to #903 (Android App Links).

## Changes
Added an `intent-filter` to `MainActivity` in `AndroidManifest.xml`:
```xml
<intent-filter>
    <action android:name="android.intent.action.VIEW"/>
    <category android:name="android.intent.category.DEFAULT"/>
    <category android:name="android.intent.category.BROWSABLE"/>
    <data android:scheme="jellyfin"/>
</intent-filter>
```

## Testing
Tested on Android tablet — navigating to `jellyfin://` from Chrome successfully  launches the app.

**Changes**

**Issues**


nielsvanvelzen:
As mentioned before, we're not against adding some kind of scheme to launch our clients but it needs to be consistent across all of them. For that we really need a proposal first (in the jellyfin-meta repository) that is agreed upon by our client developer team before implementations can happen.
Therefor I will close this for now.

Issue #677, opened 2022-05-03
Show amount of buffered video on the progress bar


SuperPrower:
**Describe the feature you'd like**
I stream over somewhat slow connection, so I would like to have some media buffered in advance. As I understand, this already occurs. However, unlike Jellyfin Web, in the standard player, I don't see how much time was buffered/preloaded/precached on the progress/playback bar.

**Additional context**
Here's how it currently looks in Jellyfin Web:
![Screenshot_20220503_171914](https://user-images.githubusercontent.com/10908110/166483119-016e87df-9374-48bd-9f7f-1d4aa2e12371.png)
Unless something is wrong with my client (version 2.4.4. from Google Play, but I don't mind installing experimental builds), I don't believe that this is visually present in the current standard player (video player type -> web player) in the android version.

susonthapa:
@SuperPrower I checked the latest version (2.4.4) and it seems to be working fine. Let me know if I misunderstood something.

https://user-images.githubusercontent.com/33973551/201458156-67e5f7ef-cbd9-4bf6-8e4d-2c9654c8fffc.mp4

SuperPrower:
@susonthapa thanks for your reply. I double checked, and it doesn't seem to be working in the Web Player (the one, I think, you have in your screen capture):

https://user-images.githubusercontent.com/10908110/201467938-b09f3099-a5b4-470a-a9b4-cfcc9fd9ca8f.mp4

It seems to be working in the ExoPlayer-based player - I quickly see buffer after pause, so this isn't an issue with connection, and the feature itself seems to be working. Maybe this is some kind of visual bug with the player? My device is Pixel 5, latest OS/updates.

susonthapa:
@SuperPrower Looking at the attached video and the color of the progress bar it looks like it is already buffered which really doesn't make sense for a video that long. I think the buffering might be broken in this case. The standard player is just the Jellyfin Web UI loaded in a WebView(you can think of it as an embedded browser). The WebView is tied to the OS and can cause some weird issues like [this](https://github.com/jellyfin/jellyfin-android/issues/853) one. Either way, I don't think we can do anything here as the UI that gets rendered([Jellyfin-web](https://github.com/jellyfin/jellyfin-web)) and the component that renders the UI (WebView) are out of our control.

susonthapa:
@SuperPrower Feel free to close this issue.

SuperPrower:
> Feel free to close this issue.

[...]

Issue #564, opened 2021-11-14
DirectPlay a > 1Mbps video stream even if Internet streaming limit set to 1 Mbps.


Elaws:
### Describe the bug

1. Set Internet streaming limit to 1 Mbs.
2. Jellyfin should transcode if video stream exceeds 1 Mbps.
3. Jellyfin does not transcode, and direcplays the video. Thus, video is constantly stuttering.

### Application version

2.4.0

### Where did you install the app from?

Google Play

### Device information

Samsung Galaxy S20FE 5G

### Android version

Android 11

### Jellyfin server version

10.7.7

### Which video player implementations does this bug apply to?

- [ ] Web player (default)
- [X] Integrated player (ExoPlayer)
- [ ] External player (VLC, mpv, MX Player)

Elaws:
Well, it seems the problem happens only with ExoPlayer, not with web player. 

For unknown reasons, any internet speed limit or quality settings that you can set in the app is absolutely not at all taken into account when using ExoPlayer. No transcoding occurs when you set limits.

Maxr1998:
Yes, known issue, I mentioned it the other day in #563. The tracking issue for this is also #24 (in addition to user-selectable bitrate limits), some work has already been done on  #535.

naimo:
By the way, the exoplayer (and I believe the external player option as well, with VLC for example) also ignores the server "Internet streaming bitrate limit"
If that helps

galmok:
This issue is still a problem and it is exacerbated by the fact the the web player doesn't work (freeze frame unless constantly touched). As such, there exists no player on Android with bitrate limiting (transcoding) applied.

Elaws:
I can confirm that the issue is still happening. Has there been any progress concerning this issue reported 17 months ago ?

extrange:
This is still happening with Jellyfin on Android, with both the integrated player and external player (VLC).

Maxr1998:
ExoPlayer now supports manual bitrate limits. The user provided limits will be supported in the future.

shaino:
Actually, it's not difficult.
in
app/src/main/java/org/jellyfin/mobile/player/interaction/playoptions.kt
line 134,add code... for example:
viewModal.changeBitrate(6000000)

But
To be precise, it is necessary to obtain locally set parameters to assist.

shaino:
If the developer has not yet completed it, they can use it urgently:
step 1 
In file: 
[...]

Issue #39, opened 2020-08-23
Playing music stutters when the app is in the background


Seil0:
When i'm paying a song with jellyfin open in the background the song start to stutter (sounds like the song stops for a few milliseconds). As soon as i bring jellyfin to the foreground the song continues to play without any issues. Changing between background and  foreground always results in this behavior. 

Device: Pixel 2
OS: Android 10 (QQ3A.200805.001)
Jellyfin App: 2.0.0-rc2
Jellyfin Server: 10.6.2

Maxr1998:
Yeah, that's an issue with playing media in WebView. I couldn't reproduce it on my device, but it's definitely a thing. Long-term, we want to switch to a media player integrated into the OS.

LeRaldo:
This exact issue started happening to me after my phone upgraded to Android 11 today. It was working flawlessly before when I was on Android 10.

Device: OnePlus 8 Pro (IN2025)
OS: Oxygen OS 11.IN11AA (Android 11)
Jellyfin App: 2.0.1
Jellyfin Server: 10.6.4

aschubert:
Same issue here with a Amazon Fire HD 10 tablet. Strange thing as a side note: as soon as I connect a bluetooth speaker to the tablet, no dropouts/stuttering sound happens. On other Android devices (MIUI 12) no issue at all.

jariha:
I have the same issues with Nokia 6.2. I have tested to disable the battery optimization but no help there.

dkanada:
For now, the suggestion is to use another Jellyfin client that isn't in a web view. At some point music playback will be handled natively at which point these issues should disappear. I suggest either Gelli or Yatse for music playback.

Maxr1998:
+1 for Gelli.

The comment I made [here](https://github.com/jellyfin/jellyfin-android/issues/222#issuecomment-726865105) applies to this issue as well, that's why I marked it as `wontfix`.

alexs20:
-1 for Gelli. It has track sorting issues and author has no time to fix it
 It is useless for albums where order is important, for example for audiobooks.

yotoprules:
> -1 for Gelli. It has track sorting issues and author has no time to fix it
> It is useless for albums where order is important, for example for audiobooks.

that and it keeps causing my whole Jellyfin server to crash: seems to happen when I'm driving which suggests if there's an unstable connection it just screws everything up. Might go back to emby for music for now until there's a usable android client.

dkanada:
[...]

PR #1946, opened 2026-04-05
Add configurable playback buffer size for integrated player

GeiserX:
**Changes**

The integrated player (Media3/ExoPlayer) uses default `DefaultLoadControl` time thresholds of 50s min / 50s max buffer, with 1s to start playback and 2s to resume after rebuffer. While these defaults work well for stable connections, users on slow, variable, or high-latency connections (e.g. satellite internet like Starlink) experience frequent rebuffering because the max buffer ceiling is the same as the min — there is no room for the player to build a deeper buffer ahead of variable bandwidth dips.

This PR adds a **"Playback buffer size"** setting under the Video player section with three presets:

| Preset | Min buffer | Max buffer | Playback start | After rebuffer |
|--------|-----------|-----------|----------------|----------------|
| **Auto** (default) | 50s | 50s | 1s | 2s |
| **Large** | 50s | 120s | 2.5s | 5s |
| **Extra large** | 80s | 240s | 5s | 10s |

Key design decisions:

- **Only time-based thresholds are changed** (`setBufferDurationsMs`). The byte-based buffer target (`targetBufferBytes`) is left at `C.LENGTH_UNSET`, preserving Media3's automatic track-aware sizing (~125–138 MB for remote video, ~20 MB for local playback). This means on high-bitrate direct play, the byte target will naturally cap the buffer before the time target is reached — providing a safe memory bound without any custom byte overrides.
- **No preset lowers any threshold below the current Media3 1.8.0 defaults** (50s/50s/1s/2s). Large keeps the same 50s minimum and raises only the ceiling and startup thresholds. Extra large raises both.
- **The higher `bufferForPlaybackMs` / `bufferForPlaybackAfterRebufferMs` values** in the non-default presets are intentional: on unreliable connections, waiting 2.5–5s before starting (vs 1s) significantly reduces the chance of an immediate rebuffer after playback begins.
- The setting applies to the shared ExoPlayer instance, so it technically affects both streaming and local playback. For local/downloaded files, the likely user-visible impact is negligible since disk reads fill the buffer far faster than any time threshold, but the startup/rebuffer thresholds are technically altered.

The setting is gated behind the integrated player type selection (disabled for web/external players) and follows the existing preference pattern used by other ExoPlayer settings.

**Scope:** 5 files, 44 lines added — `Constants.kt`, `AppPreferences.kt`, `strings.xml`, `SettingsFragment.kt`, `PlayerViewModel.kt`. No architectural changes; uses the existing `DefaultLoadControl.Builder` API from Media3.

**Iteration notes:** The initial revision used `setPrioritizeTimeOverSizeThresholds(true)` to ensure time targets were honored, but this was identified as an OOM risk on high-bitrate direct play (uncapped memory growth, ref: google/ExoPlayer#7354). A second revision tried explicit `setTargetBufferBytes()` caps, but this overrode Media3's track-aware local/remote distinction and misstated the default byte targets. The current revision takes the simplest approach: change only the time thresholds and let Media3 handle all memory management via its existing defaults.

**Research:** I reviewed all open issues and found no prior feature request for user-configurable buffer settings. Related issues for context:
- #677 requests a visual buffer indicator on the seekbar (complementary UI enhancement, not overlapping)
- #564 reports bitrate limits being ignored by ExoPlayer (different mechanism, same symptom area)
- #39 covers background music stuttering (WebView-related, different root cause)

None of the above address the ability for users to tune buffer sizes for their connection quality.

**Testing:** Screenshots of the setting on a physical device will follow in a comment once the APK is built and sideloaded.

**Issues**

No existing issue — this addresses a gap identified through researching buffering-related reports across the Jellyfin ecosystem.


GeiserX:
FYI, obviously this was all AI-driven, this comment is not. I tried to be a small change as possible in the diff, so it's easily reviewable for you. I acknowledge you will be receiving low-q submissions all day so don't want to make your job more difficult. I'm the dev behind smart-covers and whisper-subs plugins. This was brought to my attention when a user using Starlink was really bothered by using both Jellyfin and Plex servers, and using Plex is objectively better, so I decided to help him out and propose to use other clients like kodi + jellyfin kodi addon or jellyfin-mpv-shim with a higher buffering set. But I'd like to have a setting in the client so you can control this. I tried to reduce OOM surface with the force-pushes (putting opus 4.6-max against gpt-5.4-xhigh repeatedly) and iterating over and over, and that's the changes you see. Thanks for your hard work.

Happy to trigger more discussions around this (i.e. if anyone thinks a slider would be preferable over the current solution. Check next comment with the screenshots.)

GeiserX:
**On-device testing**

Built and sideloaded a libre debug APK on a Samsung device. The setting works as expected under **Client Settings** when the integrated player is selected.

**Settings dialog** — the three presets (Auto / Large / Extra large) with descriptions:

Playback buffer size setting

**Integrated player** — playback running with the buffer configuration set to EXTRA-LARGE:

Integrated player playback

GeiserX:
Companion PR for the Android TV client: jellyfin/jellyfin-androidtv#5515 — same approach (preset-based buffer size selection), adapted to the androidtv architecture (Compose UI, separate playback module).

Maxr1998 (approved):
Looks good go me, as long as you fix the detekt issues. Just running the detekt gradle task might be enough already, iirc it'll auto fix the issues.

GeiserX:
> Looks good go me, as long as you fix the detekt issues. Just running the detekt gradle task might be enough already, iirc it'll auto fix the issues.

Thanks, done @Maxr1998

PR #1954, opened 2026-04-17
Fix a memory leak inside ServiceBinder

stumpylog:
**Changes**

While playing around with trying to get my car head unit to display album artwork over Bluetooth, the debug build popped up a notification about a leak, which reported ~4.1 MB across 114 objects.  ServiceBinder held a strong reference to its outer RemotePlayerService instance. This replaces the strong reference with a WeakReference so the service can be collected normally after destruction.

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/webapp/RemotePlayerService.kt
+++ b/app/src/main/java/org/jellyfin/mobile/webapp/RemotePlayerService.kt
@@ -468,8 +468,9 @@ class RemotePlayerService : Service(), CoroutineScope {
         super.onDestroy()
     }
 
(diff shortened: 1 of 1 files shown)


nielsvanvelzen:
I don't think using a WeakReference here makes much sense, the binder should never be able to outlive the service. If we were to make changes here I'd make the binder an inner class so it doesn't need the service passed to its constructor.

And after looking into the Android documentation it seems like their code sample does exactly that: https://developer.android.com/develop/background-work/services/bound-services

Keep in mind that LeakCanary often gives false positives, it's just a tool.

stumpylog:
Ah, I assumed that the canary thing was always right.  This seemed to fix it, but if it's just wrong, no reason to do anything.

PR #1973, opened 2026-05-09
Fix bluetooth media keys for local video playback

### LINKED ISSUE #1130: MediaButton events in Jellyfin Web Player not working
### Describe the bug

1. Play a video using the Web Player
2. Press a bluetooth headset button
3. The player should pause but it doesn't

This works correctly with the media implementation

### Logs

_No response_

### Application version

2.5.2

Sideloaded APK (proprietary build)

### Device information

Pixel 7

### Android version

13

### Jellyfin server version

10.8.10

- [X] Web player (default)

### LINKED ISSUE #1707: Bluetooth play/pause button does not pause

### Describe the bug

When playing media with the Jellyfin Android client, pressing the play/pause button on my blutooth headset does not pause.  The button does work to start play.  The button can pause just fine in other media players, including Jellyfin Web running in Fennec.

### Logs

```shell

```

### Application version

2.6.3

F-Droid

### Device information

Google Pixel 9 Pro  +  Shokz OpenRun S803

### Android version

Android 15

### Jellyfin server version

10.10.7

### Other sources

_No response_

mherceg:
For local web video playback, jellyfin-web's MediaSessionSubscriber returns early in onMediaSessionUpdate when isLocalPlayer && isVideo and never calls shell.updateMediaSession. As a result the native RemotePlayerService MediaSession is never activated and has no PlaybackState. Hardware media keys (bluetooth headset clicks, etc.) still reach onMediaButtonEvent, but the framework's default implementation has no actions to translate the key against and silently drops the event.

**Changes**
Override onMediaButtonEvent to map media key codes directly to JS playbackManager commands via the existing webappFunctionChannel bridge, bypassing the dependency on PlaybackState. KEYCODE_MEDIA_PLAY/ PAUSE/NEXT/PREVIOUS/STOP/FAST_FORWARD/REWIND map to their respective commands; KEYCODE_HEADSETHOOK and KEYCODE_MEDIA_PLAY_PAUSE map to a new playPause command that toggles via the JS-side authoritative state.

**Issues**
Fixes #1707
Fixes #1130


app/src/main/java/org/jellyfin/mobile/webapp/RemotePlayerService.kt
```diff
@@ -412,6 +414,29 @@ class RemotePlayerService : Service(), CoroutineScope {
...
                 object : MediaSession.Callback() {
+                    // Hardware media keys (e.g. bluetooth headset clicks) reach this
+                    // callback even when the session is inactive and has no PlaybackState.
+                    // For local web video playback the JS-side MediaSessionSubscriber
+                    // skips updating the native session, so the framework's default
+                    // implementation has no actions to translate the key against and
+                    // silently drops the event. Route the keys to the JS playbackManager
+                    // directly to avoid that dependency.
+                    override fun onMediaButtonEvent(mediaButtonIntent: Intent): Boolean {
+                        @Suppress("DEPRECATION")
```
nielsvanvelzen: Fix this deprecation instead of suppressing it.

app/src/main/java/org/jellyfin/mobile/webapp/RemotePlayerService.kt
```diff
@@ -412,6 +414,29 @@ class RemotePlayerService : Service(), CoroutineScope {
...
                 @SuppressLint("MissingOnPlayFromSearch")
                 object : MediaSession.Callback() {
+                    // Hardware media keys (e.g. bluetooth headset clicks) reach this
+                    // callback even when the session is inactive and has no PlaybackState.
+                    // For local web video playback the JS-side MediaSessionSubscriber
+                    // skips updating the native session, so the framework's default
+                    // implementation has no actions to translate the key against and
+                    // silently drops the event. Route the keys to the JS playbackManager
+                    // directly to avoid that dependency.
+                    override fun onMediaButtonEvent(mediaButtonIntent: Intent): Boolean {
```
nielsvanvelzen: Move this fun to the end for better readability. Also I don't think we need half an essay explaining why it exists. It could be made shorter.

app/src/main/java/org/jellyfin/mobile/webapp/RemotePlayerService.kt
```diff
@@ -472,4 +497,20 @@ class RemotePlayerService : Service(), CoroutineScope {
         val isPlaying: Boolean
             get() = service.playbackState?.state == PlaybackState.STATE_PLAYING
     }
+
+    private companion object {
+        fun commandForKeyCode(keyCode: Int): String? = when (keyCode) {
```
nielsvanvelzen: Not a fan of adding an entire companion object for a little when statement here. It could be inlined or written differently instead.

nielsvanvelzen (changes requested):

nielsvanvelzen (approved):

PR #1977, opened 2026-05-11
Fix Chromecast controls disappearing immediately after casting

### LINKED ISSUE #459: Chromecast controls disappear immediately

### LINKED ISSUE #464: Jellyfin loses control of Chromecast while it is playing
**Describe the bug**

After media is playing to Chromecast for a while, the remote control becomes inoperative. It's almost as if it detaches from the Chromecast. Disconnecting, then clicking the Cast icon and selecting the same device does nothing until the app is stopped and restarted.

**To Reproduce**

1. Start playing media via Chromecast
2. Wait a while? Switch apps, I dunno exactly. Maybe even close Jellyfin app and reopen it.
3. Attempt to control the media with the Chromecast remote in Jellyfin.

**Expected behavior**

Jellyfin should never lose control of the media it is casting.

**Logs**

I do not have adb or root to capture logs, but, let me know if you really need the logs and I'll see what I can do when I have time.

**Screenshots**

Screenshot_20210710-230944

 - App version: 2.2.3
 - Android version: 10
 - Device manufacturer and model: Google Pixel 4
 - Jellyfin server version: 10.7.6
 - The type of video player implementation used: Web player, I guess?

**Additional context**

None that I can think of.

sebaplaza:
## Summary

- Initialize `queueItems` cache to empty JSONArray instead of null
- Skip null media updates in `Chromecast.java` instead of sending empty MEDIA_UPDATE events
- Guard against falsy media in `chrome.cast.js` MEDIA_UPDATE handler

## Root Cause

- `ChromecastUtilities.queueItems` was declared as `null`. 
- When the first `onStatusUpdated()` fires before queue data is available, `createMediaObject()` calls `out.put("items", null)` which throws a `NullPointerException`. 
- The catch block silently returns null, causing an empty `MEDIA_UPDATE` event. 
- The JS handler interprets missing media as "media is no longer alive" (`isAlive=false`), so Jellyfin Web immediately hides the cast controls.

## Changes

- **`ChromecastUtilities.java`** Initialize `queueItems` to `new JSONArray()` instead of `null`.
- Prevents NullPointerException when `createMediaObject()` is called before queue data arrives.
**`Chromecast.java`** Return early when `jsonMedia` is null instead of sending an empty event.
- **`chrome.cast.js`** Return early when `media` is falsy (belt-and-suspenders, consistent with `SESSION_UPDATE` handler which already guards `if (!obj)`).

Fixes #459 

Fixes #464


nielsvanvelzen:
Is this actually tested or just pure AI output?

PR #1980, opened 2026-05-13
Replace loadUrl("javascript:...") with evaluateJavascript in WebViewFragment

### LINKED ISSUE #1979: WebViewFragment uses legacy loadUrl("javascript:...") for dispatching JS function calls
`app/src/main/java/org/jellyfin/mobile/webapp/WebViewFragment.kt` dispatches JS function calls coming from `webappFunctionChannel` (used by `PlayerActivity` and other components) using the legacy form:

```kotlin
// Process JS functions called from other components (e.g. the PlayerActivity)
lifecycleScope.launch {
    for (function in webappFunctionChannel) {
        webView.loadUrl("javascript:$function")
    }
}
```

`loadUrl("javascript:...")` was deprecated in Android 4.4 in favour of `WebView.evaluateJavascript`. The Android docs explicitly recommend `evaluateJavascript` for this case because:

- The legacy form adds a navigation entry to the WebView back/forward history on every invocation, which surfaces as confusing back-button behaviour when a player or another component pushes frequent JS calls.
- It cannot return a value to the host (not load-bearing here, but the reason the platform moved on).

### Suggested fix

```kotlin
webView.evaluateJavascript(function, null)
```

Same JS payload, no back-stack pollution.

A PR is open at #1980.

jim-daf:
**Changes**

Swap the legacy `loadUrl("javascript:...")` form to `webView.evaluateJavascript(function, null)` in `WebViewFragment`, where JS function calls from `webappFunctionChannel` are dispatched to the web client. The Android docs deprecated the `loadUrl` javascript URI form in 4.4 because it adds a navigation entry to the WebView back/forward history on every call. `evaluateJavascript` does not. Semantics are identical for the existing callers, since none of them consume a return value.

**Code assistance**

An LLM helped draft the body in this PR on top of a static-analysis report I am running across a set of Android apps. The actual change is a one-line swap I have read through and understand.

**Issues**

Fixes #1979

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/webapp/WebViewFragment.kt
+++ b/app/src/main/java/org/jellyfin/mobile/webapp/WebViewFragment.kt
@@ -162,7 +162,7 @@ class WebViewFragment : Fragment(), BackPressInterceptor, JellyfinWebChromeClien
         // Process JS functions called from other components (e.g. the PlayerActivity)
         lifecycleScope.launch {
             for (function in webappFunctionChannel) {
-                webView.loadUrl("javascript:$function")
+                webView.evaluateJavascript(function, null)
             }
         }
     }


Maxr1998 (approved):
Seems good. We also use evaluateJavascript elsewhere already. Unsure why this wasn't migrated back then.

nielsvanvelzen (approved):
Changes make sense, please fill in the PR template as requested so we can proceed with merging.

jim-daf:
> Changes make sense, please fill in the PR template as requested so we can proceed with merging.

Hi @nielsvanvelzen i have filled in the PR template as requested

PR #1983, opened 2026-05-15
Change UserEntity.userId to UUID

nielsvanvelzen:
**Changes**

- Change the userId property in the user entity to be an UUID
- Add migration that changes all users in the database to use dash formatted uuids (instead of dash-less)
- Add custom UUID converter to room
- Fix some variable names in UserDao (id != userId)
- Update language getting to convert uuid into dash-less
- Rename database tables to lower_snake_case

Note: This does NOT modify the download entity as it will be changed in a future PR

**Code assistance**

**Issues**

### DIFF
--- a/app/schemas/org.jellyfin.mobile.data.JellyfinDatabase/4.json
+++ b/app/schemas/org.jellyfin.mobile.data.JellyfinDatabase/4.json
@@ -1,0 +1,155 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 4,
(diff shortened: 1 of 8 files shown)


Maxr1998 (approved):
Not too fond of the lowercase table names, but no strong objection either. Code looks good.

nielsvanvelzen:
I've always used lower_snake_case for both database tables and columns, so when I was working on the string->UUID change I opted to also make that change here.

I don't think there really is a common convention for it, looking at the androidx.room documentation they use a mix of camelCase, PascalCase and lower_snake_case and also plural/non-plural.

PR #1986, opened 2026-05-17
Add fallback icons to Android Auto items

nielsvanvelzen:
**Changes**

Add the monochrome Jellyfin logo as fallback image to Android Auto items when the item has no artwork of its own. This prevents Android Auto from showing a big ⚠️ icon in the grids.

Additionally add icons for all top-level categories. Material icons used (reused from ATV where possible for consistency).

afbeelding
afbeelding

**Code assistance**

**Issues**


app/src/main/res/drawable/ic_artist.xml
```diff
@@ -0,0 +1,10 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:tint="#FFFFFF"
```
Maxr1998, line 4: Probably unnecessary. Colors are also inconsistent across icons.

Maxr1998 (approved):
Please fix the inconsistencies with the new icons, otherwise this can be merged.

app/src/main/res/drawable/ic_artist.xml
nielsvanvelzen: Yeah looks like some icons in the ATV repo were inconsistent because the icons that were wrong were copied over.
We want to set fillColor to the `@android:color/white` and tint on the vector to `#FFFFFF`. Don't ask me why that has become the default but I'll stick with that for now 😅

Maxr1998: Oh, ok 😅 The tint is definitely redundant, but it's fine.

PR #2006, opened 2026-05-31
Direct play Live TV MPEG-TS streams

lollinoo:
## Summary

- Advertise ExoPlayer MPEG-TS/TS container support with matching video and audio codec profiles.
- Resolve Live TV `TvChannel` playback info without forcing `MediaSourceId=itemId`, then prefer direct-playable HTTP MPEG-TS sources.
- Route direct-play HTTP TS URLs through a progressive media source and support external-player Live TV stream lifecycle cleanup.

## Root Cause

Live TV channels  can return no playable media sources when the Android client sends a generated `MediaSourceId` based on the item id. When a source is opened, the server can also mark supported HTTP TS streams as `DirectStream` even with no transcode reasons, which makes the app use Jellyfin's remux URL instead of the original transport stream.

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/app/AppModule.kt
+++ b/app/src/main/java/org/jellyfin/mobile/app/AppModule.kt
@@ -149,9 +149,9 @@ val applicationModule = module {
             }
     }
 
-    single<MediaSource.Factory> {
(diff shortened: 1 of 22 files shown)


nielsvanvelzen:

PR #2009, opened 2026-06-03
Enable native app languages settings for jellyfin

ClydeFrog04:
**Changes**
The jellyfin mobile app sets it's language based on the users system default, which may not always be their preferred app language preference. This change enables users to set the jellyfin app language from the native android App Languages settings page

**Code assistance**
Claude was used to generate the list of languages for the locales.xml file when prompted with the list of values from `ls app/src/main/res/`

**Issues**
No issue was opened as this was not a bug, nor did it feel big enough to open a new feature request for, happy to create the issue if needed!


nielsvanvelzen:

ClydeFrog04:
Ahh fair, can I hand type it all out and open a new pr for this?

nielsvanvelzen:
Or maybe just test the master branch first before opening an AI generated PR :melting_face:

ClydeFrog04:
I was only trying to contribute, and i totally understand and respect you decision on blocking ai contributions. I won't put any further pr's in for jellyfin- thank you for the replies.

nielsvanvelzen:
Even if this PR wasn't AI generated it didn't accomplish anything. We've already enabled app language support 8 months ago without the need to maintain an XML file.

ClydeFrog04:
Totally fair on that as well yes. The version in the play store that i was able to access does not have app language support though which is why i did not think it was

PR #1984, opened 2026-05-16
Add audiobook support to Android Auto

nielsvanvelzen:
**Changes**

- Add support for book libraries to Android Auto.
- Add parent image fallback in baseitem->mediaitem mapping

Unfortunately I wasn't able to thoroughly test these changes. The only access to an audiobook library I have is limited and afaik the server support for this functionality is poor. There might be some query issues. Feedback appreciated.

**Code assistance**

**Issues**

Fixes #247

Issue #314, opened 2021-03-02
Make audiobooks available for Android Auto


teacupx:
**Describe the feature you'd like**
Right now, only collections of type "Music" are shown in Android Auto. Showing also "Books" collections would be very useful for people who listen to audiobooks on their way to work or their travels.

**Additional context**
Audiobooks collections are meant to be organized in folders, instead of browsing through artists/albums. This needs to be taken into account.

nielsvanvelzen:
Duplicate of #247

teacupx:
Sorry, I did a search but that one didn't show up :)

Issue #247, opened 2020-11-21
Audio books playback in Android Auto support


jcdick1:
**Describe the feature you'd like**
Add the Books library type to the Android Auto support, to allow playback of audio books

nielsvanvelzen:
We need to add the "Books" type here:
https://github.com/jellyfin/jellyfin-android/blob/master/app/src/main/java/org/jellyfin/mobile/media/car/LibraryBrowser.kt#L187

I'm not sure if that's enough to get this functionality working so some testing needs to be done for that.

jcdick1:
No real hurry, I just wanted to get it on a list somewhere for when there's time.  

My next question to go with this would be to know if automatically playing the next track is "pushed" as a server-side function or "pulled" as a client side, since all of my audio content is episodic, or file-per-chapter type, rather than one file per title.  So I know where to put in a feature request for that support.

nielsvanvelzen:
Playlists are partially clientside and partially serverside. There is a feature request for proper audiobook here: https://features.jellyfin.org/posts/243/audiobook-support

jcdick1:
Oh, I didn't realize that when the client plays the next episode of a TV series, it was basically building a playlist in the background.  I though either the server was telling the client "Hey, there's another episode in this series" or the client was asking "Is there another episode?  Cool, I'll queue it up."

That's basically all I'm after for my audio, since it is basically episodic.  Show->season->episode just without moving pictures.

Kitof:
I'm very interested in this enhancement too. Any news ?

teacupx:
I tried to do the simple change of this line: https://github.com/jellyfin/jellyfin-android/blob/43a734f281b398eeb02680b92d3e225fffddb80c/app/src/main/java/org/jellyfin/mobile/media/car/LibraryBrowser.kt#L206
Changing it for:
`.filter { item -> (item.collectionType == CollectionType.Music) || (item.collectionType == CollectionType.Books) }`

That was enough to make the libraries marked as "Books" to show up in Android Auto. But browsing them was a mess, since the app expected them to be browsable by tags, instead of folders like "Books" collections are organized. As a result, some folders appeared under "Artists", but with no files inside of them.

So this change would probably need a specific browsing menu for "Books" collection.

fireXtract:
Is there interest in this feature still? I can pick it up if nobody else is looking at it.

teacupx:
> Is there interest in this feature still? I can pick it up if nobody else is looking at it.

[...]

PR #2017, opened 2026-06-07
Improve audiobook support to Android Auto

jordanmichaelrushing:
**Changes**
  Extends the existing Android Auto audiobook browsing (added in upstream PR #1984) with several missing pieces:

  - Include audiobooks in Favorites and Recently Played sections
  - Include audiobooks in Genre browsing with direct playback action
  - Add Audiobooks group to search results
  - Resume audiobook playback from saved position (playbackPositionTicks) rather than always starting from the beginning

**Code assistance**
Opus used to help me understand where the underlying code needed to be ammended

**Issues**
- https://github.com/jellyfin/jellyfin-android/issues/247
- https://github.com/jellyfin/jellyfin-android/issues/314

**Testing**
I have tested this with my Android Auto device and confirmed all playing, pausing, selecting, etc of audiobook books works as intended. It also resumes the audiobook from where the server tells it to.

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/sessionbrowser/LibraryMediaId.kt
+++ b/app/src/main/java/org/jellyfin/mobile/sessionbrowser/LibraryMediaId.kt
@@ -15,7 +15,6 @@ sealed interface LibraryMediaId {
     data class Item(
         val itemId: UUID,
         val route: LibraryRoute,
-        val startPositionMs: Long = 0L,
     ) : LibraryMediaId
 
     @Serializable
--- a/app/src/main/java/org/jellyfin/mobile/sessionbrowser/LibraryMediaId.kt
+++ b/app/src/main/java/org/jellyfin/mobile/sessionbrowser/LibraryMediaId.kt
@@ -12,7 +12,11 @@ import java.util.UUID
 sealed interface LibraryMediaId {
     @Serializable
     @SerialName("item")
-    data class Item(val itemId: UUID, val route: LibraryRoute) : LibraryMediaId
+    data class Item(
+        val itemId: UUID,
+        val route: LibraryRoute,
+        val startPositionMs: Long = 0L,
+    ) : LibraryMediaId
 
     @Serializable
     @SerialName("route")
--- a/app/src/main/java/org/jellyfin/mobile/sessionbrowser/SessionBrowserCallback.kt
+++ b/app/src/main/java/org/jellyfin/mobile/sessionbrowser/SessionBrowserCallback.kt
@@ -42,6 +42,7 @@ import org.jellyfin.mobile.sessionbrowser.page.UserViewLibraryPage
 import org.jellyfin.sdk.api.client.ApiClient
 import org.jellyfin.sdk.api.client.extensions.universalAudioApi
 import org.jellyfin.sdk.api.client.extensions.userLibraryApi
+import org.jellyfin.sdk.model.api.BaseItemKind
 import org.jellyfin.sdk.model.api.MediaStreamProtocol
 import timber.log.Timber
 
@@ -373,7 +374,21 @@ class SessionBrowserCallback(
             }
         }
 
+        var resumePositionMs = startPositionMs
+        val currentLibraryMediaId = expandedItems.getOrNull(newStartIndex)?.mediaId?.let {
+            runCatching { Json.decodeFromString<LibraryMediaId>(it) }.getOrNull()
+        }
+        if (currentLibraryMediaId is LibraryMediaId.Item) {
+            runCatching {
+                val item by api.userLibraryApi.getItem(itemId = currentLibraryMediaId.itemId)
+                if (item.type == BaseItemKind.AUDIO_BOOK) {
+                    val positionTicks = item.userData?.playbackPositionTicks ?: 0L
+                    if (positionTicks > 0L) resumePositionMs = positionTicks / 10000L
+                }
+            }
+        }
+
         expandedItems = onAddMediaItems(mediaSession, controller, expandedItems).await()
-        MediaSession.MediaItemsWithStartPosition(expandedItems, newStartIndex, startPositionMs)
+        MediaSession.MediaItemsWithStartPosition(expandedItems, newStartIndex, resumePositionMs)
     }
 }
--- a/app/src/main/java/org/jellyfin/mobile/sessionbrowser/SessionBrowserCallback.kt
+++ b/app/src/main/java/org/jellyfin/mobile/sessionbrowser/SessionBrowserCallback.kt
@@ -53,6 +53,7 @@ class SessionBrowserCallback(
 ) : MediaLibrarySession.Callback {
     companion object {
         const val MAX_PAGE_SIZE = 250
+        private const val TICKS_PER_MILLISECOND = 10_000L
     }
 
     val pages = listOf(
@@ -383,7 +384,7 @@ class SessionBrowserCallback(
                 val item by api.userLibraryApi.getItem(itemId = currentLibraryMediaId.itemId)
                 if (item.type == BaseItemKind.AUDIO_BOOK) {
                     val positionTicks = item.userData?.playbackPositionTicks ?: 0L
-                    if (positionTicks > 0L) resumePositionMs = positionTicks / 10000L
+                    if (positionTicks > 0L) resumePositionMs = positionTicks / TICKS_PER_MILLISECOND
                 }
             }
         }
--- a/app/src/main/java/org/jellyfin/mobile/sessionbrowser/SessionBrowserCallback.kt
+++ b/app/src/main/java/org/jellyfin/mobile/sessionbrowser/SessionBrowserCallback.kt
@@ -5,6 +5,7 @@ import android.content.Context
 import android.net.Uri
 import android.os.Bundle
 import androidx.core.os.bundleOf
+import androidx.media3.common.C
 import androidx.media3.common.MediaItem
 import androidx.media3.common.MediaMetadata
 import androidx.media3.session.LibraryResult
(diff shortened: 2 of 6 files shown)


jordanmichaelrushing:
A few images showing it working (map obscured, sorry for the big orange boxes)

nielsvanvelzen:
Thanks for working on this, I wasn't able to test the audiobook support myself as I don't have any on my test servers. So I was mainly going with feedback I got from others. I'll take some time to review this later.

jordanmichaelrushing:
Thanks, I just added a second final commit to migrate the 10000 tick conversion to a const. Nothing more on my end to commit unless you have any change requests. We appreciate your hard work!

app/src/main/java/org/jellyfin/mobile/sessionbrowser/SessionBrowserCallback.kt
```diff
@@ -373,7 +375,21 @@ class SessionBrowserCallback(
...
 
+        var resumePositionMs = startPositionMs
+        val currentLibraryMediaId = expandedItems.getOrNull(newStartIndex)?.mediaId?.let {
+            runCatching { Json.decodeFromString<LibraryMediaId>(it) }.getOrNull()
+        }
+        if (currentLibraryMediaId is LibraryMediaId.Item) {
+            runCatching {
+                val item by api.userLibraryApi.getItem(itemId = currentLibraryMediaId.itemId)
+                if (item.type == BaseItemKind.AUDIO_BOOK) {
+                    val positionTicks = item.userData?.playbackPositionTicks ?: 0L
```
nielsvanvelzen: We have built-in functionality in our SDK to convert ticks to milliseconds


```suggestion
                    val positionTicks = item.userData?.playbackPositionTicks?.ticks?.inWholeMilliseconds ?: 0L
```

nielsvanvelzen:
Noticed I forgot to use translations for the search results so made #2018, you'll have to also add one for the newly added category in this PR (and probably rebase or merge master if the other PR is merged first)

app/src/main/java/org/jellyfin/mobile/sessionbrowser/page/GenreLibraryPage.kt
```diff
@@ -27,7 +27,10 @@ val GenreLibraryPage = { api: ApiClient ->
             LibraryPageElement.baseItem(
                 api = api,
                 item = it,
-                action = LibraryItemAction.Navigate(LibraryRoute.Album(it.id)),
+                action = when (it.type) {
+                    BaseItemKind.AUDIO_BOOK -> LibraryItemAction.Play(it)
+                    else -> LibraryItemAction.Navigate(LibraryRoute.Album(it.id))
```
nielsvanvelzen: I'd prefer doing it like this


```suggestion
                    BaseItemKind.AUDIO_BOOK -> LibraryItemAction.Play(it)
                    BaseItemKind.ALBUM -> LibraryItemAction.Navigate(LibraryRoute.Album(it.id))
                    else -> error("Unknown item type ${it.type}")
```

because with the query we should never receive other item types, and if we do we want the app so break so the issue is found quickly.

app/src/main/java/org/jellyfin/mobile/sessionbrowser/page/SearchLibraryPage.kt
```diff
@@ -52,6 +55,7 @@ val SearchLibraryPage = { api: ApiClient ->
             LibraryPageElement.Group("Playlists", playlists),
             LibraryPageElement.Group("Albums", albums),
             LibraryPageElement.Group("Artists", artists),
+            LibraryPageElement.Group("Audiobooks", audioBooks),
```
nielsvanvelzen: A string resource should be used here (your branch needs to be updated to current master so you have access to context). You can reuse the existing `media_service_car_section_audiobooks` string.

app/src/main/java/org/jellyfin/mobile/sessionbrowser/SessionBrowserCallback.kt
```diff
@@ -373,7 +375,21 @@ class SessionBrowserCallback(
...
         }
 
+        var resumePositionMs = startPositionMs
+        val currentLibraryMediaId = expandedItems.getOrNull(newStartIndex)?.mediaId?.let {
+            runCatching { Json.decodeFromString<LibraryMediaId>(it) }.getOrNull()
+        }
+        if (currentLibraryMediaId is LibraryMediaId.Item) {
+            runCatching {
+                val item by api.userLibraryApi.getItem(itemId = currentLibraryMediaId.itemId)
+                if (item.type == BaseItemKind.AUDIO_BOOK) {
```
nielsvanvelzen: I think we can safely use the playbackPosition from userData for all item types if `startPositionMs` is 0 (default value).
It shouldn't be returned for normal audio tracks but will make this code future proof if we ever add audio podcast support or other types.

app/src/main/java/org/jellyfin/mobile/sessionbrowser/SessionBrowserCallback.kt
```diff
@@ -373,7 +375,21 @@ class SessionBrowserCallback(
             }
         }
 
+        var resumePositionMs = startPositionMs
+        val currentLibraryMediaId = expandedItems.getOrNull(newStartIndex)?.mediaId?.let {
+            runCatching { Json.decodeFromString<LibraryMediaId>(it) }.getOrNull()
+        }
+        if (currentLibraryMediaId is LibraryMediaId.Item) {
+            runCatching {
+                val item by api.userLibraryApi.getItem(itemId = currentLibraryMediaId.itemId)
```
nielsvanvelzen: Running this API call for each and every item is extremely inefficient. We should, instead, query the userdata when fetching screens.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/mobile/sessionbrowser/SessionBrowserCallback.kt
jordanmichaelrushing: Addressed

app/src/main/java/org/jellyfin/mobile/sessionbrowser/page/GenreLibraryPage.kt
jordanmichaelrushing: Addressed

app/src/main/java/org/jellyfin/mobile/sessionbrowser/SessionBrowserCallback.kt
jordanmichaelrushing: Good catch. Addressed

jordanmichaelrushing:
Feedback addressed and rebased master in to get the conflict change.

app/src/main/java/org/jellyfin/mobile/sessionbrowser/LibraryMediaId.kt
```diff
@@ -12,7 +12,11 @@ import java.util.UUID
 sealed interface LibraryMediaId {
     @Serializable
     @SerialName("item")
-    data class Item(val itemId: UUID, val route: LibraryRoute) : LibraryMediaId
+    data class Item(
+        val itemId: UUID,
+        val route: LibraryRoute,
+        val startPositionMs: Long = 0L,
```
nielsvanvelzen: I'm not a fan of embedding the start position as part of the id for items, we already fetch the actual baseitems in the places that we want to use it so there is no need to make it part of the id.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/mobile/sessionbrowser/LibraryMediaId.kt
jordanmichaelrushing: Good catch, addressed.

app/src/main/java/org/jellyfin/mobile/sessionbrowser/LibraryMediaId.kt
```diff
@@ -12,7 +12,10 @@ import java.util.UUID
 sealed interface LibraryMediaId {
     @Serializable
     @SerialName("item")
-    data class Item(val itemId: UUID, val route: LibraryRoute) : LibraryMediaId
+    data class Item(
+        val itemId: UUID,
+        val route: LibraryRoute,
+    ) : LibraryMediaId
```
nielsvanvelzen, line 18: Please undo formatting changes

app/src/main/java/org/jellyfin/mobile/sessionbrowser/SessionBrowserCallback.kt
```diff
@@ -373,7 +375,17 @@ class SessionBrowserCallback(
             }
         }
 
+        var resumePositionMs = startPositionMs
+        val currentLibraryMediaId = expandedItems.getOrNull(newStartIndex)?.mediaId?.let {
+            runCatching { Json.decodeFromString<LibraryMediaId>(it) }.getOrNull()
+        }
+        if (currentLibraryMediaId is LibraryMediaId.Item && (startPositionMs == 0L || startPositionMs == C.TIME_UNSET)) {
+            val item by api.userLibraryApi.getItem(itemId = currentLibraryMediaId.itemId)
```
nielsvanvelzen, line 383: This still makes additional api requests that we do not need. When we convert the BaseItemDto into a MediaItem (within the toMediaItem) we can add the playback position to the metadata as an extra, something like:

```
action.item.userData?.playbackPositionTicks?.ticks?.inWholeMilliseconds?.let { extras.putLong("playbackPosition", it) }
```

(obviously need to move the extra key to a constant)

then we can read that here instead of making the API call.

nielsvanvelzen (changes requested):

nielsvanvelzen (approved):
~~I've pushed a commit to undo the formatting changes~~ (edit: you didn't allow pushes from maintainers so that didn't go through) and will make a separate PR to optimize the API calls for the resuming. Thanks for working on this!

PR #2021, opened 2026-06-10
Sync playback between Android Phone and Android Auto

jordanmichaelrushing:
Sync position and state for music & audiobooks for Android on your phone along with Android Auto. Currently these 2 use independent players. This change syncs the two. So if you are listening to a song on your phone at 2:00 and turn on Android Auto, and press play it will continue playing at 2:00. Both devices will also sync whether the audio is paused or playing.

**Code assistance**
I worked with Opus to identify the underlying issue and point me to where the underlying code is bugged and what would need to be created.

**Issues**
Doesn't depend on my [audiobook support PR](https://github.com/jellyfin/jellyfin-android/pull/2017), however it will work to support audiobooks whenever that is merged in.

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/app/AppModule.kt
+++ b/app/src/main/java/org/jellyfin/mobile/app/AppModule.kt
@@ -40,6 +40,7 @@ import org.jellyfin.mobile.player.mediasegments.MediaSegmentRepository
 import org.jellyfin.mobile.player.qualityoptions.QualityOptionsProvider
 import org.jellyfin.mobile.player.source.MediaSourceResolver
 import org.jellyfin.mobile.player.ui.PlayerFragment
+import org.jellyfin.mobile.sessionbrowser.SharedPlaybackStateRepository
 import org.jellyfin.mobile.setup.ConnectionHelper
 import org.jellyfin.mobile.utils.Constants
 import org.jellyfin.mobile.utils.PermissionRequestHelper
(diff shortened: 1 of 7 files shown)


nielsvanvelzen:
We're intentionally using separate player implementations so it will be easier to migrate them to our new player code.

jordanmichaelrushing:
Thanks Niels, that makes sense. Keeping the implementations separate for the migration is exactly what I want to support here, so I think we're aligned on the goal.

Just to make sure we're on the same page about this branch: it does keep the two players fully separate. Android Auto still has its own ExoPlayer, and the phone/webapp player is untouched. The only thing I added is a thin SharedPlaybackStateRepository that passes state one direction (phone -> Auto, for display) and commands the other (Auto -> phone: play/pause/seek/skip). The Auto MediaSession just swaps which player backs it depending on what's actively playing, so there's no merged or shared player, and either one can still be migrated independently.

If the coupling through that repository is the part you'd rather avoid, I'm happy to change how the two sides communicate; just let me know what you'd prefer.

One thing that'd help me: could you point me at the new player code you mentioned? Is that #1349, or a different branch? I'd like to build this so it ports cleanly.

nielsvanvelzen:
Within the new playback code the app itself is no longer allowed to interact with media3 directly, instead everything should go to the playback code its facades. The code for it is in the Android TV repository and will be introduced in the mobile app later (either once they are merged or via git submodule).

jordanmichaelrushing:
Got it. Would you rather I:
- Just hold this PR in this state
- Move it to a draft
- Close this out and what for whenever that work is complete?

nielsvanvelzen:
I'd say let's close this for now and we can revisit later.

PR #2034, opened 2026-06-20
Play downloaded audio in external app

nielsvanvelzen:
**Changes**

Although our video player can play audio-only items, the experience isn't great. It will still auto-hide controls and for items without any embedded cover art that would result in a black screen.

Use external apps for audio playback for now, we can change this back when we have better audio playback capabilities.

**Code assistance**

**Issues**


BotBlake:
Ur my hero :)

thornbill (approved):

PR #2039, opened 2026-06-21
(re-)Set ALLOW_CAPTURE_BY_ALL after Chromium loads

mkst:
**Changes**

This fixes WebView playback capture for apps that use Android audio playback capture.
Jellyfin already opts into playback capture, but WebView.loadUrl(...) appears to reset the app-wide capture policy back to `ALLOW_CAPTURE_BY_SYSTEM`. Reapplying `ALLOW_CAPTURE_BY_ALL` on the next UI loop keeps WebView-created audio tracks capturable, matching the existing local playback capture policy.

**Code assistance**
I used codex to help me get to the bottom of the problem and provide the minimal fix. 

**Issues**
N/A but happy to create an issue if required.

**Notes**
I have successfully tested the change on my phone (Samsung S20FE).

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/utils/AudioCapturePolicy.kt
+++ b/app/src/main/java/org/jellyfin/mobile/utils/AudioCapturePolicy.kt
@@ -1,0 +1,13 @@
+package org.jellyfin.mobile.utils
+
+import android.content.Context
+import android.media.AudioAttributes
+import android.media.AudioManager
+
+fun Context.allowPlaybackCaptureByAll() {
+    if (AndroidVersion.isAtLeastQ) {
+        getSystemService(AudioManager::class.java)
+            .allowedCapturePolicy = AudioAttributes.ALLOW_CAPTURE_BY_ALL
+    }
+}
+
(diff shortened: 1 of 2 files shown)


nielsvanvelzen (changes requested):
Can you link to the Android documentation that specifies this webview behavior? I can't find anything saying the webview chanegs the capture policy.

mkst:
Logging the audio capture policy before & after loadUrl shows the state change, I don't have any documentation to explain why...

mkst:
Presumably happening as [part of the initialisation](
https://chromium.googlesource.com/chromium/src/media/%2B/master/base/android/java/src/org/chromium/media/AudioManagerAndroid.java#191)
```java
        // Set the audio capture policy based on the device class.
        // For non-desktop devices, restrict capture to the system apps, unless the
        // feature is enabled to allow capture for all apps on all form factors.
        if (!DeviceInfo.isDesktop()
                && !AudioManagerAndroidJni.get().isAudioPlaybackCaptureAllowedFeatureEnabled()) {
            mAudioManager.setAllowedCapturePolicy(AudioAttributes.ALLOW_CAPTURE_BY_SYSTEM);
        }
```

I can keep digging?

nielsvanvelzen:
It looks like this behavior was [recently changed](https://github.com/chromium/chromium/commit/feb322a9ad8ad24f787e1f350dc5eeb5d16177cd), starting with chromium/webview version 150. So I don't think we need a workaround for it as it should be solved when the newer webview version is released.

mkst:
Aha, so the code I linked is the FIXED version (I was confused how the inner statement was reached), but makes sense if they were only checking non-desktop.

Shall I close this?

nielsvanvelzen:
Yeah I think we can close this for now, if for some reason the issue still persist when webview 150 (or newer) is released lmk and we can revisit this change.

PR #2041, opened 2026-06-21
Fixes desktop-mode cursor disappearing

### LINKED ISSUE #2040: Android Desktop mode issues. (Samsung DeX , AR glasses , etc)

### Describe the bug

Running Jellyfin 2.6.3 from google play:

The app's  default "video player type" is "web player" 
Web-player hides mouse cursor(or touchpad cursor) over the window, which basically sabotages the users control over the player.
The "integrated player" is a better default for desktop mode - this may be hard to troubleshoot for a new user.

Selecting "integrated player" works, the "brightness and volume gestures" seems to be mapped to phone's display, but it is doing something with the video brightness, but only adjusting it up while a gesture is performed.
When in desktop mode, but windowed, the player window gets extra bright as the cursor moves over it. 
Maybe just avoid messing with brightness when in desktop mode?

### Media info of the file

```shell
n/a this is regardless of media played.
```

- [x] Web player (default)

### Client logs

```shell
-
```

### FFmpeg logs

```shell
-
```

### Application version

2.6.3

Google Play Store

### Device information

Samsung S25 Ultra

### Android version

Android 16

### Jellyfin server version

10.11.11

### Other sources

_No response_

AndKe:
**Changes**
Fix cursor disappearing in Samsung DeX / desktop mode

When using the Jellyfin web player in desktop mode (e.g. Samsung DeX with the phone screen as a touchpad), the mouse cursor permanently disappears once the OSD times out and never comes back.

Root cause: The web client adds a mouseIdle class to <body> after pointer inactivity, which triggers a CSS rule hiding the cursor. It removes that class in response to mousemove events — but the mobile layout (which this app always reports via AppHost.getDefaultLayout) doesn't register that handler, so the class is never removed and the cursor stays hidden.

Checked and works perfectly fine, not affecting non-desktop operation.

**Code assistance**
Claude

**Issues**
Fixes #2040

### DIFF
--- a/app/src/main/assets/native/nativeshell.js
+++ b/app/src/main/assets/native/nativeshell.js
@@ -31,6 +31,26 @@ for (const plugin of plugins) {
     };
 }
 
+// In desktop mode (e.g. Samsung DeX) the web client adds 'mouseIdle' to <body>
+// after pointer inactivity but never removes it in the mobile layout, leaving
+// the cursor permanently hidden. Drive the hide/show cycle from our side.
+(function () {
+    const IDLE_MS = 3000;
+    let idleTimer = null;
+    document.addEventListener('pointermove', () => {
+        document.body.classList.remove('mouseIdle');
+        clearTimeout(idleTimer);
+        idleTimer = setTimeout(() => document.body.classList.add('mouseIdle'), IDLE_MS);
+    }, { capture: true, passive: true });
+}());
+
+// Prevent pointer lock from hiding the cursor in desktop mode.
+Object.defineProperty(Element.prototype, 'requestPointerLock', {
+    value: function() { return Promise.resolve(); },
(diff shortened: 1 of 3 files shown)


Maxr1998:
Shouldn't this ideally be fixed directly in jellyfin web, instead of adding a workaround to the mobile app?

nielsvanvelzen:

AndKe:
I saved a lot of time locating the problematic code this way, I never looked at this project before yesterday, so it would take me some time to even know where to look. None of what I submitted or wrote is 100% LLM generated. I even forgot til make a branch for this and made it on master.  :) 
I do not care who fixes it, as long as it's fixed.
Please implement the same, functionally same or similar change on client or server side written w/o any LLM or autocomplete.

PR #2042, opened 2026-06-22
Integrated player: inhibit brightness control for external displays

AndKe:
This prevents unwanted brightness setting when using external displays (Android Desktop mode)  - which produces unwanted effects.

**Changes**
Had to use texture_view surface type to get hover listener to work.

**Code assistance**
no

**Issues**
This resolves the second issue described in https://github.com/jellyfin/jellyfin-android/issues/2040  - the first was about web player permanently disabling mouse.  The second about Integrated player.

### DIFF
--- a/app/src/main/res/layout/fragment_player.xml
+++ b/app/src/main/res/layout/fragment_player.xml
@@ -12,7 +12,8 @@
         android:layout_height="match_parent"
         android:foreground="@drawable/ripple_background"
         app:pause_icon="@drawable/ic_pause_black_42dp"
-        app:play_icon="@drawable/ic_play_black_42dp" />
+        app:play_icon="@drawable/ic_play_black_42dp"
+        app:surface_type="texture_view" />
 
     <FrameLayout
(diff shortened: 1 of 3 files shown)


app/src/main/res/layout/fragment_player.xml
```diff
@@ -12,7 +12,8 @@
         android:layout_height="match_parent"
         android:foreground="@drawable/ripple_background"
         app:pause_icon="@drawable/ic_pause_black_42dp"
-        app:play_icon="@drawable/ic_play_black_42dp" />
+        app:play_icon="@drawable/ic_play_black_42dp"
+        app:surface_type="texture_view" />
```
nielsvanvelzen, line 16: Changing from a surface view to a texture view is really inefficient and might actually cause video to not show in some cases. So this has to be reverted.

nielsvanvelzen (changes requested):

nielsvanvelzen:
Closing as stale

PR #2050, opened 2026-06-27
Use native video player by default

nielsvanvelzen:
**Changes**

A few years back we decided to keep the web player as default because the native one was still lacking. To my understanding we've fixes most of the issues and I actually thought we already switched the default for the 2.7 release (spoiler: we did not!).

This PR changes the default video player to be the internal one. It's already used for downloads too, so this creates a more consistent experience.

**Code assistance**

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/app/AppPreferences.kt
+++ b/app/src/main/java/org/jellyfin/mobile/app/AppPreferences.kt
@@ -94,7 +94,7 @@ class AppPreferences(context: Context) {
 
     @VideoPlayerType
     val videoPlayerType: String
-        get() = sharedPreferences.getString(Constants.PREF_VIDEO_PLAYER_TYPE, VideoPlayerType.WEB_PLAYER)!!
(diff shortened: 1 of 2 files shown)


Maxr1998:
Iirc SyncPlay is still missing, which was one major reason for keeping the web player the default, also subtitle support.

nielsvanvelzen:
With SyncPlay enabled the web player will always be used, that was part of #1076. As for subtitles, can you be more specific? To my understanding they work fine.

Maxr1998:
Oh, I forgot about that PR. With subtitles, I primarily meant ASS subtitles, for which ExoPlayer support is still lacking. Not even sure how well the web player supports those though. Either way, it has actually been 5 years since we reverted, so maybe it is worth giving it another try.

Maxr1998 (approved):

nielsvanvelzen:
ASS support is a solved problem too with the [libass-android](https://github.com/peerless2012/libass-android) library that was incubated in our development chats! We're gonna test it within the next Android TV release and add it to mobile later if everything is working.

PR #2062, opened 2026-07-02
Add preferred audio transcoding codec setting

einanderson:
**Changes**

Adds a 'Preferred audio codec' setting (Automatic/AAC/AC3/MP3) for the integrated player. The chosen codec is moved to the front of the HLS/ts and mkv transcoding profiles, so the server transcodes audio to it. 'Automatic' keeps the original codec order (unchanged behaviour). The device profile is built fresh per playback, so the setting applies without an app restart.

**Motivation**

When the audio track needs transcoding, the server always picks MP3, because it is the first supported codec in the hardcoded profile list (`mp1,mp2,mp3,aac,...`). AAC would offer better quality/compatibility on mobile, but is never chosen. This makes the codec user-configurable instead of reordering the list for everyone.

**Testing**

Feature-tested end-to-end on a v2.6.4-based build against Jellyfin 10.11 (sessions report `aac` after selecting AAC). This port to master compiles and passes detekt.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


nielsvanvelzen:

PR #2063, opened 2026-07-02
Add maximum transcode resolution setting

einanderson:
**Changes**

Adds a 'Maximum resolution' setting (Automatic/1080p/720p/480p) for the integrated player. When set, a `Height <= x` `ProfileCondition` is attached to the video transcoding profiles of the device profile. The server honors it via `ApplyTranscodingConditions` when transcoding; direct play eligibility is unaffected (the condition is deliberately not added to the codec profiles).

**Motivation**

The in-player quality menu only transmits a bitrate — the resolution in the option label is cosmetic, and the server derives the actual output height from the bitrate alone. Selecting '720p - 4 Mbps' can actually produce a 540p stream, and combinations like 720p at 8 Mbps are impossible. With this setting the user controls the transcode resolution independently of the bitrate selection.

**Testing**

Verified against Jellyfin 10.11: a PlaybackInfo request with the condition returns `MaxHeight` alongside the requested bitrate, and sessions transcode at the capped resolution (e.g. 1280x720 at 4 Mbps where the server previously chose 540p). Feature-tested end-to-end on a v2.6.4-based build; this port to master compiles and passes detekt.

Note: trivially overlaps with #2062 in `Constants`/`SettingsFragment`/`strings.xml` (adjacent lines); happy to rebase whichever lands second.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/app/AppPreferences.kt
+++ b/app/src/main/java/org/jellyfin/mobile/app/AppPreferences.kt
@@ -8,6 +8,7 @@ import org.jellyfin.mobile.downloads.DownloadMethod
 import org.jellyfin.mobile.player.mediasegments.MediaSegmentAction
 import org.jellyfin.mobile.player.mediasegments.toMediaSegmentActionsString
 import org.jellyfin.mobile.settings.ExternalPlayerPackage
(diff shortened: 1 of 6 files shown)


nielsvanvelzen:

PR #2065, opened 2026-07-05
Show Google Cast streaming quality setting on Chromecast-capable builds

JDStriker456:
**Why**
I have an old chromecast and the option to set the bitrate is available in the web app but not on the andoid one. The cast feature was added in 2020 but the settings where not updated to show it is compatible with casting therefore settings the bitrate was not available.

**Changes**
I added `BuildConfig.IS_PROPRIETARY` as a way to identify casting capability and showing the setting.

**Code assistance**
I Used an AI assistant (Claude) for investigating the git log to try to find out why this feature was left behind. I then ask it to code the fix and tested it on my phone. 

This pr description is not written by an llm.
Thanks for all you do !

### DIFF
--- a/app/src/main/assets/native/nativeshell.js
+++ b/app/src/main/assets/native/nativeshell.js
@@ -16,6 +16,12 @@ const features = [
     "subtitleburnsettings"
 ];
 
+// Chromecast only works in the proprietary flavor; the libre build ships a no-op
+// cast stub, so only advertise support when the native bridge confirms it.
+if (window.NativeInterface.hasChromecast()) {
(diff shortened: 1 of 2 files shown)


nielsvanvelzen (changes requested):
Good find, we should indeed claim the "chromecast" feature for this specific setting to show up.

I think it would work best if we keep the "features" array for all static features (always available) and modify the supports function instead to check for the chromecast support on the native side if requested. That way we can avoid (another) blocking call during script initialization.

nielsvanvelzen (approved):

PR #2085, opened 2026-07-18
Remove automatic landscape mode

### LINKED ISSUE #2069: Portrait Mode In Video Player Is Forcing Notification And Gesture Bars On Large Devices

### Describe the bug

When using the integrated player (ExoPlayer) on the inner display of my foldable (Pixel 10 Pro Fold), the status bar and gesture navigation bar never hide while the device is held in portrait, and pressing the fullscreen button in the player controls does nothing at all. Physically rotating the device to landscape hides the bars correctly, but I use the inner screen in portrait since that is the natural way to hold the phone unfolded. Everything works fine on the narrow outer screen and on regular phones. I reproduced this on the latest Google Play beta on my physical device, as well as with a debug build of the latest master branch running in the emulator. It appears to be caused by Android 16 ignoring setRequestedOrientation() on large screens (smallest width ≥ 600dp) for apps targeting SDK 36, the fullscreen button for landscape videos only requests a rotation and relies on the resulting configuration change to hide the system bars, so on the inner display it becomes annoying to use.

1. Start playback of any (landscape) video on the inner display, holding the device in portrait.
2. The status bar and gesture navigation bar remain visible and never hide.
3. Tap the fullscreen button in the player controls -> nothing happens at all. The system bars stay visible.
4. Physically rotate the device to landscape -> the system bars hide correctly.
5. Rotate back to portrait -> the bars come back, with no way to hide them again short of rotating.

Image

### Logs

```shell

```

### Application version

2.7.0-beta.3

Google Play Store

### Device information

Google Pixel 10 Pro Fold

### Android version

Android 17

### Jellyfin server version

10.11.11

### Other sources

_No response_

nielsvanvelzen:
**Changes**

Starting with Android 17 [Google now expects](https://android-developers.googleblog.com/2025/01/orientation-and-resizability-changes-in-android-16.html) users to rotate their devices themselves, and not because an app forces them too. As such they implemented behavior changes that cause our automatic landscape mode option to fail.

We could make workarounds for this but it will only increase the chances of Google rejecting our app again. Instead, remove the option and require the user to manually rotate their device.

**Code assistance**

**Issues**

Fixes #2069


I-G-1-1:
I loved that the app until now autorotate the video to landscape.
In this way I could keep autorotate disabled on Android (as is very annoying for me having all the app and the home screen rotate as soon I incline my phone and have to keep the phone perfectly straight) and still be able to see the movies in landscape without have to re-enable every time autorotate.
No one will consume 16:9 media (I think one of the main mission of Jellyfin) on portrait so autorotate rotation to landscape when playing a video it's perfectly natural.
Also consider all the elderly people that until now had their tablet/phone autorotate to landscape when they start a movie and now they will have to interact with the autorotate toggle in the drop-down menù (and I'm pretty sure a lot of them don't even know what is the drop-down menu)

So I'm not very happy with this change. While I understand the challenge with Google approval, I really wish you could reconsider and find a way to reimplement this function.

nielsvanvelzen:
It's not like it was my preferred solution to remove this functionality. But we target the latest Android releases and Google decided this kind of functionality is no longer allowed. If you have an alternative solution please do share, but I can't think of any.

I-G-1-1:
> Google decided this kind of functionality is no longer allowed

are you sure this is true also if the functionality is only opt-in (default behaviour is Google compliant, but If the user want can enable the "automatic landscape mode")?

If the "automatic landscape mode" is really forbidden, I think at least there should be a button in the player interface to toggle between portrait and landscape ignoring if the main Android autorotation is toggled on or off.
Here is an example from SE Video player:
SE video player portrait
SE video player landscape

With System wide autorotation off, tapping on the icon (see the arrow) let me toggle between  portrait and landscape.

nielsvanvelzen:
Read the link in my PR description. As of Android 16 any orientation request from the app is ignored for wide devices (e.g. tablets and foldables) and they will expand this in the future for eventually all device form-factors.
There is no way to check whether the OS still supports app-requested rotation or not. If the OS offered a feature flag whether app-requested orientation is allowed we could gate the app setting behind it, but there is none so we literally cannot know if this feature can be enabled or not.

I-G-1-1:
Sorry I'm not a developer so probably I'm wrong, but in the link you posted I read:

> Also, users have control. They can explicitly opt-in to using the app’s default behavior in the [aspect ratio settings](https://developer.android.com/guide/practices/device-compatibility-mode#user_per-app_overrides).

Maybe I don't understand because the link refer to Android 16, and you were referring to a policy change for Android 17.

It seems to me more a policy for the whole app interface an not for specific app functionality like launch the play in landscape if the user opted-in to this behavior.

EDIT: in theory the rule is for tablet and inner screen of foldable, while phone (< 600dp) are not impacted by this new rule. 
So why completely remove the feature also for the device < 600dp?

app/src/main/java/org/jellyfin/mobile/settings/SettingsFragment.kt
effleurager, line 56: This member and its last remaining reference is causing a crash when changing the video player type

nielsvanvelzen:
> Sorry I'm not a developer so probably I'm wrong, but in the link you posted I read:
> 
> > Also, users have control. They can explicitly opt-in to using the app’s default behavior in the [aspect ratio settings](https://developer.android.com/guide/practices/device-compatibility-mode#user_per-app_overrides).
> 
> Maybe I don't understand because the link refer to Android 16, and you were referring to a policy change for Android 17.

That link is only for the aspect-ratio changes, those are unrelated. Also I did mean Android 16, that was a mistake on my part in the PR description.
The reason this new behavior now affects us is because starting with the 2.7.0 release we build the app with Android 16 as target.
 
> It seems to me more a policy for the whole app interface an not for specific app functionality like launch the play in landscape if the user opted-in to this behavior.
> 
> EDIT: in theory the rule is for tablet and inner screen of foldable, while phone (< 600dp) are not impacted by this new rule. So why completely remove the feature also for the device < 600dp?

Yes, in theory we could add a check "if device is Android 16 or newer and device size is above 600dp disable setting". But what if Google decides to make it 500 dp in the next release? or Samsung decides it needs to be 650dp on their Android version? These kinds of things happen A LOT. So normally to avoid those issues a feature flag would be introduced by the OS and we can read that to know whether a behavior change exists or not. But as far as I'm ware Google did not introduce one for this behavior change.

If we find a good way to detect whether orientation requests are still allowed we'll revert this change. But I have not found any.

nielsvanvelzen:
I'll be reverting this change for the next beta and enable the opt-out. This will only work for the 2.7 releases as they target Android 16. Starting with 2.8 we will be targeting Android 17 and the opt-out will no longer work.

I-G-1-1:
Thanks. I'm glad you reverted the changes for Android 16.

in the link you posted _Timeline_ paragraph tells that:
- For Android 16 the developer could opt-out: so you created [Opt-out of Android 16 restricted resizability behavior
#2096](https://github.com/jellyfin/jellyfin-android/pull/2096).
- For Android 17 the developer could **NOT** opt-out.

I was thinking about this part in that page that states:

> Also, users have control. They can explicitly opt-in to using the app’s default behavior in the [aspect ratio settings](https://developer.android.com/guide/practices/device-compatibility-mode#user_per-app_overrides).

Now I don't understand if this phrase means that:
1. there will be an option in the OS settings for each app to tell the OS to use the app’s default behavior
or
2. the developer could put an option inside the app that the user could enable (opt-in) to enable the autorotation to landscape on play.

I think it's the first case: an option in the OS settings for each app to tell the OS to use the **app’s default behavior**.
If this is the case removing the auto-rotation feature from the app will make impossible for the user to enable app’s default behavior.

As this phrase is in the Android 16 section I don't understand if this apply also for Android 17.

I think that the rules are:
1. for device ≥ 600dp the app should behave has Google want, but if the settings of the app on system settings tell to "use the app behavior" the OS won't ignore the app calls and use the app’s default behavior
2. for device < 600dp these rules doesn't apply and the OS doesn't ignore the app calls.

I don't see why you think "google will reject the app" these rules tells the developer that: 
- in Android 16 he could opt-out
- in Android 17 the OS will drop any call to rotation of the user interface if the device is ≥ 600dp. 
There isn't any warn that call to this outdated API is against rules and if you are not compliant the app won't be approved, just "the OS will ignore your call".
So why pro-actively remove one of the features that makes the use of the App a better option than the use of the browser to access Jellyfin.

> If we find a good way to detect whether orientation requests are still allowed we'll revert this change. But I have not found any.

Now that we can keep the feature on the app version that target Android 16 there will be more time to understand how prepare for Android 17

Thank you for the time and effort dedicated to developing this app.

Issue #1781, opened 2025-09-06
Background playback is stopping after 60-70 seconds


SeventhUnderhand:
### This issue respects the following points:

- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin-android/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Describe the bug

when i minimize the app to the background the playback will stop after about 60-70 seconds

- Android version = 16 BD3A.250721.001.B7
- Jellyfin App = latest according to play store
- Pixel 10 pro
- using both Nova and Pixel launcher

- Integrated player
- allow background playback is enabled
- app is set to unrestricted for battery
- Data saver if OFF
- Adaptive connectivity is OFF

### Media info of the file

```shell

```

### Which video player implementations does this bug apply to?

- [ ] Web player (default)
- [x] Integrated player (ExoPlayer)
- [ ] External player (VLC, mpv, MX Player)

### Client logs

```shell
adb logcat | grep org.jellyfin.mobile
09-06 18:49:46.639  1839  2852 I MR2ServiceImpl: Updating composite discovery preference | preference: RouteDiscoveryRequest{ preferredFeatures={com.google.android.gms.cast.CATEGORY_CAST/2FA4D21B///ALLOW_IPV6, com.google.android.gms.cast.CATEGORY_CAST/CC32E753///ALLOW_IPV6, com.google.android.gms.cast.CATEGORY_CAST/F007D354///ALLOW_IPV6, android.media.route.feature.LIVE_AUDIO}, activeScan=true }, active routers: [org.jellyfin.mobile, com.spotify.music, au.com.shiftyjelly.pocketcasts]
[...]

PR #2104, opened 2026-08-02
Fix integrated player stopping in the background

pardovot:
**Changes**
The integrated player lives in the activity, so nothing keeps the process in the foreground once the player UI is hidden, android classifies it as background process, mutes its audio and kills it shortly after, which breaks background audio
playback.

This adds a small `mediaPlayback` foreground service that owns the existing player notification, `PlayerNotificationHelper` now starts the service with the notification instead of posting it directly, and updates the notification in place while the service runs, falling back to a plain notification if the service isn't allowed to start.
The player additionally sets `WAKE_MODE_NETWORK` so CPU and network stay available while the screen is off.

**Code assistance**
Used claude code to debug and write the code, tested it on 2 android devices + emulator.

**Issues**
 - https://github.com/jellyfin/jellyfin-android/issues/1781


app/src/main/java/org/jellyfin/mobile/player/interaction/PlayerNotificationHelper.kt
```diff
@@ -127,7 +128,14 @@ class PlayerNotificationHelper(private val viewModel: PlayerViewModel) : KoinCom
...
 
-            nm.notify(VIDEO_PLAYER_NOTIFICATION_ID, notification)
+            // The notification is owned by a foreground service so the process keeps playing while
+            // the UI is hidden. Once started, update it directly instead of restarting the service.
+            when {
+                serviceStarted.get() -> nm.notify(VIDEO_PLAYER_NOTIFICATION_ID, notification)
+                PlayerService.start(context, notification) -> serviceStarted.set(true)
+                // Starting the service failed, fall back to a regular notification
+                else -> nm.notify(VIDEO_PLAYER_NOTIFICATION_ID, notification)
+            }
```
nielsvanvelzen: I believe you can always submit the notification here, and just start the service. The service can listen for start commands to see when a second start call happens when it already is started. There shouldn't be a real reason to track whether the service is started.

app/src/main/java/org/jellyfin/mobile/player/interaction/PlayerService.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    companion object {
+        /**
+         * The notification to start the service in the foreground with.
+         *
+         * Kept after the service stops so a start command that is still queued can always reach
+         * startForeground.
+         */
+        @Volatile
+        private var pendingNotification: Notification? = null
```
nielsvanvelzen: This is really dirty. It would be much better to send the notification(-id) over the start intent as extra.

nielsvanvelzen (changes requested):

renthy-dev:
I've compiled and been running this for the last few days and works perfectly. (Even if a dirty implementation). My device on Android 16 had constant micro stutters when playing in the background and when I moved to 17 it had the stutters at first and then after a few updates it wouldnt play in the background at all. This pull request has fixed the issues fully.

nielsvanvelzen:
Hi @pardovot are you still planning to address the review feedback and fix the merge conflicts?

pardovot:
@nielsvanvelzen Yes, I already made the changes but didn't had enough time to test it and verify that it works properly, will try to get to it soon. Thanks!

nielsvanvelzen (approved):
Made some final changes in the last commit, otherwise seemed fine. Thanks for the PR!

Issue #2105, opened 2026-08-02
Color matched navbar


BotBlake:
### This request respects the following points:

- [x] This request is **not** already on [GitHub](https://github.com/jellyfin/jellyfin-android/labels/enhancement) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).

### Problem description

On Android, Apps have the ability to control transparency and color of the navigation bar,
A lot of common apps (YouTube, Discord, Twitch, WhatsApp, Prime Video, ...) make use of this to "blend in" the navbar.

Jellyfin android does not.

### Proposed solution

Set the background color for the Navbar to match the background color of the app.

### Alternatives considered

Set the Navbar to transparent.

### Additional information

This issue was reported based on the linked Reddit comment below.

### Other sources

https://www.reddit.com/r/jellyfin/comments/1vdifrs/comment/p1b6rfo

klueman:
This is the same issue as reported here, but was closed for inactivity:

https://github.com/jellyfin/jellyfin-android/issues/1709

I shared screenshots of it a year ago today.

https://github.com/jellyfin/jellyfin-android/issues/1709#issuecomment-3148515693

After updating to 2.7, the notification bar at the top seems to have the correct coloring, but the navigation bar at the bottom is still white.

<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/d8839bb0-3d6c-475d-951f-51fac6cf01f1" />

It's as if Jellyfin does not impose its own coloring to the bar and it falls back to its default white.

Issue #1709, opened 2025-07-02
Status bar icons do not adapt to background color, always displayed in gray.


JonesBoton:
### This issue respects the following points:

- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin-android/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Describe the bug

**Bug Description:**
After launching Jellyfin, the status bar icons remain dark even though the status bar background is dark gray. This causes the icons (battery, signal, time) to be barely visible or difficult to distinguish against the dark background.

**Expected Behavior:**
Status bar icons should dynamically adjust based on the background color. Specifically:

Light-colored icons should be displayed on dark backgrounds

Dark-colored icons should be displayed on light backgrounds

![Image](https://github.com/user-attachments/assets/66e59b38-21d4-4de7-b9e4-38eb09fc14a5)

### Logs

```shell

```

### Application version

2.6.3

### Where did you install the app from?

F-Droid

### Device information

Samsung Galaxy S24

### Android version

Android 15

### Jellyfin server version

10.10.7

klueman:
Having this issue as well. The navigation bar also doesn't align with the phone's color settings. The previous Jellyfin on Android version did not have this issue.

For comparison, two screenshots, on of Jellyfin and one of Firefox. Notice the difference in the status and navigation bar colors. I changed the display color in Jellyfin to Blue Radiance so it's easier to see the difference

<img width="405" height="900" alt="Image" src="https://github.com/user-attachments/assets/b034d3aa-3aa2-46c2-ac31-aaaa30abd390" />

<img width="456" height="1013" alt="Image" src="https://github.com/user-attachments/assets/1660bab5-8784-4041-bbb0-a27771bc43e3" />

edual16:
Hi,

I think it is related with this: https://developer.android.com/develop/ui/views/layout/edge-to-edge

Saulimedes:
Damn this issue is still open, isn't it? I first thought I did cause it with custom branding but even when switching back to the default it is a grey top with black foreground color.

PR #2107, opened 2026-08-03
set navbar color to transparent

BotBlake:
**Changes**

Extend the system bar configuration of the app to include the navigation bar.
Currently, only the `statusBarStyle` is set. This PR applies the same styling to the `navigationBarStyle`.

**Code assistance**

None

**Issues**

https://github.com/jellyfin/jellyfin-android/issues/2105
https://github.com/jellyfin/jellyfin-android/issues/1709


BotBlake:
| Before | After |
| - | - |
|grafik |grafik|

nielsvanvelzen (approved):

PR #2136, opened 2026-08-13
Add ass/ssa subtitle support

rodis120:
**Changes**
Add support for DirectPlay with ASS/SSA subtitles using [libass-android](https://github.com/peerless2012/libass-android).

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/player/PlayerViewModel.kt
+++ b/app/src/main/java/org/jellyfin/mobile/player/PlayerViewModel.kt
@@ -27,6 +27,8 @@ import androidx.media3.exoplayer.mediacodec.MediaCodecSelector
 import androidx.media3.exoplayer.source.MediaSource
 import androidx.media3.exoplayer.trackselection.DefaultTrackSelector
 import androidx.media3.exoplayer.util.EventLogger
+import io.github.peerless2012.ass.media.AssHandler
+import io.github.peerless2012.ass.media.factory.AssRenderersFactory
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
@@ -239,6 +241,7 @@ class PlayerViewModel(application: Application) : AndroidViewModel(application),
                 .build()
             else -> DefaultLoadControl()
         }
+        val appPreferences: AppPreferences = get()
         val renderersFactory = DefaultRenderersFactory(getApplication()).apply {
             setEnableDecoderFallback(true) // Fallback only works if initialization fails, not decoding at playback time
             val rendererMode = when {
@@ -272,7 +275,7 @@ class PlayerViewModel(application: Application) : AndroidViewModel(application),
 
                 filteredDecoderList
             }
-        }
+        }.let { if (appPreferences.exoPlayerDirectPlayAss) AssRenderersFactory(get(), it) else it }
         _player.value = ExoPlayer.Builder(getApplication(), renderersFactory, get()).apply {
             setUsePlatformDiagnostics(false)
             setTrackSelector(trackSelector)
@@ -282,6 +285,7 @@ class PlayerViewModel(application: Application) : AndroidViewModel(application),
         }.build().apply {
             addListener(this@PlayerViewModel)
             applyDefaultAudioAttributes(C.AUDIO_CONTENT_TYPE_MOVIE)
+            if (appPreferences.exoPlayerDirectPlayAss) get<AssHandler>().init(this)
         }
     }
 
--- a/app/src/main/java/org/jellyfin/mobile/player/ui/PlayerFragment.kt
+++ b/app/src/main/java/org/jellyfin/mobile/player/ui/PlayerFragment.kt
@@ -27,6 +27,8 @@ import androidx.fragment.app.viewModels
 import androidx.lifecycle.lifecycleScope
 import androidx.media3.common.Player
 import androidx.media3.ui.PlayerView
+import io.github.peerless2012.ass.media.AssHandler
+import io.github.peerless2012.ass.media.widget.AssSubtitleView
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.launch
 import org.jellyfin.mobile.R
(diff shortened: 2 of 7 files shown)


rodis120:
Currently it requires `Allow SSA/ASS subtitles in direct play`. Should I enable it by default? Or remove that option entirely make is use asslib all the time?

Also, summary under that option will need to get updated or removed.

nielsvanvelzen:
Ah I was also working on this already, hence the recent bumping of the minSdk and the dep updates that were needed first. Guess this saves me some work haha. I'll review it more proper later but let me start by saying we shouldn't use the `withAssMkvSupport`/`withAssSupport` and instead configure the player instance ourselves, just like we do in the TV app.

rodis120:
Ok, I will fix it later today.

rodis120:
https://github.com/jellyfin/jellyfin-androidtv/blob/5e9496ccdcebc1d21d838ce9ac0fc1ad8f014d56/app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java#L242

I see that `withAssMkvSupport` is used in the tv client anyway.

I used it in this PR but If you need me to I can just copy the content of that method directly there.

```diff
                     else -> TsExtractor.DEFAULT_TIMESTAMP_SEARCH_BYTES
                 },
             )
-        }.withAssMkvSupport(assSubtitleParserFactory, assHandler)
+        }.let {
+            ExtractorsFactory {
+                val extractors  = it.createExtractors()
+                extractors.forEachIndexed { index, extractor ->
+                    if (extractor is MatroskaExtractor) {
+                        // Must keep the extractor order.
+                        extractors[index] = AssMatroskaExtractor(assSubtitleParserFactory, assHandler)
+                    }
+                }
+                extractors
+            }
+        }
         DefaultMediaSourceFactory(get<CacheDataSource.Factory>(), extractorsFactory)
             .setSubtitleParserFactory(assSubtitleParserFactory)
     }
```

I also used `withAssSupport` to create `AssRenderersFactory`, in the end it is equal to that:

```diff
                 filteredDecoderList
             }
-        }.withAssSupport(assHandler)
+        }.let { AssRenderersFactory(assHandler, it) }
         _player.value = ExoPlayer.Builder(getApplication(), renderersFactory, get()).apply {
             setUsePlatformDiagnostics(false)
             setTrackSelector(trackSelector)
```

But the player instance itself is created without `withAssSupport` method.

nielsvanvelzen:
Sorry my bad, should've been more specific. The library changes a bit over time so I still had an older implementation in my memory.

For the renderers factory I'd prefer not using the extension, as the alternative is to just use the constructor. For the extractors factory it is fine to use the extension function.

Do make sure you do not configure/use the libass library unless the `exoPlayerDirectPlayAss` option is enabled.

rodis120:
I added the changes you wanted. But I wander if we should enable ass directplay by default at this point.

nielsvanvelzen:
From my experience with the development builds of the TV app, the libass library is fairly stable at this point but we have not done any large scale testing yet. I do know there is an issue with zlib compressed subtitles in mkv files (which is more common in the type of media that uses ass) but that's something that needs to be fixed in media3 itself (and an issue and PR exist).
For now I'd like to keep it under the existing preference for the mobile app, and we'll likely remove the option (and enable by default) in the next major release.

app/src/main/java/org/jellyfin/mobile/player/ui/PlayerFragment.kt
```diff
@@ -209,6 +212,10 @@ class PlayerFragment : Fragment(), BackPressInterceptor {
         // Disable controller animations
         playerView.setControllerAnimationEnabled(false)
 
+        if(appPreferences.exoPlayerDirectPlayAss) playerView.subtitleView?.apply {
```
nielsvanvelzen:
```suggestion
        if (appPreferences.exoPlayerDirectPlayAss) playerView.subtitleView?.apply {
```

app/src/main/java/org/jellyfin/mobile/player/PlayerViewModel.kt
```diff
@@ -281,6 +285,7 @@ class PlayerViewModel(application: Application) : AndroidViewModel(application),
         }.build().apply {
             addListener(this@PlayerViewModel)
             applyDefaultAudioAttributes(C.AUDIO_CONTENT_TYPE_MOVIE)
+            if(appPreferences.exoPlayerDirectPlayAss) get<AssHandler>().init(this)
```
nielsvanvelzen:
```suggestion
            if (appPreferences.exoPlayerDirectPlayAss) get<AssHandler>().init(this)
```

app/src/main/java/org/jellyfin/mobile/player/PlayerViewModel.kt
```diff
@@ -272,7 +276,7 @@ class PlayerViewModel(application: Application) : AndroidViewModel(application),
 
                 filteredDecoderList
             }
-        }
+        }.let { if(appPreferences.exoPlayerDirectPlayAss) AssRenderersFactory(get(), it) else it }
```
nielsvanvelzen:
```suggestion
        }.let { if (appPreferences.exoPlayerDirectPlayAss) AssRenderersFactory(get(), it) else it }
```

nielsvanvelzen (commented):
Some formatting issues (the linter also found them after I already added my review comments 😅). Otherwise seems to work fine!

rodis120:
BTW do you want me to also change the description under the `Allow SSA/ASS subtitles in direct play.`? Currently it says "`Prevent transcoding and show subtitles with basic styling only. Advanced subtitle styling will not be available if enabled.`" and it could be confusing to the users. I could remove entire description so there would be no need to change the transactions.

nielsvanvelzen:
> BTW do you want me to also change the description under the `Allow SSA/ASS subtitles in direct play.`? Currently it says "`Prevent transcoding and show subtitles with basic styling only. Advanced subtitle styling will not be available if enabled.`" and it could be confusing to the users. I could remove entire description so there would be no need to change the transactions.

Oh good one, I thought about that and then forgot to mention it... I'd change the description to be something like "Use experimental client-side rendering, may not always work" or something like that.
To make sure translations are updated, rename the key of the description too. Make sure to **only update** the english file. Automation will deal with the translations after merge.

rodis120:
@nielsvanvelzen Done

nielsvanvelzen (approved):

nielsvanvelzen:
Thanks for the PR!

PR #2139, opened 2026-08-15
Remove custom device profile for web player

### LINKED ISSUE #1502: I can't change the audio track
### Describe the bug

I'm having trouble switching between audio tracks on some content. On the web, I can easily switch between tracks without any issues, but on the Android app (using the default Web player), the audio track doesn't change.

To clarify, the content has audio tracks in both pt-BR and en-US. However, even when I select the en-US track, the playback remains in pt-BR.

### Logs

_No response_

### Application version

2.6.2

None

### Device information

Xiaomi Redmi Note 12

### Android version

Android 14

### Jellyfin server version

10.9.11

- [X] Web player (default)

Axyss:
Whenever a piece of media with multiple audio tracks is directly played in the app using the web player, only the predetermined audio track works. Switching to others updates the selection in the UI, but does nothing.

From my understanding, the web player inside the app should mimic the behavior of the web player in a browser. But right now, media that would switch tracks normally in a browser by getting remuxed, keeps playing in `Stream type: Video` and `Play method: Direct play` modes which depend on the, not supported by WebView, [AudioTrack api](https://developer.mozilla.org/en-US/docs/Web/API/AudioTrack).

**Changes**
The proposed fix stops the client from stripping useful CodecProfiles that'd tell the server when remuxing would be appropiate and stop the client from using an unsupported api.

**Code assistance**
LLMs were used to navigate and find some key logic faster in multiple Jellyfin repos.

**Issues**
Would resolve #1502

### DIFF
--- a/app/src/main/assets/native/nativeshell.js
+++ b/app/src/main/assets/native/nativeshell.js
@@ -104,10 +104,6 @@ function getDeviceProfile(profileBuilder, item) {
         enableMkvProgressive: false
     });
 
-    profile.CodecProfiles = profile.CodecProfiles.filter(function (i) {
-        return i.Type === "Audio";
-    });
-
     profile.CodecProfiles.push({
         Type: "Video",
         Container: "avi",
--- a/app/src/main/assets/native/nativeshell.js
+++ b/app/src/main/assets/native/nativeshell.js
@@ -100,47 +100,7 @@ window.NativeShell = {
 };
 
 function getDeviceProfile(profileBuilder, item) {
-    const profile = profileBuilder({
-        enableMkvProgressive: false
-    });
-
-    profile.CodecProfiles.push({
-        Type: "Video",
-        Container: "avi",
-        Conditions: [
-            {
-                Condition: "NotEquals",
-                Property: "VideoCodecTag",
-                Value: "xvid"
-            }
-        ]
-    });
-
-    profile.CodecProfiles.push({
-        Type: "Video",
-        Codec: "h264",
-        Conditions: [
-            {
-                Condition: "EqualsAny",
-                Property: "VideoProfile",
-                Value: "high|main|baseline|constrained baseline"
-            },
-            {
-                Condition: "LessThanEqual",
-                Property: "VideoLevel",
-                Value: codecCaps.h264MaxLevel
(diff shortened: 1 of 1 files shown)


nielsvanvelzen:
Looking through this code, I'm honestly not sure why we're modifying the device profile used for the web player at this point. The webview on Android is based on chromium and jellyfi-web already has reasonable detection for codec support on chromium, there is no reason to modify this profile at all.
For the native player we already have a separate device profile.

Can you try removing the getDeviceProfile function completely and see if everything still works as expected for you? If so, go ahead and update the PR.

Axyss:
Removing `getDeviceProfile()` results in playback errors I'm afraid.

Screenshot

I've tried checking Logcat to investigate the error further, but haven't been able to find any relevant logs. My best guess is that [jellyfin-web/src/components/apphost.js](https://github.com/jellyfin/jellyfin-web/blob/master/src/components/apphost.js#L51) requires us to provide one. Otherwise, we'd be invoking an undefined function:
```js
// jellyfin-web/src/components/apphost.js
function getDeviceProfile(item) {
    return new Promise(function (resolve) {
        let profile;

        if (window.NativeShell) {
            profile = window.NativeShell.AppHost.getDeviceProfile(profileBuilder, __PACKAGE_JSON_VERSION__);
        } else {
            const builderOpts = getBaseProfileOptions(item);
            profile = profileBuilder(builderOpts);
        }
...
```

Defining a very minimal getDeviceProfile seems to do the trick though:
```js
function getDeviceProfile(profileBuilder, item) {
    return profileBuilder();
}
```

nielsvanvelzen:
Ah yeah my bad, we do need to implement it but we shouldn't change anything, so just call the profile builder as-is. What we can remove is the `getSyncProfile` property as that was removed from jellyfin-web from the start (like 10 years ago).
If you could update the PR to your minimum second code sample that would be great.

Axyss:
Tried to figure out what `getSyncProfile` was doing but couldn't find a thing, guess that explains it. Anyways, changes applied 👍

nielsvanvelzen (approved):

nielsvanvelzen:
Thanks!

PR #2155, opened 2026-08-24
Fix Live TV playback by not sending item ID as media source ID

### LINKED ISSUE #2154: Live TV - M3U - Not working on Android

### Describe the bug

I installed Jellyfin in Docker Compose, configured it with a M3U TV Tuner, installed the latest Jellyfin Android app from the app store, connected to my Jellyfin server.

When I attempt to watch any live TV I get a black screen. When I go to my public Jellyfin domain in a browser on my Phone it works fine. I had opened an issue, using LLM to help me write it up, but it was closed. Feel free to look at it if you want some more details that I don't fully understand. https://github.com/jellyfin/jellyfin-android/issues/2153

### Media info of the file

```shell
{
  "Id": "72e66d2ce1f3c5245d3616cf3ba2a0cf",
  "Protocol": "Http",
  "Path": "http://10.0.0.5:8080/stream/channel.ts",
  "Container": null,
  "Bitrate": null,
  "TranscodingUrl": null,
  "TranscodingSubProtocol": "http",
  "RequiresOpening": true,
  "LiveStreamId": null,
  "SupportsDirectPlay": true,
  "SupportsDirectStream": true,
  "SupportsTranscoding": true,
  "IsInfiniteStream": true,
  "RequiredHttpHeaders": {"User-Agent": "Mozilla/5.0 (Windows NT 10.0; ...)"},
  "MediaStreams": [
    {"Type": "Video", "Codec": null, "Index": -1},
    {"Type": "Audio", "Codec": null, "Index": -1}
  ]
}
```

- [ ] Web player (default)

### Client logs

```shell

```

### FFmpeg logs

```shell
None
```

### Application version

2.7.1

Google Play Store

### Device information

Samsung Galaxy S25 Ultra

### Android version

Android 16

### Jellyfin server version

10.11.11

### Other sources

_No response_

wymangr:
**Changes**

The app was sending the ID as the media source ID. For Live TV, those tow IDs differ, so the filter matched nothing and PlaybackInfo came back with no media sources. This change only sends that substituted ID when an audio or subtitle stream index is actually requested. 

**Code assistance**

The issue was debugged and the fix written by a LLM. Kotlin is not my language of expertise, so if this violates the rules, feel free to close out the PR and someone else can contribute to write the fix. 

I did however build the APK and test it on my device which I was experiencing the issue and confirmed it work and I confirmed I could still watch Movies and TV shows as well. 

**Issues**

Fixes #2154

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/player/source/MediaSourceResolver.kt
+++ b/app/src/main/java/org/jellyfin/mobile/player/source/MediaSourceResolver.kt
@@ -42,10 +42,14 @@ class MediaSourceResolver(private val apiClient: ApiClient) {
                 mediaInfoApi.getPostedPlaybackInfo(
                     itemId = itemId,
                     data = PlaybackInfoDto(
-                        // We need to remove the dashes so that the server can find the correct media source.
-                        // And if we didn't pass the mediaSourceId, our stream indices would silently get ignored.
+                        // The server filters media sources by this id and returns none when it matches
+                        // nothing. For most items the media source id equals the item id, but for Live TV
+                        // it is the tuner's stream id, so only substitute the item id when stream indices
+                        // are actually requested - otherwise the response would come back empty.
                         // https://github.com/jellyfin/jellyfin/blob/9a35fd673203cfaf0098138b2768750f4818b3ab/Jellyfin.Api/Helpers/MediaInfoHelper.cs#L196-L201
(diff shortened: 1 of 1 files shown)


nielsvanvelzen (approved):
Ideally we'd check if the item we're trying to play is live tv or not (that's what the TV app does) but I don't think we can easily access that information here, so this will do for now.

PR #2179, opened 2026-09-06
Reset playback speed automatically near the live edge

### LINKED ISSUE #2178: Live TV playback speed doesn't reset near the live edge, causing repeated stall/accelerate cycles
**This issue respects the following points:**
- [x] This issue is not already reported on GitHub _(I've searched it)_.
- [x] I agree to follow Jellyfin's Code of Conduct.
- [x] This report addresses only a single issue.

## Bug information

**Describe the bug**

1. Open a Live TV channel.
2. Let playback fall a few seconds behind the live edge (normal buffering).
3. Increase the playback speed above 1x (speed menu, or hold-to-fast-forward).

The player keeps playing at the increased speed until the buffer runs out, stalls for a couple of seconds, resumes at the same increased speed, and repeats this accelerate/stall cycle indefinitely instead of dropping back to 1x once caught up to the live edge.

Instead, I expect the playback speed to automatically reset to 1x once caught up to the live edge, similar to how other live-streaming clients (e.g. YouTube) behave.

**Which video player implementations does this bug apply to?**

## Environment

**Application version**: v2.7.1 (dev build)
**Where did you install the app from?**: Sideloaded APK (proprietary build)
**Device information**: Google Pixel 8
**Android version**: Android 17
**Jellyfin server version**: Not server-specific; reproducible with any server transcoding Live TV to HLS

Quin0:
**Changes**
When accelerating playback to catch up to a live TV stream, the integrated player kept the increased speed until it ran out of buffer and stalled, then resumed at the same speed once buffered again, repeating indefinitely. This monitors the remaining distance to the live edge while playing faster than normal and resets the speed to 1x once close, similar to how other live-streaming clients (e.g. YouTube) behave.

**Code assistance**
Code assistance (LLM) was used to investigate and implement this change, including on-device log analysis to tune the catch-up threshold. All changes were reviewed and tested on-device by me before submitting.

**Issues**
Fixes #2178

### DIFF
--- a/app/src/main/java/org/jellyfin/mobile/player/PlayerViewModel.kt
+++ b/app/src/main/java/org/jellyfin/mobile/player/PlayerViewModel.kt
@@ -134,6 +134,9 @@ class PlayerViewModel(application: Application) : AndroidViewModel(application),
     private val _error = MutableLiveData<String>()
     val error: LiveData<String> = _error
 
+    private val _playbackSpeed = MutableLiveData<Float>()
+    val playbackSpeed: LiveData<Float> = _playbackSpeed
+
     private val eventLogger = EventLogger()
     private var analyticsCollector = buildAnalyticsCollector()
     private val initialTracksSelected = AtomicBoolean(false)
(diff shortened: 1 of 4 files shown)


nielsvanvelzen:
This implementation looks extremely hacky. A timer that checks every 3 seconds for playback state and changes settings is extremely dirty. If you can figure out a better implementation we could accept it, otherwise I'd rather just disable speed controls for Live TV to fix the issue.

PR #2194, opened 2026-09-13
Fix player hover highlight by isolating the seek ripple

### LINKED ISSUE #1287: Issues found while playing video on Samsung DeX using integrated player.
### Describe the bug

1. When playing video using integrated player while using phone as DeX touchpad. Once you move the cursor after playing, the whole screen will have a white transparent hue similar to that of when you fast forward by double tapping right. To get rid of the hue, you have to double tap to fast forward/rewind once to return to normal.

2. When you tap the touchpad again to bring up ExoPlayer GUI while playing a video, half the screen would be stuck with the transparent circle. To get rid of this again is to double tap fast forward/rewind.

image
image example of the transparent circle/hue mentioned above

### Logs

_No response_

### Application version

2.6.0

Google Play

### Device information

Samsung Galaxy S24 Ultra, Samsung Galaxy S20+

### Android version

Android 14 (OneUI 6.1), Android 13 (OneUI 5.1)

### Jellyfin server version

10.8.13

- [ ] Web player (default)

manuelmorales:
The ripple animation on double-tap to seek was wrongly activated with mouse hover when using Samsung Dex. This caused a translucent white overlay to be nearly always on top of movie, causing colors to be washed out. See #1287 for more details.

The implemented solution was to move the ripple onto a separate view, `seek_ripple_overlay`, positioned above the video. Such view is not clickable, nor focusable, hence solving the problem.

**Changes**

- Added view `seek_ripple_overlay`.
- Updated `onDoubleTap()` to trigger the seek ripple on `playerBinding.seekRippleOverlay.foreground` instead of `playerView.foreground`.

**Code assistance**

The whole thing was reproduced and fixed by ChatGPT 6 (Astra) through Codex. The whole process was as follows:

- Experienced the bug myself manually on my Samsung Galaxy Note 8 on Dex mode (Android 9).
- Codex reproduced it on a Bliss OS 16.9.7 VM (Android 13) dispatching `ACTION_HOVER_ENTER` from `SOURCE_MOUSE` to `PlayerView`. Then checking RGB changing from 0, 0, 0 to 10, 10, 10.
- Codex produced the fix, verified RGB stays black and proposed the code in this PR as is.
- I verified the bug was gone on the Samsung phone myself. Also checked seeking gestures still work.

Please let me know if there are any contribution rules or guidelines I missed.

**Issues**

Fixes #1287


nielsvanvelzen:
I'm not sure what the issue is, when I use the double-tap seek gestures the ripple shows just fine. Both with touch and mouse (on emulator). Please share some before/after images or videos to show what this change does.

manuelmorales:
Hi @nielsvanvelzen, thank you for replying so quickly . @Des-m0nd did a very good job describing the issue here: https://github.com/jellyfin/jellyfin-android/issues/1287. Please let me know if there is anything in there that isn't clear enough.

nielsvanvelzen:
The issue only has a screenshot from a YouTube player, we're Jellyfin. Not YouTube. As I cannot reproduce the issue I need to have some proof about it so I can review this change properly.

manuelmorales:
See the video attached with a comparison. The bug is evident when the legend says "Mouse hover enters PlayerView" (at the ~3s mark). You'll see how the "Before" side is fully covered with a semi-transparent white layer that makes black look grey.

https://github.com/user-attachments/assets/c72c5d4c-faee-4497-a7be-ba54bf230f49

If it's any useful, you can find the code that Codex wrote to reproduce the bug in [RippleProbeActivity.java](https://github.com/user-attachments/files/32162900/RippleProbeActivity.java).

nielsvanvelzen (approved):
Thanks for the video, I understand what the issue is now. I'm not too sure about the approach of using a second view but at the same time it does not seem to cause any issues on mobile so we can do it "properly" once the player UI is remade in compose.

I've pushed one additional commit to remove the AI comment, otherwise good to go now. Thanks!

PR #2200, opened 2026-09-13
Fix audio codec capability merge using wrong map key

en6ads:
**Changes**
`DeviceProfileBuilder` builds `audioCodecs` keyed by codec name (e.g. `"aac"`), but checked `containsKey(mimeType)` (e.g. `"audio/mp4a-latm"`) before merging a newly discovered codec's capabilities into an existing entry. Since the map is never keyed by mimeType, that check was always false, so a codec encountered across multiple `MediaCodecInfo` entries had its capabilities silently overwritten instead of merged. Fixes the check to use the same key the map is actually indexed by (`containsKey(name)`).

This currently has no observable effect on the produced device profile — the merged `DeviceCodec.Audio` capability data isn't consumed anywhere else in this file, only `containsKey` on the codec name is. 

**Code assistance**
Found and fixed with Claude Code while reviewing DeviceProfileBuilder.kt in preparation for a separate AV1 and HEVC transcoding PR; identified as an unrelated pre-existing bug during that review.

**Issues**
None filed separately — found during code review, not reported as a standalone issue.


nielsvanvelzen:
Please write the PR description yourself per our LLM policy and fix your commit authorship.

en6ads:
I'm going to close this for now. It's a one line bug fix, the bug is real, but I'm not sure how to word it in such a way as to pass.
Thank for considering it, apologies I didn't following policy,

